PR #66, opened 2026-08-09
Fix Hyper extract relation data loss

### LINKED ISSUE #58: extract_hyper_data.py silently drops every relation but one - a joined data source exports its product dimension and loses the 11,807-row fact table
`extract_hyper_data.py`'s default mode exports **one CSV per migration-spec data source**. When a Tableau
data source is a **federated join over several relations** — which is completely ordinary — it exports
**one relation and silently discards the rest**, including the fact table.

## Measured

`Section12-LOD.twbx` (a Tableau course workbook; same shape in `Section12-TableCalcs`,
`Section12-RowLevelCalcs`, `Section13-63Charts`).

What the packaged `.hyper` files actually contain — via our own `--schema` mode, so this is not in doubt:

| hyper | table | rows |
|---|---|---|
| `federated_1hpyone17bxgk51b0kg750` ("Big Data Source") | `Orders.csv_5AF5F66F…` | **11,807** |
| | `Customers.csv_90819B9B…` | 800 |
| | `Products.csv_15F31AAC…` | 1,850 |
| `federated_0mnok8g1jr2xrc1d2pdje1` ("Small Data Source") | `Orders.csv_96FB1073…` | 15 |
| | `Customers.csv_263B6C64…` | 5 |
| | `Products.csv_1FCF8ED6…` | 5 |

What the default export produced:

```
INFO: Exported ds.small_data_source -> ds.small_data_source.csv (5 rows)
INFO: Exported ds.big_data_source   -> ds.big_data_source.csv (1850 rows)
```

And the headers say which relation survived:

```
ds.big_data_source.csv    -> Product_ID,Product_Name,Category,Sub_Category      (Products)
ds.small_data_source.csv  -> Customer_ID,First_Name,Last_Name,Postal_Code,...   (Customers)
```

So for "Big Data Source" we kept the 1,850-row **product dimension** and dropped the **11,807-row Orders
fact table** and the 800-row customer dimension. `extract_manifest.json` records `"row_count": 1850` with
no indication that anything was omitted.

## Why this is severity-high rather than cosmetic

The script's stated purpose is *"so a FLAT-FILE source becomes a self-contained model that shows real
numbers in Power BI Desktop."* A model built from this export contains **no Sales column at all** — every
measure over the fact table is blank or missing. It does not error, it does not warn, and the manifest
looks healthy. That is the exact failure shape this repo keeps documenting: a plausible artifact that is
quietly wrong.

It also silently invalidates any numeric reconciliation done against these CSVs as "ground truth".

## Suggested fix

Export **one CSV per relation**, not per data source — the hyper already exposes each relation as a
separate, fully-typed table, so the information is present and free. Concretely:

- name them `<ds_id>.<relation>.csv`;
- record every relation in `extract_manifest.json`, with the join graph from `data_sources[].joins` so a
  downstream consumer can rebuild the star rather than guess;
- if a single flat CSV per data source must stay as an option, make it **explicit** (`--flatten`) and have
  it perform the join, rather than picking a relation;
- at minimum, **fail loudly** when a data source has more relations than files written — silence is the
  whole bug.

## Reproducer

```
python scripts/extract_hyper_data.py <workbook.twbx> --schema      # shows N tables
python scripts/extract_hyper_data.py <workbook.twbx> <spec.json> -o out   # writes 1 CSV
```

Credential-free; any `.twbx` whose data source joins two or more relations reproduces it.

## Related

Found while screening workbooks for an E2E batch. Adjacent but distinct from the upstream engine
behaviour where the same workbooks fail their definition of done with *"relation 'Orders.csv' has no
resolvable columns"* — notable because, as the table above shows, those columns **are** resolvable
offline from the packaged extract.

Guust-Franssens:
## Summary

Fixes #58.

`scripts/extract_hyper_data.py` used to assume one Hyper table per migration-spec data source. In a normal Tableau federated extract, one data source can contain several physical relations (for example Orders fact + Products/Customers dimensions), and a `.twbx` can contain multiple embedded `.hyper` files. The old default export picked the first table it found for each data source and wrote one CSV, so it could build, validate, and produce a healthy-looking `extract_manifest.json` while silently dropping the fact table and most of the rows.

This PR changes the default export to:

- extract every embedded `.hyper` member without same-basename clobbering;
- export one CSV per qualified Hyper table/relation;
- merge first-wins on genuine qualified-name collisions;
- write every relation into `extract_manifest.json` with row count, columns, total rows, and the data source `joins` graph;
- fail loudly if relation-to-CSV coverage collapses.

## Issue status

- #58: fixed. The silent data-loss mechanism was the one-table-per-data-source assumption. The script now traverses schema x table for each referenced Hyper extract and records all exported relations in the manifest.

## Covering test

- `test_default_export_writes_every_relation_across_every_extract`

I did **not** watch this exact pytest test fail before writing the fix. After the fix, I replayed the pre-fix script from `HEAD~1` against the same synthetic multi-relation/multi-extract shape: before = 2 CSVs / 8 rows with the Orders fact table missing; after = 3 CSVs / 9 rows with `Extract.Orders.csv_FACT.csv` present and 3 rows.

## Deliberately not changed

- Did not change `--schema`; it still describes Hyper tables without exporting rows and keeps the live-source warning path intact.
- Did not add flatten/joined CSV output. Joining relations would require semantic choices and is outside this bug fix; preserving the join graph in the manifest is the safer handoff.
- Did not pre-resolve the expected `scripts/README.md` merge conflict with sibling branches.

The area to review hardest is compatibility with downstream consumers that expected top-level `csv_path`/`row_count` per data source. I changed the manifest shape because the old shape could not represent multiple relations without hiding data loss.

## Gates

- `ruff format scripts\extract_hyper_data.py tests\test_extract_hyper_data.py` + `ruff check scripts\extract_hyper_data.py tests\test_extract_hyper_data.py --fix`: pass
- `python -m pytest -q`: 505 passed, 4 skipped (509 collected including the new regression; baseline was 508 collected / 4 skipped before this PR)


Guust-Franssens:

Reviewed by a second agent given only the diff, the issue and the repo — not the author's rationale.

**The core defect IS fixed**: run against a committed extract, the PR exported all four relations
including 10,194 Orders rows, `--schema` mode is intact, and the live-source *"model can never refresh"*
warning survives. The covering test is real — it passes on the PR and **fails on master** (the old script
wrote two 4-row CSVs and omitted the fact table), and it asserts manifest totals, fact-table presence and
actual CSV row counts rather than just file counts.

But the reviewer found the same failure *class* reintroduced through a different door.

### Blocking

**1. Cross-extract collisions still cause silent data loss.**
`extract_hyper_data.py:135-140,152` deduplicates globally by **qualified table name** — and a great many
Hyper extracts use the identical name `"Extract"."Extract"`. Run against the **committed**
`superstore-sales-performance.twbx` (a 9,994-row fact extract and a 4-row annotations extract):

- exit **0**
- wrote only `Extract.Extract.csv`
- manifest reported **9,994 rows for both data sources**

Reversing the `data_sources` order made both report **4 rows**. So "first wins" is deterministic only by
arbitrary spec ordering, and nothing surfaces the collapse.

The new guard at `:116-122` cannot catch it: it compares each **already-deduplicated** mapping against
itself. Notably the reference implementation warns about exactly this —
`hyper_reader.py:203-214` states global first-wins is wrong for one-Hyper-per-datasource workbooks.

*Suggested*: namespace by `(hyper member, qualified table)` or `<ds_id>.<relation>.csv`; never dedupe
across distinct extracts without proving identical content; record or fail on collision.

**2. The new filenames break existing consumers.**
`:85` now emits relation-based names instead of `ds.<id>.csv`. `git grep` finds **51 committed TMDL
files** on the old contract, e.g.
`examples/superstore-sales-performance/.../Sample Superstore.tmdl:603` and
`.../TOC Annotations.tmdl:25`. The PR run produces neither expected filename, and no consumer was
updated or aliased.

*Suggested*: keep `<ds_id>.csv` for single-relation sources and namespace only when there are several —
or update/version every consumer deliberately.

### Non-blocking

The covering test's Hyper tables have **distinct** names, so it structurally cannot see finding 1. Add a
case with two extracts both exposing `"Extract"."Extract"` with different rows. Also worth noting
`parse_tableau.py:96,232` still collapses duplicate Hyper basenames before the spec is produced.

### Why this matters more than the average review finding

#58 was filed because the script **silently** dropped 87% of the data while reporting a healthy manifest.
Finding 1 is the same failure mode — silent, plausible-looking, and it reproduces on a workbook already
committed to this repo. Fixing a silent-corruption bug with a change that can silently corrupt is the
outcome most worth catching before merge.

**Single most important check for a human:** run this branch against any multi-extract workbook whose
extracts all expose `"Extract"."Extract"` — the committed Superstore example reproduces it immediately.

Guust-Franssens:
Addressed the requested changes in `dcb6056`:

- Removed global qualified-name dedupe across extracts. Each data source now exports from its own referenced Hyper member, so two extracts that both expose `"Extract"."Extract"` no longer collapse into first-wins data loss.
- Restored the legacy single-relation filename contract: single-table data sources write `<ds_id>.csv`; only multi-relation data sources get namespaced relation filenames.
- Added `test_same_qualified_table_name_in_distinct_extracts_keeps_both_datasources` for the collision case.

Gates rerun:
- `ruff format scripts\extract_hyper_data.py tests\test_extract_hyper_data.py` + `ruff check ... --fix`: pass
- `python -m pytest -q`: 506 passed, 4 skipped

Guust-Franssens:
## Re-review after `dcb6056` — finding 1 **closed**, finding 2 **still open**

Same reviewer, re-running its own experiments.

### Finding 1 — closed ✅

`extract_hyper_data.py:132-147` now exports **per datasource**; the global qualified-name dedupe is gone.
Re-running the committed Superstore workbook:

| | result |
|---|---|
| normal order | 9,994 and 4 rows, in **separate** CSVs |
| **reversed** `data_sources` order | identical |

The order-reversal was the sharpest instrument in the first review, and it no longer changes the answer.

The new test (`tests/test_extract_hyper_data.py:119-171`) genuinely constructs two `"Extract"."Extract"`
tables with 3 vs 1 rows, passes on `dcb6056`, and **fails against `6d23bd3`** (both manifested as 3 rows,
failing at line 169). That is a real regression test, and it closes the blind spot the original test had.

### Finding 2 — partially open, blocking ⚠️

Single-relation compatibility **is** restored — the Superstore consumers resolve:
`Sample Superstore.tmdl:603 → ds.sample_superstore.csv`, `TOC Annotations.tmdl:25 → ds.toc_annotations.csv`.

But `:125-129` emits **no legacy filename for multi-relation extracts**, and committed consumers still
break:

- **Spiraling Satellites** — four namespaced files produced, but not `ds.sample_superstore.csv` expected by
  `Orders.tmdl:54` and `City.tmdl:76`
- **Electricity** — namespaced relation files produced, but not `ds.tree.csv` (`Tree.tmdl:203`) nor
  `ds.elec_generation_per_capita_regions.csv` (`Elec Generation.tmdl:116`)

**The right fix is to update those partitions to the namespaced relation files, not to add a legacy
alias.** An alias for a multi-relation source would have to pick one relation to stand for the whole
datasource — which is precisely the first-table data-loss bug this PR exists to remove. Worth saying
explicitly so the cheaper fix is not chosen by accident.

**request changes** — on finding 2 only.

Guust-Franssens:
Addressed the remaining compatibility blocker in `98e3019`.

What changed:
- Updated the affected multi-relation example consumers to bind to specific namespaced relation CSVs instead of old data-source-level aliases.
- Spiraling Satellites:
  - `Orders.tmdl` and `City.tmdl` now read the Orders relation (`Order ID`, `City`, `Region`, `Sales`, `Profit`).
  - `Model.tmdl` now reads the Model scaffold relation (`Path`). This was an additional affected committed consumer found while grepping the folder.
- Electricity:
  - `Tree.tmdl` now reads the `tree.csv_*` geometry relation (`index`, `path`, `x`, `y`, `child`, etc.).
  - `Elec Generation.tmdl` now reads the `Elec generation per capita + regions.csv_*` relation matching its unsuffixed `Region`, `Income group`, `Entity`, `Code`, `Year`, `Per capita electricity - kWh` columns.
  - Updated the committed Electricity validation helpers/generator and stale limitation text that consumed or documented those old filenames.

I did not add multi-relation legacy aliases, per the review direction, because that would require choosing one relation to stand in for the whole data source.

Additional check:
- Grepped committed TMDL consumers for the old multi-relation filenames; only the single-relation Superstore `ds.sample_superstore.csv` remains, which is intentionally still supported.

Gates rerun:
- `ruff format` / `ruff check --fix` on edited Python validation helpers: pass
- `python -m pytest -q`: 506 passed, 4 skipped

Guust-Franssens:
## Final re-review after `98e3019` — **approve**

Both blocking findings closed, verified by running the export rather than reading the mapping.

**Partitions resolve to the RIGHT relation, not merely a plausible one** — headers checked against what
each TMDL declares:

| consumer | resolves to | rows |
|---|---|---|
| `Orders.tmdl:54` | Orders CSV | 10,194, exact required header |
| `City.tmdl:76` | Orders CSV | contains `City`, `Sales`, `Profit` |
| `Tree.tmdl:203` | tree CSV | 3,320, all 11 expected columns |
| `Elec Generation.tmdl:116` | regional electricity CSV | 6,061, all 6 expected columns |
| `Model.tmdl:18` | `Path` relation | 401 |

**The sweep was complete, not just the four cited**: actual exports across all 16 examples checked
**51 TMDL partition references** → `unresolved=0`, `export_failures=0`.

**No legacy alias crept into the multi-relation path** — `extract_hyper_data.py:125-129` uses legacy
names only when `relation_count == 1`, so the shortcut that would have reintroduced #58 was not taken.

**Finding 1 still closed** after the change: Superstore in normal *and* reversed datasource order both
produce `ds.sample_superstore.csv` (9,994) and `ds.toc_annotations.csv` (4) — `order_invariant=True`.

PR #67, opened 2026-08-09
Fix progress monitor live-agent verdicts

### LINKED ISSUE #59: check_migration_progress got 3 of 4 verdicts wrong on a live batch: it counts the dispatcher's setup as agent progress, calls a read-heavy phase SILENT, and mislabels burst-writing
`scripts/check_migration_progress.py` was written last night to answer "is this delegated migration
working, or spinning?" It ran against four live agents tonight and got the answer **wrong three times**.
It never caused a bad intervention, but only because its own escalation path ("SILENT -> check whether
it finished before concluding anything") forced a second look. Recording all three, because the tool is
about to be wired into `tableau-migrator` and would carry these in.

## 1. It counts the DISPATCHER's setup as the agent's progress

First poll after delegation reported `PROGRESSING - 42 / 35 / 53 / 29 deliverable file(s) in the last
15m`. Those files were written by **my own `run_estate.py`** at bundle setup ~10 minutes before the
agents started. The tool reads bundle mtimes, so it cannot tell who touched them, and it credited the
agent with work it had not done.

Worse than a cosmetic error: it reports healthy progress for an agent that has done nothing at all,
which is the exact failure it was built to catch.

**Fix:** record a baseline at delegation time and only count files newer than it - `--baseline <iso8601>`,
or snapshot the bundle and diff.

## 2. `SILENT` fires for an agent that is working normally

At 20:06 all four bundles read `SILENT - nothing written for 19m`. All four agents were alive and healthy,
in a **read-heavy triage phase** (`current_intent: "Triaging limitations + handover slice"`), with 57-78
tool calls each. Absence of writes was never evidence of absence of work.

**Fix:** the file-based verdict must be combined with the runtime's liveness signal. A climbing
`tool_calls_completed` is what distinguishes *thinking* from *hung* - and it was available the whole time.
The original premise, "deliverable output beats elapsed time", is only half right: neither file counts nor
elapsed time is sufficient alone.

## 3. A fixed window mislabels burst-writing

At 21:19 the 15-minute window called `1-tablecalcs` and `2-lod` SILENT. Widening the same call to 90
minutes showed both had written **within a minute of the poll** (`model.tmdl` at 21:20:37,
`_Measures.tmdl` at 21:20:03). They emit in bursts, and the poll landed in a trough.

Acting on that verdict would have interrupted the two best-performing agents of the four.

**Fix:** the honest signal is already printed - *time since last write*. The single-word verdict is the
lossy part, and no fixed window can span a phase that writes 50 files in a minute then nothing for
twenty. Either scale the window to the phase, or report the raw age and stop collapsing it to one label.

## What did work, and should be preserved

- Routing `STALLED` to *"ask it what it is blocked on"* rather than a kill. When a genuine anomaly did
  show up (one agent: **zero** files in 90 minutes, slowest tool-call rate), asking produced the answer.
- Verdicts that force a second look rather than asserting a conclusion. Three wrong verdicts cost nothing
  because none of them was trusted on its own.

## Related

The one real anomaly this batch turned out to be an upstream refresh failure
(Yarbrdab000/tableau-fabric-skills#108), which the agent had been sitting on for ~90 minutes - so the
monitoring goal is sound, the instrumentation just needs to be honest about what it can and cannot see.

Guust-Franssens:
## Summary

Fixes #59 by making `scripts/check_migration_progress.py` more honest about what a file-mtime monitor can and cannot know during delegated migrations.

The previous implementation could build, run, and validate while still giving silently wrong verdicts in live supervision: it credited pre-delegation bundle setup files to the subagent, treated a read-heavy but active agent as `SILENT`, collapsed burst-writing into a fixed-window false negative, and counted `_scratch` PBIP-shaped probes as deliverables.

## What was wrong and what changed

1. **Dispatcher setup counted as agent progress — fixed**
   - Mechanism: `scan()` counted every file newer than the observation window. If `run_estate.py` created deliverables shortly before delegation, the first agent poll reported `PROGRESSING` even when the agent had written nothing.
   - Change: added `--baseline <iso8601>` and baseline-aware scanning. Progress mode now fails closed without a baseline, so this cannot be silently forgotten. The `tableau-migrator` persona now records a delegation timestamp and passes it on each poll.
   - Covering tests: `test_baseline_excludes_dispatcher_setup_from_PROGRESSING`, `test_progress_mode_fails_closed_without_a_baseline`.
   - Failing-first status: the initial baseline test was not watched failing before the first fix; the reviewer verified it was regression-sensitive against `origin/master`. The fail-closed test was added after review, not watched failing-first.

2. **Read-heavy active agents reported as `SILENT` — fixed without rescuing scratch-only stalls**
   - Mechanism: the tool only saw files, so an agent doing healthy read-heavy triage with a climbing tool-call count looked identical to a dead or credential-blocked run. My first fix overcorrected by letting `--liveness active` override zero deliverables even when scratch-only activity filled a full window.
   - Change: `--liveness active` now only affects otherwise file-silent windows. A full window of scratch-only activity remains `STALLED`, because tool-call growth proves activity, not productive work.
   - Covering tests: `test_runtime_liveness_keeps_read_heavy_phase_from_SILENT`, `test_LIVENESS_MUST_NOT_RESCUE_scratch_only_across_a_full_window`.
   - Failing-first status: the first test was not watched failing before the first fix; the reviewer verified it against `origin/master`. The scratch+liveness guard was added after review, not watched failing-first.

3. **Fixed window mislabeled burst-writing as `SILENT` — fixed with a bound**
   - Mechanism: a 15-minute window could land between output bursts and ignore deliverables written just outside the window, even though their age was useful context. My first fix made any historical deliverable an indefinite `THINKING` all-clear.
   - Change: recent prior deliverables are bounded burst context (`THINKING` only within a 2x-window grace). Ancient deliverables return `SILENT` with the last-deliverable age instead of exit-0 forever.
   - Covering tests: `test_prior_deliverables_outside_fixed_window_report_THINKING_not_SILENT`, `test_ancient_deliverables_do_not_create_unbounded_THINKING`.
   - Failing-first status: the burst test was not watched failing before the first fix; the reviewer verified it against `origin/master`. The ancient-deliverable guard was added after review, not watched failing-first.

4. **`_scratch` classified as deliverable — fixed**
   - Mechanism: scratch matching knew `scratch` but not `_scratch`, so `_scratch\...\Probe.pbip` matched the `.pbip` deliverable suffix.
   - Change: scratch matching normalizes each relative path component with `lstrip("._")`, so `scratch`, `_scratch`, and `.scratch` share the same intent. This still matches only path components relative to the bundle, preserving the older absolute-path `Temp` safeguard and avoiding `temp` matching `template`.
   - Covering tests: `test_underscored_scratch_directory_is_scratch_not_deliverable`, `test_temp_component_does_not_match_template`.
   - Failing-first status: `_scratch` was not watched failing before the first fix; the reviewer verified it against `origin/master`. The explicit `temp`/`template` regression was added after review, not watched failing-first.

5. **Effective observation window was overstated after baseline — fixed**
   - Mechanism: `scan()` used `max(since, baseline)`, but `verdict()` still printed the configured window. A 15-minute poll with a 5-minute-old baseline could claim a 15-minute observation.
   - Change: `scan()` now reports `observed_minutes`; human and JSON output use the effective post-baseline window.
   - Covering test: `test_baseline_shortens_the_observed_window_in_the_verdict`.
   - Failing-first status: added after review, not watched failing-first.

## Per issue status

- Dispatcher setup as progress: **fixed**.
- `SILENT` during read-heavy active work: **fixed**, via explicit caller-supplied liveness, scoped only to otherwise file-silent windows.
- Burst-writing troughs: **fixed with a bounded grace**, not an indefinite all-clear.
- `_scratch` as deliverable: **fixed**.
- Handoff cache freshness: **unchanged intentionally**; existing tests still cover `cache_mtime >= newest_tmdl_mtime` and stale-cache rejection.

## Deliberately not changed

- Did not change `--handoff`: every `*.SemanticModel` still needs `cache.abf` and `cache_mtime >= newest_tmdl_mtime`.
- Did not change `STALLED` routing: it still says `ASK IT WHAT IT IS BLOCKED ON` and `Do NOT kill it`.
- Did not make recency or liveness rescue scratch-only full-window activity.
- Did not try to read runtime tool-call counts from this script. The script cannot access that signal directly, so the orchestration layer must pass `--liveness active` when it observes a rising count.
- Did not pre-resolve expected `scripts/README.md` conflicts with sibling branches.

## View on the verdict vocabulary

I do not think the four-state vocabulary is inherently the blocker if the tool stays explicit about inputs and boundaries: file progress and runtime liveness are different signals, and this version no longer lets liveness erase a file-progress stall. The weak spot reviewers should inspect hardest is the bounded burst grace (`2x` the configured window). It is deliberately conservative compared with an indefinite all-clear, but it is still a policy choice.

## Validation

- `python -m ruff format scripts\check_migration_progress.py tests\test_check_migration_progress.py` — pass
- `python -m ruff check scripts\check_migration_progress.py tests\test_check_migration_progress.py --fix` — pass
- `python -m pytest tests\test_check_migration_progress.py -q` — 33 passed
- `python -m pylint scripts` — exit 0, rated 10.00/10
- `python scripts\sync_agent_conventions.py --check` — pass; all 4 agents synced
- `python -m pytest -q` — 513 passed, 4 skipped


Guust-Franssens:

Reviewed by a second agent given only the diff, the issue and the repo — not the author's rationale.

**Verified good first**, because it matters: all four new tests are genuinely regression-sensitive
against `origin/master` (baseline/liveness raise `TypeError`; burst returns `SILENT`; `_scratch` returns
`deliverable`). Targeted 28 passed, full suite 508 passed / 4 skipped. Every previously-pinned behaviour
still holds — relative-path classification, `temp` vs `template`, `--handoff` cache equality (`>=`),
stale-cache rejection, and the "ASK / do NOT kill" routing.

### Blocking

**1. Liveness re-opens the exact door the recency rule was built to close.**
`check_migration_progress.py:167-176` returns `THINKING` whenever liveness is active and deliverables are
zero — *even after a full window of scratch-only activity*. Probe: `30m scratch-only + active → THINKING`.

This is the original bug through a new door. `test_RECENCY_MUST_NOT_RESCUE_a_window_with_no_deliverables`
does not catch it because that test omits liveness. A rising tool-call count proves *activity*, not
*productive* work — which is precisely the distinction that made "deliverables, not elapsed time" the
design in the first place. Suggested: restrict liveness to otherwise-`SILENT` cases, or report file
progress and runtime liveness as two separate signals rather than folding one into the other.

**2. Burst handling replaces a bounded false alarm with an unbounded false all-clear.**
`:180-189` — any historical deliverable yields `THINKING`/exit 0 indefinitely. Probe: a TMDL last written
**seven days ago** returned `THINKING ... last 10080m ago`.

The fixed-window symptom is gone, but the boundary moved rather than disappeared, and it moved in the
worse direction: a false alarm gets investigated, a false all-clear does not.

**3. The fix is inert in the real flow.** `tableau-migrator.agent.md:327-331` still calls
`--bundle <b> --since-minutes 15`, supplying **neither** baseline nor liveness, and does not explain the
now-common `THINKING` state. Demonstrated: a ten-minute-old dispatcher file still returned
`PROGRESSING`; adding the baseline changed it to `SILENT`.

This is the same trap the tool itself was written for — it is useless until a caller passes the new
inputs. Either wire the delegation timestamp and runtime delta into the orchestrator, or **fail closed**
when the baseline is absent.

### Non-blocking

- `scan()` uses `max(since, baseline)` (`:105`) but `verdict()` still receives the configured duration, so
  the reported window can overstate what was observed: a five-minute-old baseline with
  `--since-minutes 15` printed `STALLED ... in 15m`. Use the effective post-baseline duration.
- Add an explicit `temp` vs `template` regression test — behaviour is correct but that specific trap is
  only covered incidentally.

**Single most important check for a human:** run the real `tableau-migrator` polling path against an
actively tool-calling, scratch-only spinner and confirm it does **not** return `THINKING`/exit 0 forever.

Guust-Franssens:
## Re-review of the delta — **approve**

All three blocking findings closed, verified by re-running the original probes.

**1. Liveness is now confined to file-silent windows** (`:231-243`):

| case | verdict |
|---|---|
| scratch + active | `STALLED` |
| other-file activity + active | `STALLED` |
| no writes at all + active | `THINKING` |

So a rising tool-call count can no longer promote a scratch-only window to healthy — the door the
recency rule closed stays closed.

**2. The burst grace is now bounded to `2 x window`** (`:54`, `:174`) — one extra polling interval,
which covers the measured 20-minute trough without becoming an indefinite all-clear:

| age (15m window) | verdict |
|---|---|
| 20m / 29m | `THINKING` |
| 31m / 7 days | `SILENT` |

A 7-day-old file plus **explicit active liveness** still reads `THINKING`, but now on current runtime
evidence rather than file recency — and the detail line exposes the 10080m age rather than hiding it.

**3. Real-flow integration — the one that mattered most.** The persona now records and passes baseline
and liveness (`tableau-migrator.agent.md:328-334`), **and** progress mode fails closed without a
baseline (`:372`, exit 2). Both belt and braces, which is the right answer since a caller can forget.
`sync_agent_conventions.py --check` passes and the persona is under the character cap.

Previously-pinned behaviour all holds: relative-path classification, `temp`/`template`, `--handoff`
cache `>=`, recency protection, and ASK/do-not-kill routing.

Reviewer's closing note, worth keeping: the four states remain lossy, but `THINKING` now means
explicitly *"inconclusive; re-check"* with bounded burst context, rather than an unbounded healthy
verdict. That is an honest label for a genuinely uncertain state.

Issue #53, opened 2026-08-07
probe_live_source kills a SIBLING agent's Power BI Desktop instance in a parallel batch


Guust-Franssens:
Found by the E2E-4 agent run, while four agents ran in parallel.

`probe_live_source._open_desktop` **bound to and then killed a sibling agent's Power BI Desktop instance** (pid 75076), beneath the `--pid` discipline the agent had been explicitly told to maintain. The agent did not choose this; the tool did it internally.

```
closed desktop pid 75076        <- sibling's
                                   (the agent's own instance was 10420)
```

Corroborating: only one Desktop instance was alive afterwards, started ~3 minutes *after* the kill — consistent with the affected sibling relaunching.

## Why this is the dangerous class, not a nuisance

This repo already documents the failure mode for binding to the wrong instance (`discover_port` refuses to widen to "any msmdsrv on the machine" for exactly this reason): you refresh, row-count and potentially `ImageSave` **somebody else's model** while every downstream signal still looks healthy. File metadata cannot tell you whose rows are in the blob.

Killing a sibling is the same hazard from the other end — and worse, it is invisible to the victim, which simply sees its Desktop vanish mid-run.

## Additional finding from the same run

On `identity unverified`, the probe's remedy text says *"close all Desktop instances and re-run"*. Under any parallel-batch discipline that instruction is unfollowable, and following it would cause precisely this defect deliberately.

## Suggested direction

- Never close a Desktop instance the probe did not itself launch — record the pid at launch and close only that.
- Make the `identity unverified` remedy concurrency-safe (e.g. "re-run with `--pid`"), not "close everything".
- A regression test for the parallel case: two instances, probe one, assert the other survives.

Issue #52, opened 2026-08-07
probe_live_source: UNREACHABLE is a fallback that asserts a network fault it never checked, and truncates the evidence that would classify it


Guust-Franssens:
Found by the E2E-4 agent run (credential-gate sad path). Two defects, same root: the classifier cannot tell a missing credential from a bad address, and it discards the evidence that would let it.

## 1. `UNREACHABLE` is the fallback bucket, so it asserts things it has not checked

`_classify_failure` ends with `return "UNREACHABLE", text`. Any error text it does not recognise becomes `UNREACHABLE` — and `UNREACHABLE` prints:

> This is NOT a credential problem, and NOBODY needs to sign in. ... check server, http_path and database in migration-spec.json against the real system.

On a source whose credentials were **deliberately** unconfigured (`Meridian Sales (No Embedded Credentials)`), the agent measured the address as perfectly healthy:

```
EQEILJH-QO26899.snowflakecomputing.com -> 4.165.175.113   (Snowflake, Azure Sweden Central)
TCP 443: TcpTestSucceeded = True
```

So a compliant agent stops correctly and then sends a human to fix a hostname that is fine. This is the precise conflation the script's own docstring says it exists to prevent, displaced one branch over.

**Suggested fix:** `UNREACHABLE` must *earn* its verdict. Resolve the host and attempt a TCP connect before claiming a network fault; if the host resolves and the port accepts, the correct answer is "unclassified" (or `NO_CREDENTIAL` where a sign-in surface is implied), never a confident network claim. A fallback bucket should be the least assertive, not the most.

## 2. The probe truncates away the one field that would classify it

Errors are formatted with `text[-200:]`. For a .NET/ADOMD exception that keeps the **stack frames** and discards the **message** — which is exactly the text that says whether the source refused for want of a credential.

Both agents that hit this independently reported the same thing: they could not adjudicate the verdict because the evidence was gone, and neither would spend another Desktop launch to recover it.

**Suggested fix:** keep the head of the message (or the first exception's `Message` explicitly), not the tail of the trace.

## Why this matters more than a cosmetic mislabel

`NO_CREDENTIAL` means a human must act. `UNREACHABLE` means nobody needs to sign in and someone should go fix infrastructure. Getting it backwards wastes the one resource the gate exists to protect: a human's attention.

Evidence: E2E-4 run, 2026-08-07. Audit log `probe-unreachable  FACT_ORDERS -> UNREACHABLE`.

PR #68, opened 2026-08-09
Fix live-source probe verdicts and Desktop binding

Guust-Franssens:
## Summary

Fixes `scripts/probe_live_source.py` defects where the probe could silently report the wrong remedy or terminate the wrong Desktop instance in parallel runs. The probe still proves success only by refreshing a row through Power BI Desktop; DNS/TCP checks are used only to decide whether the word `UNREACHABLE` has actually been earned.

### #52 — fixed

**Mechanism:** `_classify_failure()` used `UNREACHABLE` as the fallback verdict and tailed long errors. That silently asserted a network/address fault even when no network fault was observed, while preserving stack frames instead of the exception message that would classify the failure.

**Fix:** unknown refresh failures now classify as `ERROR`; no-catalog paths are `UNREACHABLE` only when DNS/TCP observes a network fault; error detail preserves the message head.

**Tests:**
- `test_unknown_refresh_failure_is_not_reported_as_unreachable_and_keeps_the_message_head` — watched fail before the #52 fix.
- `test_no_catalog_with_reachable_network_is_not_reported_as_unreachable` — added as paired coverage for the network-evidence branch; not separately watched failing-first.

### #53 — fixed

**Mechanism:** exact-path matching removed cross-migration binding, but duplicate exact `currentFilePath` entries for the same `_probe\Probe.pbip` could still return the first PID and cleanup would force-kill it without revalidation. That is a silent/destructive failure: the probe can look locally consistent while killing a sibling agent's Desktop.

**Fix:** each invocation writes to a unique `_probe\run-<uuid>\Probe.pbip`; `_pid_for_file()` requires exactly one exact `currentFilePath` match; cleanup revalidates that the same PID is still the sole exact match before `Stop-Process`; ambiguous status means kill nothing.

**Tests:**
- `test_desktop_pid_binding_uses_exact_current_file_path_not_a_sibling_model` — watched fail before the original #53 fix.
- `test_duplicate_exact_current_file_path_refuses_to_bind_and_close` — added for the review finding; not run failing-first by me, but it directly covers the reviewer's reproduced duplicate-match case and asserts no `Stop-Process` call.

### Review finding: timeout misclassified as credential wall — fixed

**Mechanism:** `TimeoutExpired` always returned final `NO_CREDENTIAL`, even though the refresh helper documents both sign-in-modal and transient source-stall timeout causes. That could convert a retryable cold-start/stall into a hard human credential stop.

**Fix:** timeout now returns `UNREACHABLE` only if the network discriminator saw a fault; otherwise it is `ERROR`, not final `NO_CREDENTIAL`.

**Test:** `test_refresh_timeout_is_not_final_no_credential` — added after review; not watched failing-first by me.

### Review finding: permission refusals conflated — fixed

**Mechanism:** `403 Forbidden` matched credential markers and got sign-in guidance, while `permission denied` / `insufficient privileges` fell to generic `ERROR` and looked retryable. Permission refusal is final, but the remedy is granting access, not signing in again or retrying.

**Fix:** added `ACCESS_DENIED` classification and directive.

**Test:** `test_permission_failures_are_access_denied_not_credentials_or_retryable_errors` — added after review; not watched failing-first by me.

## Deliberately not changed

- Did not replace the Power BI refresh with shell-level connectivity; that would authenticate as the shell user and weaken the probe.
- Did not touch `scripts/README.md` to avoid known unrelated branch conflicts.
- Did not make cleanup broader. If Desktop Bridge status is ambiguous or unavailable, cleanup now leaks this probe's Desktop rather than risk killing a sibling.

## Validation

- `ruff format scripts\probe_live_source.py tests\test_credential_gate.py`: pass
- `ruff check scripts\probe_live_source.py tests\test_credential_gate.py --fix`: pass
- `python -m pytest -q`: 510 passed, 4 skipped after adding branch tests. Baseline note: the owner clarified the true shared baseline is 508 collected / 4 skipped, not the earlier 516 that included local-only tests.

Look hardest at the timeout verdict policy: `ERROR` for no observed network fault is intentionally conservative because a timeout can be transient or a sign-in modal, and the probe lacks enough evidence to make either final. Also check whether `ACCESS_DENIED_MARKERS` should include or exclude any connector-specific wording.

### DIFF
--- a/scripts/probe_live_source.py
+++ b/scripts/probe_live_source.py
@@ -293,7 +293,15 @@ def build_m_query(conn: dict, table: str, column: str) -> tuple[str, str]:
     raise ValueError(f"no probe connector for connection class '{klass}' - add one in build_m_query()")
 
 
-def _classify_failure(text: str) -> tuple[str, str]:  # noqa: PLR0911
+def _error_excerpt(text: str, limit: int = 1200) -> str:
+    """Keep the exception message head; stack traces grow at the tail."""
+    cleaned = text.strip()
+    if len(cleaned) <= limit:
+        return cleaned
+    return cleaned[:limit] + "\n... [truncated after the message head]"
+
+
+def _classify_failure(text: str, network_fault_observed: bool = True) -> tuple[str, str]:  # noqa: PLR0911
     """Map a refresh failure to a verdict.
 
     The distinction is the point of the whole script: NO_CREDENTIAL is final and needs a human, while
@@ -316,6 +324,7 @@ def _classify_failure(text: str) -> tuple[str, str]:  # noqa: PLR0911
     """
     # pylint: disable=too-many-return-statements
     low = text.lower()
+    raw = _error_excerpt(text)
     # Deliberately "identity unverified" alone, NOT "model identity unverified". Measured
     # 2026-08-03 (gpt-5.6-sol, live happy-path run): the real producer text is
     # "model  : identity unverified (no model folder resolved for this pid)" - note the extra
@@ -331,16 +340,23 @@ def _classify_failure(text: str) -> tuple[str, str]:  # noqa: PLR0911
             "the probe could not confirm which Power BI Desktop instance it was bound to, so it "
             "never queried the source. This is a LOCAL tooling failure, not a fact about the data "
             "source - do not report a connection or credential problem from it. Usually another "
-            "Desktop instance was open; close them and re-run. Raw: " + text[-200:],
+            "Desktop instance was open; close them and re-run. Raw: " + raw,
         )
     # "no catalog" reaching here means the model failed to load DESPITE the readiness wait, so it is
     # a genuine load failure rather than the race it used to be confused with.
     if "no catalog" in low or "no model folder resolved" in low:
+        if not network_fault_observed:
+            return (
+                "ERROR",
+                "unclassified refresh/load failure: Power BI did not produce a model catalog, but "
+                "the probe did not observe a network fault. Do not report this as UNREACHABLE or "
+                "send anyone to fix server/http_path without stronger evidence. Raw: " + raw,
+            )
         return (
             "UNREACHABLE",
             "the probe model failed to load even after waiting - the data source did not resolve. "
             "Check server and http_path in the spec before treating this as a credential problem. "
-            "Raw: " + text[-200:],
+            "Raw: " + raw,
         )
     # Order matters: check BAD_TABLE before NO_CREDENTIAL. A "not found" message proves the server
     # answered us, so it can never be a credential problem, but the text often also mentions the
@@ -349,7 +365,11 @@ def _classify_failure(text: str) -> tuple[str, str]:  # noqa: PLR0911
         return "BAD_TABLE", text
     if any(marker in low for marker in CREDENTIAL_MARKERS):
         return "NO_CREDENTIAL", text
-    return "UNREACHABLE", text
+    return (
+        "ERROR",
+        "unclassified refresh failure: the probe ran, but the error did not match a credential, "
+        "bad-table, or observed network fault signature. Do not report it as UNREACHABLE. Raw: " + raw,
+    )
 
 
 def _npx(args: list[str], timeout: int) -> tuple[int, str]:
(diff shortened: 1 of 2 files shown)


Guust-Franssens:

Reviewed by a second agent that saw only the diff, the issues and the repo — deliberately **not** the
author's rationale, so this is not an echo of the PR body.

### Blocking

**1. Termination is still not fail-closed — #53 survives in a narrower form.**
`probe_live_source.py:473-481` returns the **first** exact-path match. Because every run uses the same
`_probe\Probe.pbip` (`:444-450`), two concurrent probes *of the same migration* are indistinguishable,
and cleanup then force-kills that PID without re-checking its path (`:397-402`, `:929-932`).
Demonstrated in-memory: two identical `currentFilePath` entries returned PID `333` rather than refusing.

Exact-path matching removed the *cross-migration* kill, which is the common case and a real improvement
— but the invariant we asked for was **fail closed when it cannot positively identify its own
instance**, and duplicate matches are precisely that case.
*Suggested*: a per-invocation unique PBIP path, require **exactly one** match, and revalidate PID +
path immediately before `Stop-Process`; otherwise kill nothing.

**2. A timeout is still classified as a final `NO_CREDENTIAL`.**
`:617-624` treats `TimeoutExpired` as missing credentials unconditionally, ignoring network evidence —
and the refresh helper it invokes explicitly documents that a timeout has both transient and sign-in
causes (`refresh_pbip_model.py:549-582`). Demonstrated: both `network_fault_observed=False` and `True`
returned `(1, "NO_CREDENTIAL")`.
This matters beyond tidiness: our convention is that a **missing credential is final and must not be
retried**, while a transient timeout earns one retry. Misclassifying a slow warehouse as a credential
wall converts a recoverable stall into a hard stop that asks a human for something they cannot supply.

**3. Permission failures are conflated with credentials or with unknown errors.**
`:97-110,364-371` maps `403 Forbidden` to `NO_CREDENTIAL`, while "permission denied" / "insufficient
privileges" fall through to a generic `ERROR`. The first gives wrong sign-in guidance; the second
allows retrying a refusal that is final. Suggests a distinct `ACCESS_DENIED` verdict.

### Non-blocking

- The sibling test (`tests/test_credential_gate.py:839-861`) is genuine — it constructs two instances,
  and an old/new comparison gave `new=222, old=111`. But it covers only `_pid_for_file`, not
  `_open_desktop`'s removed fallback nor the eventual kill. Worth adding: no exact match, duplicate
  exact matches, and cleanup refusing to terminate an unverified PID.
- `network_fault_observed=True` is an unsafe **default** (`:304,573`) — omitted evidence silently
  recreates #52. Make it required, or default to `False`.

### Verified good

- No newest-process, window-title, or table-name selection fallback remains anywhere in the file.
- A successful probe still requires a row refreshed **through Power BI** — the property that makes this
  script meaningful is intact.
- The `tests/test_credential_gate.py` changes are legitimate, not a test loosened to fit the fix.
- CI is green.

The reviewer did not open or terminate any Desktop instance (other agents hold live ones), so the
process-selection findings are static plus in-memory reproduction rather than a live kill test.

**Single most important check for a human:** simulate a duplicate/pre-existing exact `currentFilePath`
and confirm **no `Stop-Process` is issued at all**.

Guust-Franssens:
## Re-review after `47c864c` — **approve**

Same reviewer that raised the three blocking findings, re-running its own experiments against the new
head rather than reading the fix description.

1. **Termination is now fail-closed.** Unique per-run PBIP paths; duplicate matches return `None`; and
   `_close()` revalidates exact PID **and** path before terminating
   (`probe_live_source.py:410-424,458-513`). Experiment: duplicate bind → `None`, close → `False`,
   **`Stop-Process` calls = 0**, including on the timeout and error cleanup paths. No alternate-PID
   branch remains anywhere in the file.
2. **Timeouts no longer collapse into a credential verdict** (`:603-660`). A no-network timeout returns
   a **retryable** `ERROR`; an observed network fault returns `UNREACHABLE`; neither becomes
   `NO_CREDENTIAL`. An explicit sign-in refusal still produces a **final** `NO_CREDENTIAL` — so the
   operational distinction that matters (transient earns a retry, a credential wall does not) is intact
   rather than merely relabelled.
3. **`ACCESS_DENIED` is reachable from the real error strings** (`403 Forbidden`, "permission denied",
   "insufficient privileges"), and its directive explicitly forbids retry while asking for a permission
   grant (`:112-121,375-380,882-897`).

The non-blocking item is also gone: both network-evidence parameters are now **required**, with every
call site supplying them explicitly — so an omitted argument can no longer silently recreate #52.

No Power BI Desktop instance was opened or terminated during this review; findings are static analysis
plus in-memory reproduction.

Ready to merge from my side.

Issue #51, opened 2026-08-07
parse_tableau.py crashes when a dashboard's desktop layout has no zones (device-only layout)


Guust-Franssens:
## Summary
A dashboard whose **desktop layout has no zones** - because its layout is defined entirely under `<devicelayouts>` - makes `parse_tableau.py` emit `zones: {}` (an empty dict) instead of `zones: []`, which fails schema validation and aborts the whole parse.

## Reproduction
```xml
<dashboard name='Commissions'>
  <style />
  <size maxheight='800' maxwidth='1000' minheight='800' minwidth='1000' />
  <zones />                      <!-- dashboard-level, EMPTY -->
  <devicelayouts>
    <devicelayout auto-generated='true' name='Phone'>
      <zones><zone h='100000' id='13' type='layout-basic' w='100000' x='0' y='0'>...</zone></zones>
    </devicelayout>
  </devicelayouts>
</dashboard>
```

```
jsonschema.exceptions.ValidationError: 'type' is a required property
Failed validating 'required' in schema['properties']['dashboards']['items']['properties']['zones']
On instance['dashboards'][0]['zones']: {}
```

## Precise trigger (an obvious hypothesis that is WRONG)
An empty `<zones />` **is not sufficient** on its own. `PovertyandSchoolPerformance.twbx` contains one and parses fine, because there the empty element sits *inside* a `<devicelayout>` while the dashboard-level zones are populated. The crash requires the **dashboard-level** `<zones />` to be the empty one.

## Impact
Hard failure - no `migration-spec.json` is produced for the workbook, so one dashboard with a device-only layout blocks the entire migration. The other dashboard in the same file (`Sales Overview`) is perfectly normal.

## Frequency
Measured across 79 real workbooks (43 Tableau Public + 17 course + 19 book): **1 occurrence**. Rare, but device-specific layouts themselves are common (14/43, 10/17, 1/19), so the surrounding construct is mainstream.

## Reproducer
`chapter08/8-1-Dashboards.twbx` from https://github.com/aloth/tableau-book-resources (CC-BY-4.0, a genuine Tableau Desktop export that publishes cleanly to Tableau Cloud).

Guust-Franssens:
## Closing as already fixed — no code change was needed

Investigated as part of the #69 parser batch. **The reported crash does not reproduce on `master`.**

Verified twice, independently:

1. The agent assigned to fix it parsed the issue's XML, found it already parses and schema-validates,
[...]

Issue #49, opened 2026-08-07
Keyword-less LOD '{SUM([x])}' not flagged as an LOD (skips the high-severity grain warning)


Guust-Franssens:
## Summary
A Tableau LOD expression does **not** require a `FIXED` / `INCLUDE` / `EXCLUDE` keyword. `{SUM([Sales])}` is valid and means a table-scoped (grand-total) LOD. `_classify_calculation()` appears to key on the keywords, so this form is not flagged with `is_lod`.

## Reproduction
```xml
<column caption='Grand Total' datatype='real' name='[Grand Total]' role='measure' type='quantitative'>
  <calculation class='tableau' formula='{SUM([Sales])}' />
</column>
```
Parsed result: `is_lod` is not set.

## Impact
LODs carry a `high` severity limitation telling a human to verify grain/filter context. A keyword-less LOD skips that warning, so the riskiest class of calculation passes through unflagged - and a naive translation to plain `SUM()` changes the result, because the LOD is deliberately insensitive to view grain.

## Frequency
8/43 local real Tableau Public workbooks contain a `{ AGG(` form with no keyword.

## Suggested fix
Treat a formula whose braces contain an aggregate with no LOD keyword as a table-scoped LOD.

## Credit
Trap identified by studying https://github.com/cyphou/Tableau-To-PowerBI.

Guust-Franssens:
Fixed and merged in #69 (commit `e93e8cf`).

Keyword-less LODs are flagged again, and the detection is now **structural** rather than an allowlist. The first fix enumerated aggregate functions and so still missed `{CORR(...)}` — Tableau's own documented example — plus `PERCENTILE`, `COLLECT` and the `COVAR` family, each silently skipping the high-severity grain warning. Any allowlist would keep drifting behind Tableau's function set, so the match is now shape-based. Measured for noise: across **2,999 calculations in 16 committed workbooks** it adds only two extra matches, both genuine. Known follow-up: the matcher is syntax-unaware, so `{SUM(...)}` inside a comment or string literal would false-positive — no committed workbook does this, so it is tracked separately rather than scope-crept here.

The fix was reviewed independently by an agent that saw only the diff and the issue, not the author's reasoning — that review returned **request changes** on this PR, and the finding above is what it caught.

Issue #48, opened 2026-08-07
Bin fields (calculation class='bin') parsed as generic calculated fields, bin-size lost


Guust-Franssens:
## Summary
Tableau encodes a **bin** field as `<calculation class='bin'>`, not `class='tableau'`. `_build_field_entry()` only recognises the latter, so a bin is recorded as `kind: "calculated"` with no formula, and `bin-size` is lost entirely.

## Reproduction
```xml
<column caption='Sales (bin)' datatype='real' name='[Sales (bin)]' role='dimension' type='ordinal'>
  <calculation class='bin' column='[Sales]' bin-size='200' new-bin='true' peg='0' />
</column>
```

Parsed result: `kind='calculated'` (expected `kind='bin'`), and the 200-wide binning is silently dropped.

Note `docs/migration-spec.schema.json` **already declares** `kind: [... "bin" ...]` - so the schema anticipates this, but no code path emits it.

## Impact
The migrated model loses the binning entirely: a histogram/distribution visual would be rebuilt against a raw continuous measure. Silently wrong rather than failing.

## Frequency
10/43 local real Tableau Public workbooks contain `class='bin'`.

## Related
`kind: "group"` and `kind: "hierarchy"` are likewise declared in the schema; worth checking whether either is ever emitted.

## Credit
Trap identified by studying https://github.com/cyphou/Tableau-To-PowerBI.

Guust-Franssens:
Fixed and merged in #69 (commit `e93e8cf`).

Bin fields now preserve their semantics — including the **modern Tableau 2024.3 encodings**, which the first attempt missed. An independent review found that `formula='...' size='0.1'` and `size-parameter='[Parameters].[Parameter 2]'` still produced `bin_size: null`, because only the older `bin-size`/`column` attributes were read. Now verified against the two committed 2024.3 workbooks: `bin_size='0.1'` with source `[Pivot Field Values]`, and source `[Profit]` with `bin_size_parameter`, both schema-validating against the new field. Covering test: `test_bin_field_preserves_bin_class_and_size`.

The fix was reviewed independently by an agent that saw only the diff and the issue, not the author's reasoning — that review returned **request changes** on this PR, and the finding above is what it caught.

Issue #47, opened 2026-08-07
Tableau soft line-break sentinels (U+00C6 / U+00A0) leak into parsed titles and tooltips


Guust-Franssens:
## Summary
Tableau Desktop emits **U+00C6 (AE)** - and sometimes **U+00A0** - as *invisible soft line-break sentinels* inside `<formatted-text>` runs. `_text_from_runs()` concatenates run text verbatim, so these characters leak into `title_text` and `customized_tooltip_text` in the parsed spec, and would render as a stray character in a migrated Power BI title/tooltip.

## Evidence
```xml
<formatted-text>
  <run fontsize='12'>Revenue by Region</run>
  <run>&#198;
</run>
  <run fontsize='9'>FY1998</run>
</formatted-text>
```

Parsing `_harvest/twbx/NFLLeadersandDynamicZoneZooming.twbx` (a real Tableau Public export) leaks the sentinel into **6 worksheet tooltips**. It also appears in `RESTAPISample.twb` from `tableau/server-client-python` (5 occurrences).

## Frequency
**43/43** local real Tableau Public workbooks contain U+00C6; 16/43 contain U+00A0. This is effectively universal.

## Why it is easy to miss
The character is invisible in Tableau itself, so nobody notices until the migrated report renders it.

## Suggested fix
Strip runs matching `^[\s]*[\u00c6\u00a0]+[\s]*$` (treating them as line breaks) in `_text_from_runs()`.

## Credit
Behaviour documented in https://github.com/cyphou/Tableau-To-PowerBI (`_TABLEAU_LB_SENTINEL_RE` in `tableau_export/extract_tableau_data.py`).

Guust-Franssens:
Fixed and merged in #69 (commit `e93e8cf`).

Soft line-break sentinels (U+00C6 / U+00A0) no longer leak into titles and tooltips, **and** the same normalization now covers dashboard text zones — an independent review found the first fix had missed `:970-972`, leaving two `text_html` values in the real Airline Alliance workbook still carrying U+00C6. Re-verified: zero leaks remain, while a legitimately embedded NBSP and a real `AE` survive untouched. Covering test: `test_formatted_text_soft_line_break_sentinels_do_not_leak`.

The fix was reviewed independently by an agent that saw only the diff and the issue, not the author's reasoning — that review returned **request changes** on this PR, and the finding above is what it caught.

PR #69, opened 2026-08-09
Harden Tableau parser edge cases

Guust-Franssens:
## Summary

Hardens `scripts/parse_tableau.py` against four parser defects that either crashed schema validation or silently produced an incomplete migration spec. Also records the #51 finding: the device-only/empty dashboard shape is already handled on this branch baseline and was not changed.

### #46 — fixed: top-N filters crashed schema validation

**What was wrong:** the parser already emitted Tableau's raw filter class (`topn`), but `docs/migration-spec.schema.json` did not allow that enum value. That made a normal Tableau top-N filter a hard parse failure. The parser also discarded `direction` and `max`, so even if validation were relaxed the report builder would not know whether to recreate top/bottom N or what N was.

**Change:** schema now accepts `topn`; worksheet filter entries carry raw `direction` and `max`.

**Covering test:** `test_topn_filter_carries_limit_and_validates_against_schema`. I watched it fail before the fix with `KeyError: 'direction'`.

### #47 — fixed: Tableau soft line-break sentinels leaked into text

**What was wrong:** Tableau writes standalone U+00C6 and U+00A0 runs as invisible formatted-text line-break sentinels. `_text_from_runs()` concatenated every run verbatim, so migrated titles/tooltips could contain stray invisible/garbage characters. This was silent: parsing and validation succeeded, but the rendered Power BI text would be wrong.

**Change:** sentinel-only runs are treated as line breaks and not emitted as literal characters; ordinary run text remains untouched.

**Covering test:** `test_formatted_text_soft_line_break_sentinels_do_not_leak`. I watched it fail before the fix with `Revenue by RegionÆ\nFY1998` leaking into `title_text`.

### #48 — fixed: bin fields were treated as ordinary columns

**What was wrong:** Tableau histogram bins serialize as `<calculation class='bin' ... bin-size='...'>` without a normal `formula`. `_build_field_entry()` only classified formula-backed calculations, so bins silently became `kind: column` and lost their bin width/source column. Parsing and schema validation could still succeed while the semantic meaning was gone.

**Change:** `class='bin'` is classified as `kind: bin`; the raw `bin-size` and source `column` are preserved as `bin_size` and `bin_source_column`, and the schema documents those optional properties.

**Covering test:** `test_bin_field_preserves_bin_class_and_size`. I watched it fail before the fix with `kind == 'column'` instead of `bin`.

### #49 — fixed: keyword-less aggregate LODs skipped the high-severity warning

**What was wrong:** the LOD detector only looked for `{ FIXED ...`, `{ INCLUDE ...`, or `{ EXCLUDE ...`. Tableau also supports table-scoped aggregate LODs like `{SUM([Sales])}`. Those parsed and validated, but silently avoided the high-severity grain/filter-context limitation that downstream humans rely on.

**Change:** braced aggregate calls (`SUM`, `AVG`, `COUNT`, `COUNTD`, `MIN`, `MAX`, `ATTR`, `MEDIAN`, `STDEV`, `STDEVP`, `VAR`, `VARP`, `AGG`) are now classified as LODs even without a keyword.

**Covering test:** `test_keywordless_aggregate_lod_is_flagged_high_severity`. I watched it fail before the fix with `is_lod` false.

### #51 — not fixed here: already handled

**Finding:** the reported crash shape is already fixed on this branch baseline. A dashboard-level self-closing `<zones />` now produces a valid empty `layout-basic` root with empty `children`, and schema validation passes. Existing regression coverage is `test_empty_dashboard_survives_parse_and_validates_against_schema`, which asserts both the valid zone tree and the parse-stage empty-dashboard limitation. I also manually verified a minimal device-only dashboard with dashboard-level `<zones />` plus populated `<devicelayouts>` parses and validates.

## Deliberately not changed

- Did not add a fifth commit for #51 because the code path and regression test already exist; inventing a no-op change would make review harder.
- Did not broaden filter modeling beyond top-N `direction`/`max`; other Tableau filter classes and richer top-N semantics remain out of scope.
- Did not normalize sentinel characters outside standalone sentinel-only formatted-text runs; embedded U+00C6/U+00A0 inside real text may be intentional data/text and should not be stripped blindly.
- Did not resolve bin source columns to field ids; this preserves the raw Tableau XML attribute and leaves semantic resolution to downstream builders.
- Did not parse device layouts themselves for #51; the existing behavior is only to keep the dashboard parse/schema-valid when desktop zones are empty.

## Gate results

- `ruff format scripts\parse_tableau.py tests\test_parse_tableau.py` + `ruff check scripts\parse_tableau.py tests\test_parse_tableau.py --fix`: exit 0, all checks passed.
- `python -m pylint scripts`: exit 0, rated 10.00/10.
- `python -m pytest -q`: exit 0, 508 passed, 4 skipped in 67.53s (508 collected baseline; skipped live fixture tests are expected).

The main review point is #49's aggregate list for keyword-less LOD detection: it is intentionally conservative and Tableau-aggregate-oriented, but reviewers should confirm no aggregate spelling important to this corpus is missing. For #48, review whether preserving `bin_size` as the raw string is preferable to numeric coercion; I kept it raw to match XML and avoid changing parser-wide typing conventions.

### DIFF
--- a/docs/migration-spec.schema.json
+++ b/docs/migration-spec.schema.json
@@ -132,6 +132,8 @@
                   "additionalProperties": { "type": "string" }
                 },
                 "tableau_formula": { "type": ["string", "null"], "description": "Raw formula text, required when kind=calculated" },
+                "bin_size": { "type": ["string", "number", "null"], "description": "Raw Tableau calculation@bin-size when kind=bin" },
+                "bin_source_column": { "type": ["string", "null"], "description": "Raw Tableau calculation@column that the bin groups when kind=bin" },
                 "referenced_fields": { "type": "array", "items": { "type": "string" }, "description": "field ids this calculation depends on" },
                 "is_lod": { "type": "boolean", "default": false },
                 "is_table_calc": { "type": "boolean", "default": false },
--- a/docs/migration-spec.schema.json
+++ b/docs/migration-spec.schema.json
@@ -134,6 +134,7 @@
                 "tableau_formula": { "type": ["string", "null"], "description": "Raw formula text, required when kind=calculated" },
                 "bin_size": { "type": ["string", "number", "null"], "description": "Raw Tableau calculation@bin-size when kind=bin" },
                 "bin_source_column": { "type": ["string", "null"], "description": "Raw Tableau calculation@column that the bin groups when kind=bin" },
+                "bin_size_parameter": { "type": ["string", "null"], "description": "Raw Tableau calculation@size-parameter when kind=bin and bin width is controlled by a parameter" },
                 "referenced_fields": { "type": "array", "items": { "type": "string" }, "description": "field ids this calculation depends on" },
                 "is_lod": { "type": "boolean", "default": false },
                 "is_table_calc": { "type": "boolean", "default": false },
--- a/docs/migration-spec.schema.json
+++ b/docs/migration-spec.schema.json
@@ -190,7 +190,9 @@
               "type": "object",
               "properties": {
                 "field_id": { "type": "string" },
-                "type": { "type": "string", "enum": ["categorical", "quantitative", "relative_date", "relative-date", "parameter_binding", "parameter-binding", "context"] },
+                "type": { "type": "string", "enum": ["categorical", "quantitative", "relative_date", "relative-date", "parameter_binding", "parameter-binding", "context", "topn"] },
+                "direction": { "type": ["string", "null"], "enum": ["top", "bottom", null] },
+                "max": { "type": ["string", "number", "null"] },
                 "exclude_nulls": { "type": "boolean", "default": false },
                 "members": { "type": "array", "items": { "type": "string" } },
                 "note": { "type": ["string", "null"], "description": "e.g. 'parameter-equality idiom: field = IF [Param]=[Dim] THEN [Dim] END, filtered to exclude null -> collapses to a plain PBI slicer on the dimension, no calculated column needed'" }
--- a/scripts/parse_tableau.py
+++ b/scripts/parse_tableau.py
@@ -39,10 +39,7 @@ logger = logging.getLogger("parse_tableau")
 SPEC_VERSION = "1.0"
 
 _LOD_KEYWORD_RE = re.compile(r"\{\s*(FIXED|INCLUDE|EXCLUDE)\b", re.IGNORECASE)
-_KEYWORDLESS_LOD_RE = re.compile(
-    r"\{[^{}]*\b(SUM|AVG|COUNT|COUNTD|MIN|MAX|ATTR|MEDIAN|STDEV|STDEVP|VAR|VARP|AGG)\s*\(",
-    re.IGNORECASE,
-)
+_KEYWORDLESS_LOD_RE = re.compile(r"\{[^{}]*\b[A-Z][A-Z0-9_]*\s*\(", re.IGNORECASE)
 _TABLE_CALC_RE = re.compile(r"\b(WINDOW_\w+|RUNNING_\w+|INDEX|RANK\w*|LOOKUP|TOTAL|PREVIOUS_VALUE)\s*\(", re.IGNORECASE)
 _PARAM_EQUALITY_RE = re.compile(r"if\s*\[Parameters\]\.\[[^\]]+\]\s*=\s*\[[^\]]+\]\s*then", re.IGNORECASE)
 _BRACKET_TOKEN_RE = re.compile(r"\[([^\[\]]+)\]")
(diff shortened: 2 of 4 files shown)


Guust-Franssens:

Reviewed by a second agent given only the diff, the issues and the repo — not the author's rationale.

**Verified good first.** The PR suite is **520 passed**, all 16 packaged example `.twbx` files
schema-validate, and running the four new cases through the **base** parser confirms every one genuinely
fails old behaviour. The fixture's XML shapes match the issues and its title mutation did not weaken an
existing assertion.

**#46 is correct** — the cited `Financial_Report.twb` parses and schema-validates with `type='topn'`,
`direction='top'`, `max='10'`, and the parser/schema spellings match (the runtime-only failure mode I
was worried about).

**#51's "already fixed" claim is confirmed** — checked against the issue's actual `8-1-Dashboards.twbx`:
master parses and schema-validates with zero desktop zones and one device zone. I had flagged this as
the claim deserving the most scepticism, since "no change needed" is the cheapest possible answer. It
holds.

### Blocking

**1. #48 is still broken for modern Tableau bin XML.** `parse_tableau.py:473-475` reads only `bin-size`
and `column`. Two **committed** Tableau 2024.3 workbooks instead use:

```
formula='[Pivot Field Values]'  size='0.1'
formula='[Profit]'             size-parameter='[Parameters].[Parameter 2]'
```

PR output classifies both as `kind: bin` but emits `bin_size: null` and `bin_source_column: null` — so
the bin semantics the issue is about are still lost. The test (`tests/test_parse_tableau.py:47-54`)
covers only the convenient quoted variant.

This is the specific failure mode worth naming: **a fix that satisfies a synthetic fixture but not the
real XML already in this repo.** Suggested: support `size`/`formula` fallbacks, preserve
`size-parameter` in parser **and** schema, and test both real variants.

**2. #49's aggregate allowlist misses valid keyword-less LODs.** `:42-45` omits `CORR`, `COVAR`,
`COVARP`, `PERCENTILE`, `COLLECT`. Tableau's own documentation uses `{CORR(Sales, Profit)}` as the
canonical table-scoped LOD example. Against PR head:

```
{CORR([Sales], [Profit])}    => False
{PERCENTILE([Sales], 0.95)}  => False
{COLLECT([Geometry])}        => False
```

These still skip the high-severity grain warning — meaning a downstream agent never learns the
calculation needs grain verification, which is the entire point of #49. Suggested: detect table-scoped
LOD **structurally** rather than via a partial function allowlist, and add a `{CORR(...)}` regression
test.

### Non-blocking

**#47 normalization is not applied to dashboard text zones.** `:970-972` still joins runs verbatim
instead of using `_text_from_runs()`; PR head leaves two `text_html` values containing U+00C6 in the
real Airline Alliance workbook. The title/tooltip paths *are* fixed, and legitimate embedded characters
are correctly left untouched.

**Single most important check for a human:** validate bin extraction against this repository's real
Tableau 2024.3 `size` / `size-parameter` encodings, not only the synthetic `bin-size` fixture.

Guust-Franssens:
## Re-review of the delta — **approve with follow-ups**

Both blocking findings closed, verified against the **real committed workbooks** rather than the fixture.

**#48 closed** (`parse_tableau.py:470-473`) — both modern forms now handled. The two Tableau 2024.3
workbooks emit:
- `bin_size='0.1'`, source `[Pivot Field Values]`
- source `[Profit]`, `bin_size_parameter='[Parameters].[Parameter 2]'`

Both **schema-validate** against the new field (`migration-spec.schema.json:137`), so the
parser/schema mismatch that would only fail at runtime is not present. The fixture
(`tests/fixtures/minimal.twb:39-43`) matches the real attribute shapes rather than a convenient
simplification.

**#49 closed for under-detection** — end-to-end parsing now raises high-severity LOD warnings for
`CORR`, `PERCENTILE` and `COLLECT`, while `SUM([Sales])*100` stays non-LOD.

**On the over-matching risk I asked about**: measured across **2,999 calculations in 16 committed
workbooks**, the structural regex added only **two** extra matches — both genuine braced references to
an aggregate `AVG(...)` field. So the noise concern does not materialise in practice, which is the
answer that matters; the warning stays meaningful.

**#47 closed** — zero U+00C6 leaks remain in the real Airline workbook's dashboard text, and embedded
NBSP and `Æ` survive untouched. The trade-off is handled correctly rather than over-normalized.

**#46 and #51 unchanged** — the topn reproducer is still schema-valid with `topn/top/10`, and the
device-only workbook still parses on both head and master.

### Follow-up (non-blocking, deliberately not fixed here)

`parse_tableau.py:42` is **syntax-unaware**: a synthetic formula containing `{SUM(...)}` **inside a
comment or string literal** is falsely flagged. **No committed workbook exhibits this**, which is why it
is a follow-up rather than a blocker — stripping comments and string literals before matching would
harden it. Worth an issue rather than scope creep in this PR.

Suite: **522 passed** locally.

PR #70, opened 2026-08-09
Fix preflight exit tiers and dangling hook config

### LINKED ISSUE #43: Tracked hook config references a script that does not exist (subagent-context.json)
## The problem

`.github/hooks/subagent-context.json` is tracked and active, and points at a script that does not exist in the repo:

```json
"subagentStart": [{
  "matcher": "pbi-migration-validator",
  "powershell": "powershell ... -File scripts/hooks/probe_subagent_start.ps1",
  "bash": "bash scripts/hooks/probe_subagent_start.sh"
}]
```

Neither `scripts/hooks/probe_subagent_start.ps1` nor `.sh` exists on `master`.

## Not merge damage — pre-existing

Verified this predates the credential-gate merge (57585fa): the script is absent at `4acc50f` (the pre-merge branch head) as well. It was gitignored as ad-hoc scaffolding in `c6dfeb2` ("Keep ad-hoc test scaffolding out of the public repo") while the **tracked** config that references it was left in place.

So every clone of this repo has an active hook pointing at a missing file.

## Why it should be removed rather than restored

The script (recoverable from `c6dfeb2^`) is a **research probe**, not production tooling:

```powershell
$ctx = "HOOK_SENTINEL_ORBIT_58231 - injected by subagentStart at ..."
@{ additionalContext = $ctx } | ConvertTo-Json -Compress
```

It injects a sentinel token into the context of **every** `pbi-migration-validator` subagent and appends every raw payload to `_hook_probe.log`. That experiment has concluded — its finding is documented in `docs/agent-architecture.md` §6 (subagentStart fires and `additionalContext` reaches the subagent).

Restoring it would inject a meaningless sentinel into every validator run and silently grow a log file.

## Plan

1. Remove `.github/hooks/subagent-context.json`, **or** keep the file and strip the `subagentStart` entry if the hook config is wanted as a scaffold for future use.
2. Remove the `scripts/hooks/probe_subagent_start.ps1` row from `scripts/README.md` if the script is not coming back.
3. Consider a guard: a test asserting every command referenced by a tracked `.github/hooks/*.json` actually exists on disk. This class of defect — tracked config pointing at a gitignored file — is invisible until the hook fires.

## Acceptance criteria

- [ ] no tracked hook config references a non-existent script
- [ ] `scripts/README.md` and `tests/test_repo_layout.py` stay consistent
- [ ] (optional but recommended) a test that would have caught this

### LINKED ISSUE #54: preflight exits 1 for a RECOMMENDED item whose hint says it is handled - and it is the one that stops the run
Reported independently by BOTH agents that ran a live-source path (E2E-3 and E2E-4), which is why it is worth fixing rather than remembering.

`preflight.ps1` reports:

```
Power BI Desktop  [OK]  ... (via MSIX discovery)
PBI_DESKTOP_PATH  [MISS] "not set - the bridge is using its own version-pinned discovery."
```

...and **exits 1**.

Three problems compound:

1. **The hint says it is handled. It is not.** The bridge died with `DESKTOP_EXE_NOT_FOUND` on exactly this. Its discovery is *version-pinned*, so a Power BI Desktop auto-update (measured today: `2.157.480 -> 2.157.627`) moves the versioned WindowsApps path out from under it. The wording "the bridge is using its own discovery" reads as reassurance for the one item that will actually stop the run.
2. **Exit 1 for a RECOMMENDED item makes the exit code unusable.** The personas say "if preflight exits non-zero, stop and surface the missing items." Taken literally that aborts every run over a cosmetic nit — so in practice the rule trains everyone to ignore preflight, and here ignoring it cost an agent one of its three probe attempts.
3. **It already knows the answer.** Preflight performed MSIX discovery and printed the resolved path. It could simply export it.

## Suggested fix

- Exit non-zero **only** for genuinely blocking items.
- Reclassify `PBI_DESKTOP_PATH` as blocking *or* have preflight set it from the discovery it just did.
- Change the hint to say what happens without it: "the Desktop bridge will fail with DESKTOP_EXE_NOT_FOUND after any Desktop auto-update."

## Wider point

This is the second instance today of an **unpinned dependency that changes with no repo diff** breaking a gate — the other was a new `pylint` release adding `W4903` and turning CI red. Worth a general note in the preflight prose: the things most likely to break are the ones with no version in the repo.

Guust-Franssens:
Fixes #54
Fixes #43

## Summary

Fixes two tooling traps in the preflight/hooks path and addresses the review findings that warning-only exit semantics exposed stale severity tiers.

### #54 — fixed

`preflight.ps1` rendered three tiers (`CRITICAL`, `RECOMMENDED`, `OPTIONAL`) but its old renderer counted `CRITICAL` and `RECOMMENDED` failures together. That made advisory drift indistinguishable from a missing critical dependency to migration agents, which stop on preflight exit 1.

The renderer is now deterministic: exit 1 is reserved for missing `CRITICAL` checks; `RECOMMENDED` renders `[WARN]` and does not halt. The important follow-up was auditing which checks are actually blockers under those semantics. A check is now treated as blocking if any persona Definition of Done depends on it, even if the failure would otherwise surface later at handoff.

Known blockers are explicitly `CRITICAL`:
- stale/missing shipped skill bundles (`STALE in plugin` / `NOT INSTALLED`), because agents would execute different code from the repo or lack a required skill;
- `npx`, `powerbi-desktop` CLI, Power BI Desktop itself, and `PBI_DESKTOP_PATH`, because persisted PBIP refresh/bridge verification depends on them and an unset pin can hit `DESKTOP_EXE_NOT_FOUND` after Desktop auto-update;
- `dotnet`, because offline `tmdl_validate` deserialization is in the semantic-builder Definition of Done;
- below-floor Power BI CLI versions, because those are correctness failures.

Remaining `RECOMMENDED` checks were audited and left warning-only because they have documented fallbacks or are advisory: the migration-skills plugin itself in this repo (repo-local skills still load), `powerbi-modeling-mcp`, and Desktop version drift.

Covering verification:
- Manual verification, not a full automated PowerShell harness: with `PBI_DESKTOP_PATH` unset, preflight reports it under `CRITICAL` and exits `EXIT_CODE=1`.
- Manual verification: with a stale installed skill bundle, preflight reports `skill bundles match published plugin` under `CRITICAL` and exits `EXIT_CODE=1`.
- Manual verification in this environment: current preflight reports stale bundle under `CRITICAL` and exits `EXIT_CODE=1`; `npx` and `dotnet` are listed under `CRITICAL` when present.
- Source-contract test: `tests/test_preflight_contract.py::test_known_blocking_preflight_checks_are_critical` asserts the named blocking checks are tiered `critical`. This test was added after the review fix, so I did not watch it fail against the pre-fix tree; the reviewer later mutation-tested the first version, and this update extends the same gate to `npx` and `dotnet`.

### #43 — fixed

A tracked hook config, `.github/hooks/subagent-context.json`, referenced `scripts/hooks/probe_subagent_start.ps1` and `.sh`, but those scripts were not tracked. The issue evidence showed the removed script was research scaffolding: it injected a sentinel into `pbi-migration-validator` subagents and wrote probe logs. Restoring it would reintroduce experiment behavior into normal validator runs, so the correct fix was to remove the dangling active hook config.

Covering test:
- `tests/test_repo_layout.py::test_tracked_hook_configs_do_not_point_at_missing_scripts`
- This test was written after the fix. The independent review verified it catches both missing filenames against the pre-fix tree.

## Deliberately not changed

- I did not make preflight ignore all failures; critical blockers still exit 1.
- I did not downgrade stale skill-bundle checks to optional/recommended; they are critical because stale plugin code has already invalidated a measurement in this repo.
- I did not restore the subagent probe scripts because they were temporary measurement scaffolding, not production hooks.
- I did not pre-resolve expected `scripts/README.md` conflicts with sibling branches.

## Things to review hardest

- The audited boundary between `CRITICAL` and remaining `RECOMMENDED` checks in `scripts/preflight.ps1`.
- The source-contract test is intentionally a regex over `preflight.ps1`, not a full PowerShell simulation, because CI is Ubuntu and the script intentionally uses Windows/Power BI Desktop primitives.

## Gate results

- `python -m ruff format --check tests\test_preflight_contract.py tests\test_repo_layout.py`: clean
- `python -m ruff check tests\test_preflight_contract.py tests\test_repo_layout.py`: clean
- `python -m pytest -q`: 506 passed, 4 skipped in 75.47s
  - Corrected baseline note: base has 508 collected / 4 skipped. This branch adds two tests, so 510 total collected is expected.

## Additional manual checks

- `scripts/preflight.ps1` with current stale bundle: `EXIT_CODE=1` and `skill bundles match published plugin` appears in `CRITICAL`.
- `scripts/preflight.ps1` with `PBI_DESKTOP_PATH` unset: `EXIT_CODE=1` and `PBI_DESKTOP_PATH (bridge exe pin)` appears in `CRITICAL`.
- `git grep -n "subagent-context\|probe_subagent_start" -- .`: no references.


Guust-Franssens:

Reviewed by a second agent given only the diff, the issues and the repo — not the author's rationale.

The tier→exit mapping itself is now **deterministic and correct** (critical+recommended → 1;
recommended-only → 0; optional-only → 0), and no check was moved *down* a tier — two CLI version checks
correctly moved **RECOMMENDED → CRITICAL**. #43 is correctly handled, and the new test in
`test_repo_layout.py:282-317` was demonstrated to catch **both** missing filenames when run against the
pre-fix tree, so it is generic rather than hard-coded to the file just deleted.

The problem is what stayed in `recommended`.

### Blocking

**1. A known blocker now exits 0.** `PBI_DESKTOP_PATH` remains `recommended`
(`preflight.ps1:327`) even though lines 287-298 of the same file document that an unset pin causes
`DESKTOP_EXE_NOT_FOUND`. Controlled run with the variable unset:

| | result |
|---|---|
| PR head | `[WARN]`, **"Ready to migrate"**, exit **0** |
| master | exit **1** |

So this PR converts a caught failure into a green light on the exact condition the script warns about.

**2. Skill-bundle drift was behaviourally downgraded.** Both bundle checks remain `recommended`
(`preflight.ps1:220-225`), so a hash mismatch now renders `STALE in plugin` **and returns 0**
(demonstrated with a controlled mismatch).

This contradicts a documented contract — `AGENTS.md:148-156` requires **both** `STALE` and
`NOT INSTALLED` to block — and the reason it blocks is not tidiness: a stale bundle means an agent
executes different code from the one you are reading, which has already invalidated a measurement in
this repo once. That is why it is a block and not a warning.

### The shape of the fix

The diagnosis in #54 was right and the renderer rewrite is good. The gap is that **"which checks are
truly blocking" was never re-derived** — the new semantics were applied over the existing tier
assignments, and two checks that had been *effectively* blocking (because everything exited 1) silently
became advisory.

Suggested: classify `PBI_DESKTOP_PATH` and both bundle checks as `critical`, then keep
recommended-as-warning. Worth adding a **source-contract test** asserting the named blocking checks are
tiered `critical` — cheap, and it would have caught this. (Full automation of the script is awkward:
CI is Ubuntu, `checks.yml:16-18`.)

### Honesty note, in the PR's favour

The PR explicitly states that #54's verification is **manual**, rather than implying automated coverage.
That is the right call and I would rather have it than a confident claim.

**Single most important check for a human:** audit every remaining `RECOMMENDED` item under the new exit
semantics and confirm each true blocker still exits 1 — Desktop pinning and bundle drift first.

Guust-Franssens:
## Re-review after `00b0ba0` — **request changes** (round 3)

The two original blockers are **fixed**, verified by re-running the controlled experiments rather than
reading the tier table:

| experiment | result |
|---|---|
| unset `PBI_DESKTOP_PATH` | `[MISS]`, exit **1** |
| injected bundle hash mismatch | `STALE in plugin`, `[MISS]`, exit **1** |
| exit mapping | total — all-pass 0, recommended-only 0, optional-only 0, critical+recommended 1 |
| contract test | passes, **and mutation testing caught all five** asserted checks when flipped back to `recommended` |

That last line matters: the new test can actually fail, so it is a gate rather than decoration.
No check moved downward; the below-floor CLI versions remain `critical` (`:134`). #43 stays correct,
with no dangling references and a `.gitignore` change that removes only the exact obsolete rule.

### Still blocking — the sweep reached two wrong conclusions

The `RECOMMENDED` audit was **documented**, but two items classified as non-blocking are true blockers:

**1. `npx` is warning-only** (`preflight.ps1:245`). It is a stated full dependency of the
`pbip-model-refresh` skill (`SKILL.md:11-13`), whose persisted refresh is a **mandatory handoff gate**
(`pbi-semantic-builder.agent.md:281-285`) and is invoked at `refresh_pbip_model.py:222`. Demonstrated
with `npx` removed: `_bridge_status()` returned `{}` and `cache_file()` returned `None`, so the model
never persists. Preflight would say "Ready to migrate" and the handoff gate would then fail with no
warm cache — exactly the empty-model race this repo already documented.

**2. `dotnet` is warning-only** (`:362`). The semantic-builder Definition of Done requires offline
`tmdl_validate` deserialization (`pbi-semantic-builder.agent.md:239-242,279-280`). Demonstrated with
`dotnet` off PATH: `[WARN]`, **"Ready to migrate"**, exit **0**.

Both should be `critical` and added to the source-contract test.

The reviewer explicitly cleared the rest — migration-skills plugin, modeling MCP, and Desktop version
drift are **legitimately** non-blocking, because each has a documented fallback. So this is a
two-item correction to the audit, not a rejection of it.

### Why this round was worth running

The first review found two symptoms. The instruction for round 2 was that **the sweep is the real
deliverable**, because the tiers had never been re-derived when the exit semantics changed. That turned
out to be exactly right: the sweep was performed and still mis-classified two dependencies that the
personas *require*. Reclassifying only the named symptoms would have shipped a preflight that green-lights
a migration which cannot persist a model or validate TMDL.

Guust-Franssens:
## Final delta after `53bb832` — **approve**

Clean. Verified by re-running every experiment rather than reading tier assignments.

| check | result |
|---|---|
| `npx` isolated as missing | `[MISS] cli: npx`, exit **1** |
| `dotnet` removed from PATH | `[MISS] cli: dotnet`, exit **1** |
| mutation: flip each back to `recommended` | **both** fail the contract test |
| unset `PBI_DESKTOP_PATH` (round 1 finding) | exit **1** |
| forced bundle hash mismatch (round 1 finding) | exit **1** |
| exit mapping | total — all-pass 0, recommended-only 0, optional-only 0, critical+recommended 1 |

No tier moved downward. The cleared `recommended` set is unchanged (migration-skills plugin, modeling
MCP, Desktop version drift). #43 remains resolved, hook regression passes, and `.gitignore` still
removes only the obsolete exact rule.

**On over-strictness**, which I asked about explicitly since three rounds had all pushed the same
direction: none found. `preflight.ps1` is by design the **Windows-only, full-migration** gate — not a
Linux CI prerequisite and not a parser-only one — so the seven critical checks are all genuinely
required for the workflow it guards.

`Fixes #54` and `Fixes #43` are present, so both issues close on merge. CI green.

---

### Four rounds, and each one found something real

Worth recording, because the root cause was never a coding error:

1. exit code did not match severity tier → the reported bug
2. the tier *assignments* had never been re-derived when the semantics changed, so two known blockers
   silently became advisory
3. the sweep that was supposed to catch that reached **two wrong conclusions** — `npx` and `dotnet`,
   both hard dependencies of gates other personas are required to pass
4. confirmation, plus a check that the correction had not overshot

The generalisation that came out of it: **a check is blocking if any persona's Definition of Done
depends on it** — visible only by tracing from the personas' DoD back to the tooling, not by reading
the preflight list on its own.

Issue #65, opened 2026-08-09
screenshot-all captures a slow azureMap page BLANK - --settle only delays the first capture, so a correct page looks like a regression you just caused


Guust-Franssens:
`powerbi-desktop screenshot-all` walks pages at roughly **2.6 s each** and its only timing knob is
`--settle <ms>`, documented as *"Delay before **first** capture"* — so it waits once, then races through
every remaining page at full speed.

That is fine for a bar chart. It is not fine for a visual that does real work on activation.

## What it cost

During the `book_6-1-Maps` repair pass, `Combined Map` — an `azureMap` that fetches a **remote boundary
GeoJSON** and then lays out **604 pie marks** — captured **completely blank**. The page was correct;
the capture simply happened before it finished rendering.

The agent nearly logged it as a regression it had just introduced. A single `powerbi-desktop screenshot`
of the same page after a ~20 s settle rendered it perfectly.

**This is the worst kind of tooling trap**: it produces a plausible artifact (a real PNG of a real page)
that supports a false conclusion, and the natural reaction — "my last change broke this" — sends you
debugging correct code. It is the visual-verification analogue of the structural-validation problem
this repo keeps hitting.

It also lands precisely where we are least able to absorb it: **render verification is the step that
catches what `validate` cannot**, so a silently-wrong capture undermines the one check that was working.

## Verified

```
$ powerbi-desktop screenshot-all --help
  --scale <number>          Capture scale, from 1 to 3 (default: "2")
  --settle <ms>             Delay before first capture, in milliseconds
  --wait-seconds <seconds>  Wait/retry budget for bridge readiness and retryable errors

$ powerbi-desktop screenshot --help
  --wait-seconds <seconds>  Wait/retry budget for bridge readiness and retryable errors
```

Neither verb offers a **per-page** settle. Bridge CLI `0.1.2`.

## Proposed

- [ ] **Promote the wrapper into `scripts/`.** The repair pass already wrote
      `_build/capture_pages.py` (per-page settle, one `screenshot` call each) inside a scratch bundle,
      where the next migration will not find it. It belongs next to the other verification tools with a
      `scripts/README.md` row.
- [ ] **Record it in `powerbi-report-gotchas` §7** (Desktop verification mechanics), stated as the rule
      that generalises: *a blank capture is not evidence of a blank page — re-capture that page alone
      with a settle before believing it.* §7 already carries the sibling rule that a clean screenshot
[...]

PR #71, opened 2026-08-10
Harvest Azure Maps render fixes

Guust-Franssens:
## What was wrong
- `powerbi-desktop screenshot-all --settle` can capture progressive `azureMap` pages before all marks arrive; sparse maps can look plausible and silently under-count marks.
- A single equal screenshot pair is only a plateau, not completion, so the helper now requires a configurable byte-identical dwell (`--stable-seconds`, default 20).
- The dwell excludes time spent inside the blocking screenshot call; only idle time between completed frame samples counts.
- `azureMap` validates with multiple `Column` projections in `Category`, but that builds a drill hierarchy and usually collapses Tableau's finest-grain map.

## What changed
- Added `scripts/capture_powerbi_pages.py`, generalized from the scratch bundle to take `<report.Report>` and `<output-dir>` plus `--pid`, `--poll`, `--stable-seconds`, and `--max-wait`.
- Added offline tests for partial-plateau avoidance, slow-screenshot dwell accounting, never-converged captures, process-level non-zero exit on instability, and zero-page fail-closed behavior.
- Extended `scripts/README.md` and `powerbi-report-gotchas` §7 so the capture trap is findable, while explicitly labeling the dwell as a heuristic.
- Added one terse `pbi-report-builder` Definition-of-Done blocker for multi-column `azureMap` `Category` wells.

## Covering tests
- `tests/test_capture_powerbi_pages.py::test_capture_stable_ignores_a_partial_plateau_until_final_frame_dwells`
- `tests/test_capture_powerbi_pages.py::test_capture_stable_does_not_count_screenshot_duration_as_dwell`
- `tests/test_capture_powerbi_pages.py::test_capture_stable_flags_newest_frame_when_page_never_converges`
- `tests/test_capture_powerbi_pages.py::test_capture_report_exits_nonzero_when_any_page_is_unstable`
- `tests/test_capture_powerbi_pages.py::test_capture_report_exits_nonzero_when_report_has_no_pages`
- Not verified failing-first; this is a harvest/review-fix from an already-verified scratch migration, with review-provided mutations converted into regression tests.

## Deliberately not changed
- Did not copy any bundle-specific `fix_*.py` files.
- Did not add a new top-level folder or a second script.
- Did not duplicate `.github/pbi.kb/visuals/azureMap.md`; it already contains the detailed Location-well evidence.
- Did not run a live Desktop capture; the new logic is covered offline by faking the screenshot call.
- Checked the installed `powerbi-desktop --help` surface; it exposes status/manifest/open/reload/screenshot/screenshot-all, but no page-render readiness signal, so this remains a dwell heuristic with residual plateau risk.

## Gates
- `python -m ruff format scripts\capture_powerbi_pages.py tests\test_capture_powerbi_pages.py`
- `python -m ruff check scripts\capture_powerbi_pages.py tests\test_capture_powerbi_pages.py --fix`
- `python -m pylint scripts` -> exit 0, 10.00/10
- `python -m pytest -q` -> 534 passed, 4 skipped
- `python scripts\sync_agent_conventions.py --check` -> OK

## Weakest point
Byte-identical dwell is still a heuristic because the bridge exposes no page-render readiness signal. A partial plateau longer than `--stable-seconds` can still pass; dynamic/animated visuals may never converge and are kept as newest-frame evidence with a non-zero exit.


Guust-Franssens:

Reviewer given only the diff, the issues and the original script — not the author's reasoning.

### Blocking: pairwise equality can certify a *partial-render plateau*

`scripts/capture_powerbi_pages.py:141-150` declares convergence after **one** unchanged polling
interval. Offline sequence:

```
frames:  partial, partial, complete
result:  converged=True, frames=2
saved:   partial          <-- exit 0, reported OK
```

A visual waiting several seconds on a remote GeoJSON fetch can sit **unchanged and incomplete** across
one poll before completing. So the script can confidently save an under-drawn image and exit 0 — which
is precisely the silent under-draw bug it was written to prevent, reproduced in the fix itself.

Suggested: require stability across a configurable **time window / minimum dwell** rather than a single
equal pair, plus a `[partial, partial, complete, complete…]` regression test. A real readiness signal
would be better still if one is obtainable.

### Non-blocking

- **Process-level failure is untested.** The tests genuinely catch the "accept first capture" and
  "compare frame 0" mutations — but mutating line 194 so that an unstable report exits **0** still
  passed both. Add an offline `capture_report()` test asserting a non-zero exit on instability.
- **Zero pages succeeds.** A missing or invalid report path prints `0/0 captured` and exits 0
  (`:173-194` never rejects an empty page list). Should fail closed.

### Verified clean

- Animated/continuously-changing pages hit `--max-wait`, keep the newest frame, print
  `UNSTABLE`/`PARTIAL`, and **exit 1** — the loud path works.
- A genuinely static page converges after two captures.
- **The original header reasoning survived the move**, including all three approaches that do not work
  (`--settle`, `sleep()`-then-screenshot, one long fixed sleep). That was the main thing at risk in a
  harvest and it is intact.
- **Clutter constraint met**: 5 files, 316 insertions, exactly one new script, no new top-level
  directory, no bundle-specific `fix_*.py` copied, and `azureMap.md` **unchanged** — so no prose
  restating what the cookbook already carries.
- The `scripts/README.md` row, the one-line persona check, and the placement under `scripts/` are all
  appropriate; the persona already routes Azure Maps work to the cookbook for detail.

**Most important check for a human:** retain *every* frame from a cold Azure Maps capture and confirm
whether identical partial frames genuinely occur before the final nationwide render — that is the
real-world evidence behind the blocking finding.

Guust-Franssens:
## Final delta check — **approve**, no findings

Third round, reviewed by the same agent that raised both earlier blockers.

**The dwell timing is now correct** (`capture_powerbi_pages.py:146-159`). The slow-first-capture
reproduction — the round-2 blocker — now saves the **complete** frame after 5 frames instead of
certifying the partial plateau at 2. Delays *inside* every capture and during digest/comparison are
excluded too, so there is no remaining path where time spent capturing counts as time spent stable.

**The new test genuinely fails against the previous commit**: against `61228b6`,
`test_capture_stable_does_not_count_screenshot_duration_as_dwell` produced 2 frames instead of 5 and
saved the partial. The pairwise and unstable-exit mutations are still killed by their own regressions.

**The 20-second default is defensible in both directions.** It requires 6 captures/page — 200 s idle
for 10 pages, ≈5m56s at issue #65's measured 2.6 s capture cadence. Acceptable for a final sign-off
pass, which is what this tool is for.

**The prose is now honest** — the heuristic and its residual plateau risk are stated explicitly in all
three places (`capture_powerbi_pages.py:35-38`, `scripts/README.md:47`,
`powerbi-report-gotchas/SKILL.md:337-342`). Overstating certainty in a doc is the same class of defect
as overstating it in code, so this mattered.

Nothing regressed: invalid report path exits 1, the header retains all three failed approaches, exactly
one new script, no new top-level directory, `azureMap.md` untouched, README row and the one-line persona
check intact. Gates reproduced independently: 534 passed / 4 skipped, ruff clean, pylint 10.00/10,
convention check passed.

Not verified live: a real Desktop cold-map capture and the true 10-page wall time. No Desktop instances
were opened or terminated during review.

---

Three rounds on a harvest is a lot, and each round found something real: convergence declared on a
single equal pair; then the dwell clock started before the capture it was timing; then honesty in the
prose. The through-line is the same one this batch keeps producing — **"converged" was treated as an
event when the phenomenon is a duration**, and each fix moved the boundary before removing it.

PR #72, opened 2026-08-10
Validate migration specs after agent appends

Guust-Franssens:
## What was wrong

`migration-spec.json` was schema-validated only when `parse_tableau.py` first wrote it. Later agents append `limitations_encountered`, so a bad append such as `severity: "critical"`, a missing `stage`, or a typo key could silently become downstream input.

## What changed

- Added `scripts/validate_spec.py` for mid-migration re-validation of one spec or all committed specs.
- Reused the parser's schema validation path by extracting actionable error collection in `scripts/parse_tableau.py`; this avoids a second validator implementation.
- Made the standalone validator fail closed when validation cannot run: missing `jsonschema`, invalid Draft 7 schema (`check_schema`), or a schema whose compiled behavior fails inline contract canaries.
- Added inline canaries: one known-good spec must pass; bad severity, missing stage, non-object limitation, and typo-key limitation must fail. This tests behavior rather than schema shape.
- Improved errors to name the path and limitation entry context, e.g. `limitations_encountered[...].severity (item=..., stage=...): ... expected one of: info, low, medium, high`.
- Tightened `limitations_encountered` schema entries to require `stage` and reject unknown keys.
- Wired the gate into CI, the writer personas' final append/report action, and orchestrator gates before report-builder and sign-off delegation.
- Corrected the semantic-builder persona to use schema-valid `stage: "semantic_build"`.

## Tests

- Watched failing-first: `python -m pytest tests\test_spec_contract.py -q` failed before implementation with `ModuleNotFoundError: No module named 'validate_spec'`.
- Watched the `$ref` bypass test fail against the previous commit (`7588f4e`) before replacing shape checks with canaries.
- Added coverage:
  - `test_malformed_appended_limitation_names_the_bad_field`
  - `test_missing_appended_limitation_stage_is_rejected`
  - `test_typoed_appended_limitation_key_is_rejected`
  - `test_cli_rejects_malformed_append_with_actionable_output`
  - `test_cli_fails_closed_when_jsonschema_is_unavailable`
  - `test_cli_fails_closed_when_schema_cannot_validate_the_contract`
  - `test_legitimate_appended_limitation_passes_and_is_not_rewritten`
  - `test_every_committed_spec_validates_against_the_schema`

## Gates run

- `ruff format/check` on changed Python files
- `python -m pylint scripts` → 10.00/10
- `python -m pytest -q` → 540 passed, 4 skipped
- `python scripts/validate_spec.py --all` → 16 specs valid
- Corpus count rechecked: 16 specs, 794 limitation entries
- `python scripts/sync_agent_conventions.py --check` → OK; `tableau-migrator` 29,986 chars

## Deliberately not changed

- Did not merge/cherry-pick `feat/battle-test-hardening` wholesale.
- Did not add re-parse overwrite protection from the stale branch; this PR is scoped to post-append validation.
- Did not make every object in the whole schema `additionalProperties: false`; only the appended limitation entries are tightened to avoid rejecting legitimate existing spec shapes.
- Did not reject exact duplicate limitations. They are schema-valid today and 8 of 16 committed specs already contain duplicates, so that needs a separate contract/corpus cleanup decision.

## Weakest point

The gate is still instruction-driven for subagent writes rather than an atomic append-and-validate helper. It now runs in each writer persona and before downstream delegation, but a helper would be harder to forget.

### DIFF
--- a/.github/agents/pbi-report-builder.agent.md
+++ b/.github/agents/pbi-report-builder.agent.md
@@ -235,7 +235,7 @@ it is slow, and `validate` will not catch a wrong encoding.
    survive a landing re-run.
 7. Report back: what you repaired, what you left as an accepted limitation *and why*, any
    `viz_fidelity` row you believe is a false claim (route it back, never silently fix), and new
-   `limitations_encountered` entries (`stage: "report_build"`).
+   `limitations_encountered` entries (`stage: "report_build"`); then run `python scripts/validate_spec.py <migration-spec.json>`.
 
 **If a page must be built from scratch** (no rebuilt equivalent - rare), fall back to the full
 authoring chain: `powerbi-report-planning` -> `powerbi-report-design` -> **empty layout skeleton,
--- a/.github/agents/pbi-semantic-builder.agent.md
+++ b/.github/agents/pbi-semantic-builder.agent.md
@@ -177,7 +177,7 @@ enrich it, and hand it over refreshed.
    what Desktop's own Save does. Edit → refresh → save, in that order.
 9. **Report back**: model location, what you authored vs. what the engine did, every table-calc
    decision (visual calc vs measure, and why), anything you routed rather than fixed, and new
-   `limitations_encountered` entries (`stage: "model_build"`).
+   `limitations_encountered` entries (`stage: "semantic_build"`); then run `python scripts/validate_spec.py <migration-spec.json>`.
 
 ## Prep the model for AI (Copilot readiness) — final build phase
 
--- a/.github/agents/tableau-migrator.agent.md
+++ b/.github/agents/tableau-migrator.agent.md
@@ -231,8 +231,9 @@ it must show up in `limitations_encountered`, not be silently dropped.
      happen before any report work begins. Do not run report and model fixes concurrently against one
      bundle.
 9. **Delegate to `pbi-report-builder`** — only AFTER step 8's landing re-run has finished, because
-   that re-run recreates the `.Report` folder and would destroy its work. **Gate the handover:**
-   `python scripts/check_migration_progress.py --bundle <bundle> --handoff` must exit 0. Exit 1 means
+   that re-run recreates the `.Report` folder and would destroy its work. **Spec+handoff gates (both
+   exit 0):** `python scripts/validate_spec.py <spec>`;
+   `python scripts/check_migration_progress.py --bundle <bundle> --handoff`. Exit 1 means
    a model has no `cache.abf`, or one **older** than its TMDL — the report builder would open an
    EMPTY model and trigger its own refresh (minutes, plus a credential prompt on a live source).
    Measured: a cache written at 22:22 against a Desktop opened at 22:19 did exactly that, and a stale
@@ -240,7 +241,8 @@ it must show up in `limitations_encountered`, not be silently dropped.
    Give it: the handover
    slice, the validator's classification from step 7, the model location, and the reference bundle.
    Its edits must land as re-runnable `_build/fix_*.py` scripts, not bundle-only edits.
-10. **Delegate to `pbi-migration-validator` again — full sign-off mode**, on a FRESH invocation. Name
+10. **Delegate to `pbi-migration-validator` again — full sign-off mode**, on a FRESH invocation. First
+   rerun `python scripts/validate_spec.py <spec>`; block on failure. Name
    the mode explicitly; it is a different job from step 7's triage. It sees the artifacts, the
    reference bundle and the triage classifications, but **not the builders' rationale** — and it is
    told the classifications are **claims to verify, not settled facts**, including the ones an earlier
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -49,6 +49,9 @@ jobs:
       - name: Tests
         run: uv run pytest -q
 
+      - name: Migration spec contract gate
+        run: uv run python scripts/validate_spec.py --all
+
       - name: Privacy gate - no absolute user paths in tracked files
         run: uv run python scripts/set_data_folder.py --check
 
--- a/docs/migration-spec.schema.json
+++ b/docs/migration-spec.schema.json
@@ -277,7 +277,8 @@
       "description": "Appended to by every pipeline stage (parser + subagents). Raw material for the capabilities & limitations writeup.",
       "items": {
         "type": "object",
-        "required": ["item", "issue", "severity"],
+        "required": ["item", "issue", "severity", "stage"],
+        "additionalProperties": false,
         "properties": {
           "item": { "type": "string", "description": "id of the affected worksheet/dashboard/field" },
           "issue": { "type": "string" },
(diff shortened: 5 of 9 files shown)


Guust-Franssens:

Reviewer given only the diff, the issue and the repo — not the author's reasoning.

### Blocking

**1. The validator fails OPEN when `jsonschema` is unavailable.**
`parse_tableau.py:1527-1533` returns `[]` after an `ImportError`, and `validate_spec.py:68-70` reads
that as *valid*. With a `severity: "critical"` append and the dependency simulated missing:

```
WARNING: jsonschema not installed - skipping validation
INFO: ok ... malformed-without-jsonschema.json
INFO: 1 spec(s) valid.
errors=[]  exit=0
```

So the new contract gate can explicitly report malformed input as valid — the precise silent-failure
class this PR exists to remove, reproduced inside the fix for it. **A gate that cannot run must not
report clean.** Suggested: `validate_spec.py` exits non-zero when validation cannot be performed; if
the parser needs optional validation for compatibility, give it a strict mode the standalone validator
always uses. Plus a regression test.

**2. It is wired too late to do its job.** Not unused — but the only runtime invocation is orchestrator
step 12 (`tableau-migrator.agent.md:268`), *after* semantic build, report build and sign-off delegation.
Meanwhile both writer personas append entries (`pbi-semantic-builder.agent.md:178-180`,
`pbi-report-builder.agent.md:236-238`) and **neither references `validate_spec.py`**; subagents receive
only their own persona file (`tableau-migrator.agent.md:14`), so they cannot inherit it. The report
builder explicitly *consumes* the spec at `:116`.

Net effect: a malformed semantic-build append still reaches the report builder and the validator before
the gate fires. CI (`checks.yml:52-53`) is later still. The issue's premise was catching a bad append
**at the moment it happens**. Suggested: make validation a mandatory final action in every writer
persona and a blocking gate before orchestrator steps 9 and 10 — ideally an atomic append-and-validate
helper, so it cannot be forgotten.

### Non-blocking

**3. The tests do not exercise the failing CLI path.** The three malformed cases call the helper
directly (`test_spec_contract.py:32,53,78`); the only subprocess test is the *valid* case (`:83-103`).
So a regression where the CLI returns 0 for invalid input, or the dependency is absent, stays green.

### Verified good — and some of this is genuinely strong

**Detection, dependencies installed:**

| append | result |
|---|---|
| bad `severity` | caught |
| missing `stage` | caught |
| misspelled key | caught |
| string instead of object | caught |
| **exact duplicate entry** | **slips** |

Duplicates are schema-valid today, and **8 of 16 committed specs already contain 39 exact duplicate
rows** — so adding `uniqueItems` needs a corpus cleanup and an explicit contract decision first. Worth
a separate issue rather than scope creep here.

**No over-strictness**, which was my main worry: `--all` reports **16 specs valid**; an audit of **794
committed entries** found all use exactly `item`/`issue`/`severity`/`stage`; every persona-described
stage passes.

**No duplicated validation logic** — `validate_spec.py:15,31` reuses the parser's shared collector
rather than reimplementing schema handling.

**The error message is genuinely actionable**, which was the deliverable:

```
ERROR: limitations_encountered[22].severity (item='worksheet:profit',
stage='semantic_build', severity='critical'): 'critical' is not one of
['info','low','medium','high']; expected one of: info, low, medium, high
```

**Against master's validator**, missing `stage`, a stage-key typo and duplicates all slipped — so the
detection improvement is real, not cosmetic.

**Most important check for a human:** confirm a malformed semantic-builder append **blocks launching the
report builder**, rather than being discovered at final sign-off.

Guust-Franssens:
Addressed review findings in 245a62d: standalone validator now fails closed when jsonschema is unavailable, with subprocess coverage for invalid CLI and missing-dependency paths; writer personas now validate after appends; orchestrator gates validate before report-builder and sign-off delegation. Left duplicate limitations as a documented known gap per review.

Guust-Franssens:
## Re-review of the delta — **request changes** (one finding, and it is the same shape one level down)

Both original blockers are **confirmed resolved**:

- Missing `jsonschema` → malformed spec now exits **1** with `validation unavailable`. The new
  dependency test is genuine: against `3245965` it fails, because the old CLI returned 0.
- **The wiring is now real.** Both writer personas explicitly run validation after appending;
  orchestrator step 9 requires exit 0 *before* the report builder launches, and step 10 blocks before
  sign-off. Missing explicit path → 1; empty `--all` → 2; unreadable files and malformed schema JSON
  terminate non-zero. Persona sync passes at **29,986/30,000** chars, so something was rewritten rather
  than bolted on.
- No over-strictness: **16 specs, 794 entries, 0 invalid**. Duplicate gap documented in the PR body
  (tracked separately as #75).

### Still blocking: a malformed SCHEMA fails open

`parse_tableau.py:1544` constructs `Draft7Validator` **without validating the schema itself**. Mutating
`severity.enum` from an array into an invalid object, then validating a real spec carrying
`severity: "critical"`:

```
INFO: ok ...\migration-spec.json
INFO: 1 spec(s) valid.
exit=0
```

`jsonschema.Draft7Validator.check_schema(schema)` raises `SchemaError` for that mutation — it is simply
never called.

This is the *same* finding as round 1, one level down: the missing-dependency path was patched, but the
principle — **a gate that cannot run must not report clean** — was not generalised. A corrupt schema is
another way validation cannot meaningfully run, and it currently reports clean.

*Fix*: call `check_schema` before constructing the validator; convert `SchemaError` into
validation-unavailable for strict callers; add a non-zero CLI regression test.

Guust-Franssens:
Addressed the remaining fail-open case in 7588f4e: the strict validator now calls Draft7Validator.check_schema and rejects vacuous/non-contract schemas before validating specs. Added CLI regression coverage for invalid enum schema, empty schema, and irrelevant schema. Re-ran: pylint scripts 10.00/10; pytest 539 passed, 4 skipped; validate_spec --all 16 valid; persona sync OK.

Guust-Franssens:
Addressed round-3 finding in 575321d: replaced schema-shape contract checks with compiled-validator behavior canaries. Added failing-first CLI coverage for the $ref accept-anything bypass; canaries now require one good spec to pass and bad severity/missing stage/non-object/typo-key specs to fail. Re-ran: pylint scripts 10.00/10; pytest 540 passed, 4 skipped; validate_spec --all 16 valid; corpus count 794 limitation entries; persona sync OK.

Guust-Franssens:
Final delta check clean.

- The top-level `$ref` accept-anything bypass, removed item object type, and enabled additional properties all fail closed with the offending named canary.
- The `$ref` regression test fails against `7588f4e` for the expected exit-0 behavior.
- The old shape-proxy helpers were removed; the replacement uses inline good/bad behavioral canaries only.
- Missing-dependency, writer/orchestrator wiring, corpus compatibility, and persona-cap checks remain intact.
- Verified: 540 passed / 4 skipped; 16 specs and 794 limitation entries valid.

Canaries are a behavioral sample rather than proof of full schema equivalence. I consider that residual acceptable: the scope is explicit in the inline named canaries, and extending coverage is a straightforward additional canary rather than another design change.

Verdict: approve

Issue #61, opened 2026-08-08
Edits to generated artifacts must be DECLARED and re-runnable - an undeclared (and correct) fix made an upstream engine bug look like an agent defect


Guust-Franssens:
## What happened

`e2e5-maps` (a `tableau-migrator` run on `book_6-1-Maps`) hand-edited a generated TMDL partition and
broke its own model, then spent ~90 minutes on the resulting refresh failure — writing **1** deliverable
file in that time.

Measured:

| artifact | timestamp |
|---|---|
| engine generated the bundle (`report.json`) | 19:48:24 |
| agent rewrote `.../tables/Orders.tmdl` | **20:20:40** |

The edit changed the Power Query navigation key:

```diff
- Navigation = Source{[Item="Orders",  Kind="Sheet"]}[Data],
+ Navigation = Source{[Item="Orders$", Kind="Sheet"]}[Data],
```

`Excel.Workbook()` exposes worksheets **without** a trailing `$`. The `[Orders$]` form is the legacy
ACE/OLEDB identifier Tableau stores in the relation's `table=` attribute. The result is a refresh that
fails with **"The key didn't match any rows in the table"**, plus a Desktop
`InvalidPackageReferenceException` while the error dialog was up.

A clean re-run of the same `.twbx` on the same engine build emits `Item="Orders"`. So this was
self-inflicted: the agent almost certainly took Tableau's `table='[Orders$]'` attribute as the sheet
name while "fixing" something, and *caused* the failure it then tried to debug.

## Why this needs a rule, not just a correction

The personas already forbid hand-editing for the DAX case — *"land approvals through `--approved-dax`,
never by hand-editing `_Measures.tmdl`"* — but that is written as a **DAX/measures** rule. Nothing
generalises it to **generated TMDL partitions / M expressions**, and the agent did not read it as
covering them.

The rule should be the general one, for the same two reasons that motivated the DAX version:

1. a bundle-only edit is **destroyed by the next landing re-run**, so the work is throwaway even when it
   is correct; and
2. editing generated output **converts a reportable finding into a local mutation** — nobody upstream
   ever learns, and the artifact silently diverges from what the generator produces.

Reason 2 bit hard here: I built a differential across bundles, saw `Orders$` in one and `Orders` in two
others, and **filed an upstream bug against the deterministic engine**
(`Yarbrdab000/tableau-fabric-skills#108`, since retracted and closed as invalid). The agent's undeclared
[...]

PR #73, opened 2026-08-10
Detect undeclared generated artifact edits

Guust-Franssens:
## What was wrong

Generated TMDL/PBIR/`.pbip` artifacts could be edited in place after the deterministic engine ran with no durable signal. That made a correct local repair both disposable on the next landing re-run and indistinguishable from pristine engine output during diagnosis.

## What changed

- `run_estate.py` records stable generated-artifact hashes under `input_manifest.json.generated_artifacts` immediately after the engine run, including a run id and a hash of the current `report.json`.
- `check_migration_progress.py --tamper` recomputes the current `report.json` hash before trusting the baseline, then reports changed/missing/added generated files.
- Legitimate generated edits are declared through `_build/generated-edit-declarations.json`, matched by run id, target path, baseline hash, and expected current hash.
- Added `scripts/declare_generated_edit.py`, a generic audit wrapper for replayable `_build/fix_*.py` scripts. It runs the fix, records the target before/after hashes, and supports composed-path fixes without scanning source text.
- `pbip-model-refresh` now writes the same declaration record for its normal `database.tmdl` compatibility-level rewrite during refresh/save.
- `.pbi` sidecars are excluded so refresh caches/Desktop autosave files do not become false positives; `data/` is deliberately out of scope because this gate is for generated model/report artifacts.
- Shared agent conventions require generated edits to include file/change/why plus replay script plus hash record, and `tableau-migrator` runs `--tamper` before sign-off.

## Design choices

I chose option (a): make declaring generic and cheap. Narrowing the gate would have avoided false positives by reducing coverage, but the known real maps repair is exactly the class of edit issue #61 needs to make visible. The new wrapper lets existing composed-path fixes participate without being rewritten to literal path strings.

The declaration is an audit record, not authorization. The editing script writes its own evidence, so this is not a security control; it makes drift visible and replayable for review.

## Covering tests

- `test_generated_artifact_manifest_records_only_stable_generated_files`
- `test_generated_artifact_manifest_ignores_foreign_roots_from_before_this_run`
- `test_tamper_detects_an_undeclared_generated_artifact_edit`
- `test_tamper_allows_a_generated_edit_declared_by_a_fix_script`
- `test_declare_wrapper_records_a_composed_path_fix_script`
- `test_tamper_rejects_a_source_comment_without_structured_hash_evidence`
- `test_tamper_rejects_a_structured_declaration_with_the_wrong_output_hash`
- `test_tamper_rejects_a_manifest_bound_to_a_different_report`
- `test_compatibility_alignment_declares_generated_edit`

I watched the original new tests fail before the first implementation. Review-driven tests were added for the later failure boundaries.

## Deliberately not changed

- Did not prohibit generated artifact edits; declared replayable repairs pass.
- Did not hash `.pbi` cache/autosave sidecars.
- Did not include `data/`, `report.json`, `summary.md`, or handoff slices in the tamper scope.
- Did not make declarations an authorization/security mechanism.

## Gates

- `ruff format` / `ruff check --fix` on changed Python files ✅
- `python -m pylint scripts` ✅
- `python -m pylint .github\skills\pbip-model-refresh\scripts\refresh_pbip_model.py` ✅
- `python -m pytest -q` ✅ `541 passed, 4 skipped`
- `python scripts\sync_agent_conventions.py --check` ✅ (`tableau-migrator` 29984 chars)

## Weakest point / reviewer probe

A matching declaration can be hand-forged by someone who can edit the bundle. That is intentional for an audit trail, but reviewers must still inspect whether the referenced fix script is legitimate and idempotent.


Guust-Franssens:

Reviewer given only the diff, the issue and the repo — not the author's reasoning. Three blocking
findings; the second is the most important.

### 1. A NORMAL refresh is reported as undeclared tampering

`refresh_pbip_model.py:301-338` deliberately rewrites `definition/database.tmdl` when Desktop raises the
compatibility level. Probed on a real pristine→refreshed bundle:

```
state=DRIFT
UNDECLARED changed: .../definition/database.tmdl
```

The only difference was `compatibilityLevel: 1604 → 1702`; both files stayed 37 bytes. So running
`--tamper` after the **required** refresh/persist phase produces a false positive — on the happy path.

That is the failure mode that kills this feature: a checker that fires on a correct migration gets
bypassed, and then it protects nothing. Suggested: have the refresh tool emit a structured declaration
tied to before/after hashes — **not** a blanket exclusion of `database.tmdl`, which would blind the
check to a real edit of the same file.

### 2. `_build/fix_*.py` matching launders in one direction and rejects in the other

`check_migration_progress.py:137-162` does a raw, case-sensitive **substring search of script source**.
Adversarial probe:

```
stale_comment_launders:        True
docstring_only_launders:       True
wrong_model_prefix_launders:   True
real_composed_path_matches:    False
case_variant_matches:          False
```

So a stale script mentioning a path in a comment **permanently launders every future edit to that
path** — while a genuine replayable fix
(`C:\tfmig\e2e2\3-maps\out\_build\fix_post_engine.py:171-174,382-397`, which really does change
`Orders.tmdl`) is **not** recognised, because it builds the path from components rather than writing it
literally.

Both directions are wrong, which means the mechanism does not establish what it claims. And the
covering test encodes the flaw: `tests/test_check_migration_progress.py:263-277` creates an otherwise
empty script whose docstring merely mentions the path.

Suggested: replace source-text matching with a per-run structured record — script identity/hash, target
path, baseline hash, expected post-fix hash — and count the edit as declared only when the resulting
hash matches.

### 3. The manifest is not bound to the engine run

`check_migration_progress.py:124-134` ignores manifest `version`, `recorded_at`, inputs and engine-run
identity; a manifest with `version: 999` and a **1999** timestamp was accepted without objection.
Separately, `run_estate.py:132-159` stamps every matching path already present in the bundle — not only
paths the current run emitted — so a landing re-run can bless stale or foreign `.Report`/`.SemanticModel`
roots the engine never recreated.

Credit: **missing evidence does fail closed** (`NO_BASELINE`, exit 2), and malformed JSON propagates an
error. It is *different-run* evidence that is accepted.

### Non-blocking

- **No persona invokes `--tamper`** — zero hits across `AGENTS.md` and `.github/agents`. Also
  `--handoff --tamper` silently runs only handoff and returned `READY` with no tamper baseline; the
  modes should be mutually exclusive or composable, not silently one-wins.
- Coverage excludes all of `data/`, `report.json`, `summary.md` and `handover/`, so an edit to generated
  **data** can change refreshed numbers while the check reports clean.

### Verified good

Same-size modification **detected**; added generated file **detected**; deleted generated file
**detected**; `.pbi` cache/settings correctly stayed clean on real bundles. Convention sync exit 0 with
terse text — though `tableau-migrator` is now **29,995/30,000** characters, which is worth noting.

**Most important check for a human:** run the full sequence — engine → standard refresh/save → replay
existing semantic `_build` fixes → `--tamper`. It must report no undeclared drift.

Guust-Franssens:
Addressed the request-changes review in amended commit `108e130`.

Changes:
- Normal refresh compatibility-level rewrites now emit `_build/generated-edit-declarations.json` with run id, target, before hash, expected after hash, script identity/hash, and reason.
- `--tamper` no longer scans `_build/fix_*.py` source text. It accepts drift only when structured declaration evidence matches this engine run and the current output hash.
- Baselines now require current run identity (`version`, `run_id`, `recorded_at`, `report_sha256`) and `run_estate.py` only stamps files written during the current engine phase, so stale/foreign roots are not blessed.
- `--handoff --tamper` now fails closed as mutually exclusive modes.
- `tableau-migrator` now invokes `--tamper` before sign-off; README documents the TMDL/PBIR/`.pbip` scope and that `data/` is deliberately out of scope.

Added/updated regressions for the three blockers: refresh declaration, comment/docstring laundering rejection, wrong output hash rejection, bad/old manifest rejection, stale foreign root exclusion, mode exclusivity.

Gates after the final changes:
- `ruff format` / `ruff check --fix` on changed Python files ✅
- `python -m pylint scripts` ✅
- `python -m pylint .github\skills\pbip-model-refresh\scripts\refresh_pbip_model.py` ✅
- `python -m pytest -q` ✅ `539 passed, 4 skipped`
- `python scripts\sync_agent_conventions.py --check` ✅ (`tableau-migrator` exactly 30000 chars)

Guust-Franssens:
## Round 2 delta review — **request changes**

Formal “request changes” reviews are unavailable on the authenticated author's own PR, so posting the same verdict as a PR comment.

### Blocking

1. **The strictness inversion is real: the gate rejects the repository's real replayable fixes.**

Against the actual maps bundle, I used the earliest `_build/backups/Orders.tmdl...` hash as the engine baseline while leaving the current, legitimately repaired `Orders.tmdl` and its real `C:\tfmig\e2e2\3-maps\out\_build\fix_post_engine.py` in place. There is no generated-edit declaration because that script has never been taught to write one. Current result:

```
real_fix_script_exists=True
declaration_file_exists=False
state=DRIFT
UNDECLARED changed: pbip/book_6-1-Maps/Orders (Sample - Superstore).SemanticModel/definition/tables/Orders.tmdl
```

The source-prose laundering is fixed, but the new truth table is: stale comment false; docstring-only false; wrong-model prefix false; **real composed-path fix without a record false**; case variant false; exact structured record true; stale/wrong-run record false.

Conceptually, requiring editing tooling to emit an audit record is acceptable. It is not usable as shipped when only `refresh_pbip_model.py` does so, no generic wrapper/helper exists, and the newly wired sign-off gate blocks the known-good fix pattern that motivated the issue. The covering test still hand-fabricates JSON via `_write_declaration`; it does not run a representative composed-path fix through a supported declaration path. Please add a generic “run fix and record before/after hashes” wrapper (or retrofit every canonical fix framework) and test the actual composed-path shape.

2. **The baseline is still not tied to the current engine run; required fields are only present/parse checks.**

`load_generated_artifact_baseline()` does not recompute `report_sha256`, compare `report_generated_at`, or apply any freshness/relationship check to `recorded_at`. A well-formed version-1 manifest with `run_id='foreign-run'`, a **1999** timestamp, and `report_sha256='definitely-not-the-current-report'` was accepted and `tamper_check()` returned:

```
('CLEAN', ['0 generated artifact(s) are pristine against their engine-run hashes'])
```

The new mtime cutoff does fix the ordinary “old root left from before this landing run” case, but it does not make a copied/foreign manifest fail closed. Recompute and compare the current `report.json` hash (and/or bind the manifest to report `generated_at` / input identity) rather than treating non-empty identity fields as proof.

### Non-blocking documentation

- `--tamper --help` still says “without a matching `_build/fix_*.py`”, describing the deleted mechanism.
- The declaration is intentionally a self-authored audit record, not authorization: `script_identity` and `script_sha256` are only checked for non-emptiness and an exact hand-appended record is accepted. State that trust boundary explicitly.
- The PR body is also still the round-1 description (source-text matching, old test names, 533-test result), so it now misdescribes the implementation.

### Re-verified as fixed

- **Actual refresh, no hand-written declaration:** on the live maps Desktop instance I stamped a 1604 baseline, ran the real refresh/save path, got `DATA_OK + PERSISTED`, the tool aligned `1604 -> 1606`, and `--tamper` returned `DECLARED_DRIFT` via `pbip-model-refresh/refresh_pbip_model.py`. Artifacts were restored after the probe.
- Source-comment/docstring/wrong-prefix laundering is gone; stale output hashes and wrong run IDs do not match.
- `--handoff --tamper` errors with exit 2.
- `data/` exclusion is now explicitly documented as deliberate.
- Migrator invokes `--tamper` before sign-off. Convention sync passes; its prompt is exactly `30000` characters (valid, zero headroom).
- `539 passed, 4 skipped`; both pylint gates are 10.00/10.

Guust-Franssens:
Addressed round 2 in amended commit `ed98adc`.

Choices made:
- Picked option (a): added generic `scripts/declare_generated_edit.py` wrapper so composed-path `_build/fix_*.py` repairs can record before/after hashes without source-text matching.
- Kept and validated run-binding fields: `--tamper` now recomputes `report.json` hash (and compares `generated_at` when recorded) before trusting the baseline.

Also fixed: stale help text, trust-boundary docs, PR body, and prompt headroom (tableau-migrator now 29984 chars).

Final gates: `pylint scripts`, skill-script pylint, `pytest -q` = 541 passed / 4 skipped, `sync_agent_conventions.py --check`.

Guust-Franssens:
## Final delta review — **approve**

The two blockers are closed.

- **Real maps repair passed.** Using the actual `C:\tfmig\e2e2\3-maps\out\_build\fix_post_engine.py` and its real `Orders.tmdl` engine backup, I stamped the engine baseline, ran the fix through `scripts/declare_generated_edit.py`, then ran `--tamper` against the actual bundle. Result:

  ```
  DECLARE: RECORDED .../Orders.tmdl
  TAMPER [DECLARED_DRIFT] out
    DECLARED changed: .../Orders.tmdl via _build/fix_post_engine.py
  ```

  The bundle was restored after the probe. This is the composed-path fix that failed round 2; it now works end to end.

- **Foreign baseline now bites.** The exact version-1 / foreign-run / 1999 timestamp / incorrect-report-hash probe returns `NO_BASELINE`, not `CLEAN`. Wrong `report_sha256` and wrong `report_generated_at` are independently rejected; a missing run id is rejected. The run id scopes declarations, while `recorded_at` remains parse-validated audit metadata rather than an age-expiry rule.

- The helper computes the actual target hash before and after execution; hashes are not caller-supplied. The checker still verifies the recorded pre-edit hash against the engine baseline. `target.relative_to(bundle)` prevents an outside-bundle target from being written into the declaration record. The documented audit-not-authorization trust boundary accurately describes the remaining forgery model.

- Help text and PR body describe hash declarations, not source matching. Trust-boundary wording is present. `tableau-migrator` still runs `--tamper` before sign-off and is 29,984 characters.

- Gates reproduced: `541 passed, 4 skipped`; both pylint runs 10.00/10; convention sync clean.

Formal GitHub approval is unavailable on the authenticated author's own PR, so recording the approval here.

PR #103, opened 2026-08-11
Consolidate Tableau auth on PAT-only, and fix two scripts broken in opposite ways

### LINKED ISSUE #97: Our tier is PAT-only: consolidate every Tableau auth path through tableau_env.py (tableau_lineage.py is currently broken against our own .env)
**Decision (owner, 2026-08-11):** our tier supports **PAT only**. The deterministic engine keeps whatever it supports (JWT/Connected Apps, keyring) — that is his code, it works, and we do not touch it. This issue is about making *our* half honest, single-pathed, and actually functional.

## Why PAT-only is the right call

PAT is the simplest credential that works against Tableau Cloud, where MFA makes it effectively mandatory for REST. One auth mode means one failure mode, one set of variables, one thing to document.

**The knowing trade, recorded so nobody re-litigates it later:**

- A PAT **inherits its owner's permissions and cannot be scoped down**. On a real site that often means `SiteAdministratorCreator`.
- **There is no create-PAT API** — `POST .../personal-access-tokens` returns **HTTP 405**. List and revoke only. A human always provisions it.
- A tenant whose policy **forbids PATs** cannot run our assessment at all. Connected Apps/JWT is the only unattended path there. We decline those tenants rather than half-supporting them.

## The live defect this closes

`scripts/tableau_lineage.py` is **step 2 of our own documented site path** (`AGENTS.md` triage table). It is broken for anyone who set up `.env` from `.env.example`:

```
$ python scripts/tableau_lineage.py --plan
Missing environment variable(s): TABLEAU_SERVER, TABLEAU_PAT_NAME, TABLEAU_PAT_SECRET
```

Two independent faults:

1. **It reads `os.environ` directly** (`:213-216`) and never loads `.env`. `grep` for `load_env|tableau_env|dotenv` in that file returns nothing.
2. **It wants `TABLEAU_SERVER`**, while `assess_estate.py`, `capture_tableau_oracle.py`, `capture_tableau_reference.py`, `harvest_estate_assets.py`, `stamp_tableau_provenance.py` and `.env.example` all use **`TABLEAU_SERVER_URL`**.

So a user follows our documented flow, step 1 works, step 2 fails with a message that names variables our own example file does not define. This is the same class as #88 (`TABLEAU_PAT_SECRET` vs `TABLEAU_PAT_VALUE`) — that one is fixed; this one was never noticed because the dry run reached `assess_estate.py` and stopped.

## Scope

**1. Route every Tableau-auth script through `scripts/tableau_env.py`.** It landed in #91 and is already the single place the two tiers' naming is reconciled. Currently used by four scripts; `tableau_lineage.py` and `make_showcase.py` are not among them.

**2. One canonical variable set**, defined in `.env.example` and nowhere else:

| variable | meaning |
|---|---|
| `TABLEAU_SERVER_URL` | e.g. `https://10ax.online.tableau.com` |
| `TABLEAU_SITE` | site contentUrl (`''` for Default) |
| `TABLEAU_PAT_NAME` | token name |
| `TABLEAU_PAT_SECRET` | token secret (bridged to the engine's `TABLEAU_PAT_VALUE` by `tableau_env`) |
| `TABLEAU_REST_API_VERSION` | optional override |

`TABLEAU_SERVER` becomes an **accepted alias with a deprecation warning**, not a silent second name — an existing `.env` must keep working.

**3. Delete the per-script auth reimplementations.** `#88` found four divergent `load_env` copies, two of which crashed on a missing file while two returned empty. `tableau_env.load_env` is now the one implementation; anything else goes.

**4. State PAT-only where a reader hits it** — `.env.example`, `docs/`, and the `AGENTS.md` triage row. Include the "cannot be scoped below owner permissions" and "no create-PAT API" facts, because both surprise people.

## Definition of done

- Every script that authenticates to Tableau imports from `tableau_env`; no direct `os.environ.get("TABLEAU_*")` outside that module.
- `python scripts/tableau_lineage.py --plan` runs from a `.env` written from `.env.example`, with **no** exported environment variables.
- A test pins the canonical set and **fails if a new script introduces a variant name** — this is the third naming divergence found (`TABLEAU_PAT_VALUE`, `TABLEAU_SERVER`, and whatever is next), so the guard matters more than the fix.
- A test proves the `TABLEAU_SERVER` alias still authenticates and warns.
- `ruff`, `pylint scripts` (exit 0, 10.00/10) and `pytest -q` green. **Check exit codes, not score text** — `pylint` prints `9.99/10` and exits 16.

## Out of scope

JWT, Connected Apps, keyring, OAuth. Not "later" — declined. If a tenant forbids PATs, that is a scoping conversation with the customer, not a feature.

## Related

- #88 — the PAT secret naming mismatch (fixed, #91).
- Dry-run findings F001/F006 in `migrations/dry-run-2026-08-11/findings.md`.

Guust-Franssens:
Fixes #97

`tableau_lineage.py` -- step 2 of the site path AGENTS.md documents -- could not run from a `.env` written from our own `.env.example`. While fixing it, a third fault appeared: the scripts were broken in **opposite** directions.

| script | fault |
|---|---|
| `tableau_lineage.py` | read `os.environ` directly, **ignored .env entirely**; wanted `TABLEAU_SERVER` while everything else uses `TABLEAU_SERVER_URL` |
| `assess_estate.py`, `capture_tableau_oracle.py`, `stamp_tableau_provenance.py` | read only the **file**, so exported variables raised `KeyError` |

`resolve_env()` layers the file over the process environment (file wins, matching `engine_child_env`) and normalises both alias names. `TABLEAU_SERVER` still works, with a `DeprecationWarning`.

**The guard matters more than the fix.** This is the third naming divergence to reach main. The new pinning test earned its place on its first run by catching one I had missed -- `TABLEAU_PRODUCT_VERSION`, read by `stamp_tableau_provenance.py` and absent from my canonical list. It matches environment *access*, not any uppercase identifier, because `make_showcase.py` has RGB constants `TABLEAU_STRIP`/`TABLEAU_ACCENT` and `probe_lab.py` a database `TABLEAU_MIGRATION`.

## Verified by execution, all TABLEAU_* cleared from the shell first

| setup | result |
|---|---|
| canonical `.env` | reaches the network (DNS failure on a bogus host) |
| `TABLEAU_SERVER` `.env` | reaches the network -- back-compat holds |
| exported vars, no file | reaches the network -- was KeyError-shaped before |
| nothing set | names **all three** missing vars, and the HTTP 405 fact |

`ruff` 0 / `pylint scripts` 0 (10.00/10) / `pytest -q` 0 (**696 passed**, was 682).

Also adds `migrations/workshop-2026-08-17/` (plan + customer prerequisites) -- the PAT is the critical path for Monday's workshop and has multi-day lead time outside our control.


Guust-Franssens (commented):
Independent review against PR head `4cb3aabc26afc30ec687a3682f66ea3d5bff0e72` and issue #97. I cleared every `TABLEAU_*` variable before the probes.

## Blocking issues

1. **The consolidation is incomplete, and most live credential consumers still fail with raw `KeyError`s when credentials are absent.**

   `resolve_env()` is wired into the callers, but only `tableau_lineage.py` calls `require()` (`scripts/tableau_lineage.py:222-228`). The other required-auth paths still index the merged dict directly: `scripts/assess_estate.py:78-81`, `scripts/capture_tableau_oracle.py:533-537`, and `scripts/harvest_estate_assets.py:86-93`.

   Reproduction, after clearing `TABLEAU_*`, using an absent env file:

   - `.venv\Scripts\python.exe scripts/assess_estate.py --out <scratch> --env <absent>` -> exit 1, traceback ending `KeyError: 'TABLEAU_SERVER_URL'`.
   - `.venv\Scripts\python.exe scripts/capture_tableau_oracle.py --out <scratch> --env <absent>` -> exit 1, same raw `KeyError`.
   - `.venv\Scripts\python.exe scripts/harvest_estate_assets.py --out <scratch> --env <absent> --db <one-row-db> --limit 1` -> exit 1, same raw `KeyError` after beginning the sweep.
   - The equivalent `tableau_lineage.py --plan --env <absent>` path does print the new actionable PAT instructions, proving the intended behavior exists but was not applied consistently.

   **Impact:** a stranger following the PAT-only setup gets a Python traceback from three primary commands instead of being told which values to provide. This does not meet issue #97's “route every Tableau-auth script through one path” / actionable-error definition of done.

   **Fix:** call `require(env)` immediately after `resolve_env()` in every path that requires server authentication (or make a single resolver API that both resolves and validates), before constructing clients or opening the estate DB.

2. **A PAT can be persisted in `parse-sweep.json` through the new engine bridge.**

   The new flow sends the secret to `fetch_tds.py` via `engine_child_env()` (`scripts/harvest_estate_assets.py:96-103`). On failure, `download()` returns the last 300 characters of the engine's stderr (`scripts/harvest_estate_assets.py:104-105`), and `main()` assigns that verbatim to `row["download_error"]` (`scripts/harvest_estate_assets.py:274-277`). `summarise()` then writes all rows to `parse-sweep.json` (`scripts/harvest_estate_assets.py:224-225`).

   Reproduction command: a local HTTP server returned the sign-in request body as the 401 response, then I called `harvest_estate_assets.download(...)` with `TABLEAU_PAT_SECRET=SENTINEL_HARVEST_REFLECTED_SECRET`. Result: `secret_in_returned_detail=True`, `secret_in_json_result=True`; the captured detail ended with `"personalAccessTokenSecret": "SENTINEL_HARVEST_REFLECTED_SECRET"`.

   This is not theoretical sanitization hygiene: the installed engine raises with the first 500 response characters (`fetch_tds.py:273-278`), while this wrapper deliberately captures and stores that exception text. A proxy/WAF/debug endpoint echoing a request can therefore write the owner's full-permission Tableau token to a durable artifact.

   **Fix:** redact the resolved secret (and preferably PAT name/session tokens) from both stdout/stderr before returning, logging, or serializing engine errors. Add an adversarial echo-response test at the wrapper boundary.

## Non-blocking issues

3. **The advertised pinning guard is trivially bypassed and currently gives false confidence.**

   The regex at `tests/test_tableau_env.py:197` only recognizes a literal immediately after `.get(` or `[`. In a scratch copy outside the repo I added all requested bypasses:

   ```python
   A = os.getenv("TABLEAU_OS_GETENV_BYPASS")
   B = aliased_getenv("TABLEAU_ALIASED_GETENV_BYPASS")
   lookup = aliased_environ.__getitem__
   C = lookup("TABLEAU_ALIASED_ENVIRON_BYPASS")
   D = aliased_environ.get(f"TABLEAU_{suffix}")
   E = aliased_environ.get("TABLEAU_" + "CONCAT_BYPASS")
   F = indirect(**{"name": "TABLEAU_KWARGS_BYPASS"})
   ```

   Command: `.venv\Scripts\python.exe -m pytest -q tests\test_tableau_env.py::test_every_tableau_env_var_used_in_scripts_is_an_accepted_name` in the scratch copy. Exit 0 (`1 passed`); applying the test regex directly returned `regex_matches=[]`.

   There is also a real missed caller now: `scripts/capture_tableau_reference.py:285` still reads `os.environ.get("TABLEAU_SERVER_URL")` directly, ignores `.env`, and is omitted from `scripts/README.md`'s consumer list. Reproduction from a scratch cwd containing a canonical `.env`: `capture_tableau_reference.py <slug> --structural-only --force` exited 0 and fell through as unconfigured; exporting only `TABLEAU_SERVER_URL` made the same command enter the configured server path and exit 3. The guard did not identify this non-consolidated auth/config path because its name happens to be accepted.

   **Fix:** replace the regex with an AST-based rule that flags *any* direct `os.environ`/`os.getenv` access to `TABLEAU_*` outside `tableau_env.py`, and separately assert that credential-consuming scripts import/use the shared resolver. Dynamic names cannot be exhaustively proven by a literal-name allowlist, so do not claim that invariant.

4. **`.env` winning over exported values makes emergency token rotation silently use the stale token.**

   `scripts/tableau_env.py:130-131` merges `{**process, **file}`. Reproduction with a stale file plus freshly exported values resolved the server/site/name and secret from the stale file; `resolved["TABLEAU_PAT_SECRET"] == fresh` was false. `run_estate.py` compounds this by hard-coding `Path(".env")` for provenance (`scripts/run_estate.py:252`) and offering no auth-env override.

   **Consequence:** a user exports a newly rotated PAT specifically to supersede a stale home/project `.env`, but every affected command silently authenticates with the expired/revoked token and reports an auth failure against what the user believes is the fresh token. That is operationally worse than process-env precedence and conflicts with the conventional expectation that explicit per-process configuration overrides a dotenv default.

   **Fix:** use file as defaults and process environment as override (`{**file, **process}`), or add an explicit documented precedence switch and report the selected *source* (never the value). At minimum, do not silently choose the stale file.

## Verified

- Legacy `TABLEAU_SERVER` works: `_env_config(<old env>)` resolved `https://legacy.invalid`; its `DeprecationWarning` went to stderr, stdout was empty, and neither stream contained the secret.
- `--help` for all changed scripts contained no sentinel secret.
- Canonical process vars, canonical `.env`, legacy server alias, and engine `TABLEAU_PAT_VALUE` resolution all worked in direct resolver probes.
- Required gates all exited 0 on the PR: `ruff format --check`, `ruff check`, `pylint scripts`, `pytest -q` (696 passed), and `validate_spec.py --all` (16 valid). The clean `origin/master` worktree also exited 0 on all five gates (670 passed, 4 skipped).

**Verdict: request changes.**

Guust-Franssens:
All four addressed in `1ab3d34`. This was a genuinely good review — the echo-server reproduction in particular found something I would not have found by reading, and finding 3 identified a structural weakness in the guard rather than just an instance of it. Each item below was reproduced before and after.

## 1. Secret leak — fixed

Reproduced your finding, then the fix:

| | before | after |
|---|---|---|
| secret in returned detail | `True` | **`False`** |
| secret in persisted `parse-sweep.json` | `True` | **`False`** |
| detail still names the 401 | — | **`True`** |

```
detail -> ...401: {"credentials": {"personalAccessTokenSecret": "[REDACTED]", "name": "[REDACTED]"}}
```

Added `redact()` and applied it **at the point of capture** in `download()`. Scrubbing there rather than at serialisation is deliberate: the secret is ours, the file is ours, and the engine's error text is not something we control — so this cannot regress if the engine's logging changes.

Your "add an adversarial echo test at the wrapper boundary" is now `test_a_reflected_signin_error_cannot_persist_the_pat`. It also asserts the **diagnostic survives** redaction (`"401" in detail`), because a scrub that destroys the error just trades one failure for another. I also redact the PAT *name*, as you suggested.

## 2. Raw `KeyError` — fixed

You were right that `require()` existed and was wired into exactly one script. All three now call it:

```
$ python scripts/assess_estate.py --out <scratch> --env <absent>
Missing Tableau credential(s): TABLEAU_SERVER_URL, TABLEAU_PAT_NAME, TABLEAU_PAT_SECRET
  Write them to a git-ignored .env (see .env.example) or export them:
  ...
```

One deliberate nuance: `harvest_estate_assets.py` calls it **only when it will download**. `--skip-download` re-parses local assets and needs no credentials — requiring them there would replace a traceback with a wrong refusal.

## 3. Guard bypass and the missed caller — replaced, not patched

This was the most valuable finding, because the missed caller and the bypasses share one cause: **the guard checked names, so an accepted name reading `os.environ` directly passed by construction.** `capture_tableau_reference.py` was exactly that — and its consequence was worse than a lint gap: the "configured-but-unavailable Server must HALT" branch never fired for anyone who used `.env`, the mechanism our own docs recommend. Now routed through `resolve_env(args.env)`, with an `--env` flag.

Replaced with an AST rule over the **access pattern**: no direct environment read of a `TABLEAU_*` key outside `tableau_env.py`. I kept the name guard *alongside* it, since they catch different things — a new **name** read through the resolver, versus **any** name read around it — and neither is sufficient alone.

Building it found one more bypass your list did not include: a bare aliased `environ.get(...)` is an `ast.Name`, not an `ast.Attribute`, so my first version of the check passed it. My own pinning test caught that:

```
AssertionError: 'environ.get(f"TABLEAU_{suffix}")' slipped past the structural guard
```

Your four bypasses are now regression cases. And I took the last part of your fix literally — the docstring states that a dynamically-built name **cannot** be proven safe by either guard, so a pass here is not a claim that every access was seen.

## 4. Precedence — flipped to `{**file, **process}`

Persuaded, on your reasoning rather than convention. Both orders have a stale-source failure, so the tie-break is **recoverability**: a shell export dies with the shell, a `.env` persists indefinitely. Rotation is the case that decides it — someone exporting a freshly issued PAT to supersede a revoked one must not keep authenticating with the revoked token. It also matches `python-dotenv`'s default, which is what users will expect.

I also took your "at minimum, do not silently choose" as the more general point: **the real defect in either order is silence.** Added `env_source()` — returns `"environment"` / `"file"` / `"unset"`, never a value — and log it at sign-in, so an auth failure names the source that supplied the credential.

Not addressed: your note that `run_estate.py` hard-codes `Path(".env")` for provenance. It is a real limitation, but it is a *plumbing* change to an unrelated entry point rather than part of this consolidation, so I would rather file it than widen this PR.

```
ruff format --check   0  (80 files)
ruff check            0
pylint scripts        0  (10.00/10)
pytest -q             0  (709 passed, was 696)
```

Happy to have this re-reviewed by you specifically, since you already have the reproductions built.

Guust-Franssens (commented):
Re-review of PR head `1ab3d34ae22270ef3f578c1507d671fdefbf1f64`. I cleared every `TABLEAU_*` variable before the probes and attacked the fixes rather than only rerunning their happy-path tests.

## Blocking issues

### 1. Secret safety is still incomplete: truncation leaks a PAT suffix, and the oracle can still leak the full PAT

The harvest fix currently does this at `C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\scripts\harvest_estate_assets.py:108-109`:

```python
detail = (proc.stderr or proc.stdout or "").strip()[-300:]
return False, redact(detail, pat_secret(env), env.get("TABLEAU_PAT_NAME", ""))
```

That truncates **before** redacting. I made the secret straddle the 300-character boundary and monkeypatched the child result, using `.venv\Scripts\python.exe -`. The result was:

```
full_secret_in_detail=False
secret_suffix_in_detail=True
detail_starts_with_suffix=True
secret_suffix_in_log_slice=True
```

So the current test passes because the whole sentinel is inside the retained slice, but a real suffix is still both logged by the `detail[:80]` warning and persisted verbatim in `parse-sweep.json`. Redact the complete child output first, then truncate, and pin a boundary-straddling regression.

A second path still leaks the **full** PAT. `C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\scripts\capture_tableau_oracle.py:242,249` stores the first 200 response characters in `last` and appends them to the exception. The same local echo-server technique produced:

```
secret_in_exception=True
... "personalAccessTokenSecret": "SENTINEL_ORACLE_REFLECTED_SECRET" ...
```

Command: `.venv\Scripts\python.exe -` with a local `ThreadingHTTPServer` returning the sign-in request body as its 401 response. This is the same proxy/WAF/debug-endpoint threat model used to justify `redact()` in harvest, so the sign-in exception must also redact or omit the response body.

Edge results: empty input is a safe no-op and regex metacharacters are handled because this is `str.replace`. Values shorter than 8 characters are deliberately left visible (`tableau_env.py:199`), and JSON-escaped quote/backslash forms also survive. Those are not separate blockers if Tableau's PAT format guarantees they cannot occur, but that invariant should be validated rather than assumed by the redactor.

### 2. `require()` still leaves a valid Default-site setup with a raw `KeyError`

`require()` correctly fixed the nothing-set case, but it intentionally does not require `TABLEAU_SITE` because an empty site is valid. Two callers still index it instead of defaulting it:

- `C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\scripts\assess_estate.py:80`
- `C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\scripts\capture_tableau_oracle.py:535`

After exporting server, PAT name and PAT secret but omitting `TABLEAU_SITE`:

```
.venv\Scripts\python.exe scripts/assess_estate.py --out unused --env <absent>
.venv\Scripts\python.exe scripts/capture_tableau_oracle.py --out unused --env <absent>
```

Both exit 1 with `KeyError: 'TABLEAU_SITE'`; `require()` has already passed. Treat absence as the documented Default site (`env.get("TABLEAU_SITE", "")`) or canonicalize it to `""` in `resolve_env()`.

### 3. Process precedence fails when the accepted alias and file use different spellings

`C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\scripts\tableau_env.py:135-143` overlays the raw dictionaries and only then normalizes aliases. Canonical keys therefore beat aliases regardless of source.

Probe: mocked the file as stale canonical values and the process as fresh accepted aliases (`TABLEAU_SERVER`, `TABLEAU_PAT_VALUE`), then called `resolve_env()`:

```
resolved_server=https://stale.invalid
resolved_secret_is_fresh=False
engine_child_secret_is_fresh=False
reported_secret_source=file
reported_server_source=file
```

Command: `.venv\Scripts\python.exe -` using `unittest.mock.patch` on `load_env`. This defeats the token-rotation rationale for anyone exporting the engine's accepted `TABLEAU_PAT_VALUE`, and `env_source()` reports the wrong winner. Normalize each source independently, then overlay normalized file defaults with normalized process values; make source reporting alias-aware. `engine_child_env()` is not independently inconsistent—it faithfully propagates the already-wrong resolved value.

### 4. The AST guard still misses ordinary aliases/copies, recreating the exact accepted-name bypass

The access-pattern visitor at `C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\tests\test_tableau_env.py:186-218` only recognizes calls/subscripts whose immediate object is literally named `environ` (or has attribute name `environ`). I fed the actual helper these forms, all reading accepted `TABLEAU_SERVER_URL`:

```python
from os import environ as E; E.get("TABLEAU_SERVER_URL")
from os import getenv as G; G("TABLEAU_SERVER_URL")
E = os.environ; E["TABLEAU_SERVER_URL"]
E = os.environ.copy(); E.get("TABLEAU_SERVER_URL")
E = dict(os.environ); E.get("TABLEAU_SERVER_URL")
getattr(os, "environ").get("TABLEAU_SERVER_URL")
lookup = E.__getitem__; lookup("TABLEAU_SERVER_URL")
```

Command: `.venv\Scripts\python.exe -`, importing `_direct_environ_reads` from the test module. Every case returned `structural=[]`; because the literal is an accepted name, the companion name guard also returned no unexpected names.

`getattr` is deliberate enough that I would not require a static convention test to catch it. Imported aliases, assignment, `.copy()` and `dict(os.environ)` are normal refactors and matter in practice: they can reintroduce the exact “accepted name but `.env` ignored” defect. Track import/assignment aliases and flag any direct environment-object/getenv reference outside `tableau_env.py`, not only the final `.get`/subscript shape.

## Non-blocking issues

- The precedence documentation was not flipped with the code: `C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\.env.example:3-4`, `C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\scripts\README.md:24`, and `tableau_env.py:31-33` still say the file is layered over the process environment. Only `assess_estate.py` calls `env_source()`; harvest, lineage and oracle remain silent about the winning source. `.github\agents\tableau-migrator.agent.md:159` also still teaches the deprecated `TABLEAU_SERVER` spelling.
- I agree with deferring `run_estate.py`'s hard-coded `Path(".env")` to a follow-up. Provenance is explicitly best-effort, and process variables still provide an override. It should be filed, but it need not widen this PR.
- I found no concrete PAT path through `parse_ours`/`parse_theirs`: the former is in-process, and the installed `connection_to_m.py` has no Tableau environment reads. `parse_theirs` inherits exported credentials unnecessarily, so stripping `TABLEAU_*` from that offline child's environment would still be sensible defense-in-depth.

## Verified fixes

- Nothing-set now gives the actionable multi-variable PAT message in assess, oracle, lineage and harvest.
- `--skip-download` does not call `require()`.
- `capture_tableau_reference.py --env <canonical.env>` now recognizes the configured server path.
- Legacy `TABLEAU_SERVER` still resolves; its `DeprecationWarning` stays on stderr, stdout stays empty, and the secret is absent from both.
- Changed-script `--help` output contains no sentinel secret.
- All required gates exited 0: ruff format, ruff check, pylint 10.00/10, pytest **709 passed**, and 16 specs valid.

**Verdict: request changes.**

Guust-Franssens (commented):
Re-review of freshly fetched, force-pushed head `c7453ff74a2947685f2c4d33cd8dff86e3ea547f`, run from a clean detached worktree. The four round-2 fixes themselves reproduce correctly. I found one remaining secret-leak path.

## Blocking issue

### A reflected authenticated response can expose the Tableau **session token** in exceptions, logs, and `oracle-manifest.json`

The sign-in PAT response is now correctly redacted before slicing. After sign-in, however, every authenticated request sends the equally privileged session credential in `X-Tableau-Auth`, while later response bodies remain unsanitized:

- `C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\scripts\capture_tableau_oracle.py:268` appends raw response text to a `get_json()` exception.
- Lines `293-316` pass raw response text through `classify_export_error()` into `ExportFailed.detail`.
- Lines `381` and `519` persist/log that detail.

I reused the echo-server attack, this time having a local endpoint return the incoming `X-Tableau-Auth` header as its 400 body. Command: `.venv\Scripts\python.exe -` with a local `ThreadingHTTPServer`, a session whose `token` was `SENTINEL_SESSION_TOKEN_FULL_PERMISSION`, then calls to `get_json()` and `export()`.

```
session_token_in_get_exception=True
session_token_in_export_detail=True
```

The export detail is subsequently written into the oracle manifest and printed by the failure summary. A Tableau session token authorizes the same owner-permission session, so this violates the original no-secret-in-errors/logs/artifacts gate even though the PAT itself is now safe.

**Fix:** centralize response-text sanitization in `TableauSession` and redact `self._creds.pat_secret`, PAT name, and `self.token` before *any* server body reaches sign-in errors, `get_json`, error classification, retry details, logs, or records. Add a reflected-header regression beside the reflected-sign-in test.

## Verified fixes

- Harvest now redacts before truncating: the boundary probe found no PAT suffix of 8+ characters in the returned detail or its log slice.
- Oracle sign-in echo: PAT absent, `[REDACTED]` and HTTP 401 diagnostic retained.
- Alias precedence/source reporting: fresh `TABLEAU_SERVER` / `TABLEAU_PAT_VALUE` exports beat stale canonical file values and report `environment`.
- Missing `TABLEAU_SITE` resolves to `""`; `require()` passes and indexed callers no longer raise.
- The four intended AST alias forms are now caught.

## Non-blocking observations

- `parse_theirs()` launches its offline child without an explicit `env`, so an exported PAT is inherited, and its returned JSON is persisted without redaction. I could make a fake child result carry the PAT (`parse_theirs_secret_in_result=True`), but the current installed `connection_to_m.py` does not read Tableau environment variables. Passing a child environment with `TABLEAU_*` removed is worthwhile defense-in-depth, not a current blocker. `parse_ours()` has no subprocess-output path.
- `_normalise()` is idempotent; an empty alias emits no warning. Under `warnings.simplefilter("always")`, `resolve_env()` followed by `env_source()` emits the same alias warning twice, while normal/default filtering reduces this to the single warning visible in pytest. This is test noise, not a correctness issue.
- Defaulting the site means a misspelled runtime variable such as `TABLEAU_SITENAME` silently selects `""`. Defaulting is still better than breaking the documented Default-site case; optionally reject/warn on unknown supplied `TABLEAU_*` names in a follow-up.
- The AST guard is now honestly documented as non-exhaustive. Ordinary forms including `"TABLEAU_SERVER_URL" in os.environ` and `os.getenv(key="TABLEAU_SERVER_URL")` still pass; adding those cheap cases would improve the convention guard, but a pass is no longer claimed as proof.
- `stamp_tableau_provenance.py` does authenticate without `require()`. Its offline/best-effort contract and `run_estate.py` integration make that deliberate and acceptable. Partial configuration is asymmetric (server+name attempts and degrades; server+secret skips); a warning or strict standalone mode can be a follow-up.
- Existing precedence prose in `.env.example`, `scripts/README.md`, and the module overview still says the file is layered over the process environment, and the agent guidance still teaches deprecated `TABLEAU_SERVER`. These remain documentation follow-ups.

## Gates

On clean `c7453ff`: ruff format **0** (80 files), ruff check **0**, pylint **0** (10.00/10), pytest **0** (**702 passed, 4 skipped, 1 deprecation warning**), spec validation **0** (16 valid). The higher count in the main worktree came from untracked/ignored tests, so I used the clean worktree as required.

**Verdict: request changes.**

Guust-Franssens:
Fixed the remaining session-token leak in 9f3ee7e. Reproduced locally with a ThreadingHTTPServer echoing request headers before the change: session_token_in_get_exception=True, session_token_in_manifest=True, diagnostic_status_survives=True. After the fix: session_token_in_get_exception=False, session_token_in_manifest=False, diagnostic_status_survives=True. The fix redacts the current TableauSession token at response-capture time for get_json/export/sign-in paths, and also scrubs reflected X-Tableau-Auth header values from shared redaction used by harvest child output before parse-sweep persistence.

PR #104, opened 2026-08-12
Emit a connections manifest: what the customer must connect, and what breaks until they do

### LINKED ISSUE #100: Emit a connections manifest: tell the customer which data sources they must connect, and what breaks until they do
**Decision (owner, 2026-08-11):** data-source credentials do not travel with a migrated item, and that is accepted — *"they will need to work with connections anyways."* But then we owe the customer a straight answer to **"which data sources will I have to connect, and to what?"** — as a deliverable, not as something they discover one failed refresh at a time.

## The problem

After deployment, every model needs its connection established in the target workspace. Today the customer learns this by importing everything, hitting a refresh failure, opening the model, reading the connection string, and repeating. For a 38-workbook estate that is dozens of round trips through the portal.

Worse, the information is **already in our hands at parse time** — we simply never present it as a list. The dry run made this concrete: 11 published datasources across Snowflake, Databricks and Tableau's own admin sources, plus embedded extracts, and nothing anywhere says so in one place.

## What we already know (so this is assembly, not discovery)

| source | what it holds |
|---|---|
| `estate.db` `datasource` / `upstream_table` | datasource name, project, certified flag, extract state, upstream table names |
| `estate.db` `dependency` | which workbooks bind to which datasource |
| `scripts/connection_target.py` | what a partition should actually connect to — extract folder vs live source |
| `scripts/preflight_source_credentials.py` | classifies which sources are **live** and so need a credential |
| engine handover `storage_decision` | Import / DirectQuery / needs-decision, per datasource |

Every field exists. Nothing joins them into one artifact aimed at a human.

## Scope

**1. Emit a connections manifest** — `connections.md` (human) plus `connections.json` (machine), per estate:

- **Connection identity**: server/host, database, warehouse, schema where known.
- **Connector type**: Snowflake, Databricks, SQL Server, flat file, extract-only…
- **Auth expected**: what kind of credential the customer must supply. Never a value.
- **Blast radius**: which models and which reports break until it is connected — this is what turns a list into a priority order.
- **Storage mode**: Import / DirectQuery / needs-decision, from the engine's own decision.
- **Extract-only sources**, flagged distinctly: there *is* no upstream connection; the data was materialised from a `.hyper`, and the model is a **snapshot** that will never refresh. Customers consistently mistake these for broken connections.

**2. Order it by impact.** A published datasource feeding 12 workbooks is not the same task as one feeding a single archived report. We already compute the dependency graph — use it.

**3. Never emit a secret.** Host, database and account name are configuration; passwords, keys and tokens are not. The manifest must be safe to email, and a test should prove no credential-shaped value reaches it.

**4. Produce it at assessment time, not just after deployment.** Its highest value is *before* anyone commits: it is the honest answer to "what will this migration ask of my platform team?" It should be reproducible from a bundle alone, so it can also be regenerated at hand-off.

## Definition of done

- One command produces `connections.md` / `connections.json` from an estate bundle.
- Every model in the bundle appears exactly once, with its connector, its expected auth, and the reports it affects.
- Extract-only models are unmistakably distinguished from live ones needing credentials.
- Blast-radius counts reconcile with `estate.db`'s dependency table — assert it in a test rather than trusting the join.
- A test proves no secret-shaped value is ever written.
- The document is comprehensible to a platform engineer who has never heard of Tableau.

## Why this matters more than it looks

The customer feedback that motivated the dry run showed migrations stalling not on translation quality but on **operational unknowns** — nobody could say what the migration would require of them. A connections manifest converts the largest of those unknowns into a checklist, before anyone commits capacity.

## Related

- Landing-zone deployer (companion issue) — promotion re-establishes connections in the target workspace; this manifest is what makes that tractable.
- `scripts/preflight_source_credentials.py`, `scripts/connection_target.py` — existing pieces to assemble, not replace.
- #95 — extract materialisation; extract-only sources are exactly the cohort that must be labelled as snapshots here.

Guust-Franssens:
Fixes #100

Credentials do not travel with a migrated item. Today the customer learns which sources need connecting by importing everything, hitting a refresh failure, opening the model, reading the connection string, and repeating. The information was already in our hands at parse time — it was just never presented as a list.

Assembles what exists rather than collecting anything new: `migration_bundle` for the sources, the engine's handover slices for the dependency graph, `preflight_source_credentials.classify_source` for the verdict. It deliberately holds **no classification policy of its own** — that duplication has already failed open once in this repo.

## Three refusals

1. **Never emits a secret.** Connection fields are allow-listed, with a credential-shaped-key pattern as a second alarm, so a future field carrying a token cannot reach a document meant to be emailed. Tested against ten credential spellings and end-to-end through the rendered markdown.
2. **Never calls an extract 'connected'.** A materialized `.hyper` is a **snapshot** — no credential exists and it will never refresh. Customers otherwise hunt for a credential that does not exist. Distinct from an extract *of a live system*, which still needs the upstream connected; a test pins that fail-safe direction.
3. **Never tells you to connect to `sqlproxy`.** That is Tableau's front end for a published data source, whose upstream is defined server-side — printing the class sends a platform engineer to connect to nothing.

Ordered by **blast radius**, not alphabetically. An absent handover reports impact as **unknown**, never zero.

## Measured on the real dry-run estate (38 workbooks)

27 sources → **16 needing a connection, 10 snapshots, 0 unclassified**.

Building it against real data surfaced two defects a synthetic fixture would not have: a datasource repeats once per consuming workbook (two rows read as two jobs — now collapsed, consumers merged), and `sqlproxy` dominated the 'connect to' column with a value meaning nothing to the reader.

`ruff` 0 / `pylint scripts` 0 (10.00/10) / `pytest -q` 0 (**736 passed**, was 714).


Guust-Franssens (commented):
## Independent review

### Blocking Issues

#### 1. Allowed values and free-text fields can leak credentials into both email artifacts

`C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\scripts\connections_manifest.py:59-86` only filters by key, then copies the value verbatim:

```python
SAFE_CONNECTION_FIELDS = ("class", "server", "dbname", "warehouse", "schema", "service", "port")
...
out[field] = str(value)
```

`C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\scripts\connections_manifest.py:128-145` also emits `name`, `why`, and `used_by` without sanitization. URL userinfo, credential query parameters, ODBC/JDBC strings, source names, classification reasons, and workbook names therefore bypass `SECRET_KEY_PATTERN`. JSON has additional leak surfaces (`schema` and `why`) that Markdown does not render.

Exact reproduction (exit 0):

```powershell
$env:PYTHONUTF8='1'
& 'C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\.venv\Scripts\python.exe' -c "import json,sys; from pathlib import Path; from types import SimpleNamespace; sys.path.insert(0,r'C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\scripts'); import connections_manifest as cm; source={'name':'Sales SOURCE_NAME_MARKER','connection':{'class':'snowflake','server':'https://user:URL_PASSWORD_MARKER@acct.example/?password=QUERY_TOKEN_MARKER','dbname':'Driver=X;Pwd=ODBC_PASSWORD_MARKER','schema':'SCHEMA_SECRET_MARKER','powerbi_target':'live_source','powerbi_target_reason':'REASON_SECRET_MARKER'}}; cm.load_bundle=lambda _:SimpleNamespace(path=Path('bundle/migration-spec.json'),kind='migration-spec',data_sources=[source]); cm.blast_radius=lambda _:{'Sales SOURCE_NAME_MARKER':['Workbook WORKBOOK_NAME_MARKER']}; m=cm.build(Path('.')); md=cm.render(m); js=json.dumps(m); markers=('SOURCE_NAME_MARKER','URL_PASSWORD_MARKER','QUERY_TOKEN_MARKER','ODBC_PASSWORD_MARKER','SCHEMA_SECRET_MARKER','REASON_SECRET_MARKER','WORKBOOK_NAME_MARKER'); [print(x,'markdown='+str(x in md),'json='+str(x in js)) for x in markers]"
```

Observed:

```text
SOURCE_NAME_MARKER markdown=True json=True
URL_PASSWORD_MARKER markdown=True json=True
QUERY_TOKEN_MARKER markdown=True json=True
ODBC_PASSWORD_MARKER markdown=True json=True
SCHEMA_SECRET_MARKER markdown=False json=True
REASON_SECRET_MARKER markdown=False json=True
WORKBOOK_NAME_MARKER markdown=True json=True
```

Impact: the generated document is not safe to email, and the JSON-only paths are not covered by the current end-to-end test.

Recommended fix: validate/sanitize values, not just keys. Reject or redact URL userinfo, credential query parameters, and DSN properties in every emitted string; apply the same fail-closed sanitizer to names/reasons/workbooks; and independently assert the complete serialized Markdown and JSON contain none of those forms.

#### 2. The real bundle loses connection target/auth/storage data and contains an alarming wrong instruction

The manifest projects only the top-level connection (`connections_manifest.py:125-142`) and never walks `connection["connections"]`. It also allow-lists `dbname`, while both parser and engine contracts normalize the field as `database` (`parse_tableau.py:239`, `migration_bundle.py:160-164`). `_connection_summary()` repeats the `dbname` mismatch at `connections_manifest.py:177-180`.

Command: an in-memory `load_bundle()`/`build()` inspection of `C:\tfmig\dryrun\tableau-to-powerbi-migration\_sweep\phase2a_bundle_2126` reported:

```text
federated_sources 14
nested_legs 16
server available_on_legs 4 emitted_top_level 0
database available_on_legs 6 emitted_top_level 0
warehouse available_on_legs 1 emitted_top_level 0
schema available_on_legs 4 emitted_top_level 0
safe_connection_keysets Counter({('class',): 26})
```

The rendered customer document consequently says only ``federated`` for multi-system sources instead of naming their Azure SQL/Databricks/Snowflake targets. It also omits available `auth_method` and every handover `model_facts.storage_mode`, despite issue #100 requiring expected auth and Import/DirectQuery/needs-decision.

Worse, the rendered `World Indicators` row appears under **Connect these**. The command

```powershell
rg -n 'connection_class|flatfile_data|connector|storage_mode|No live upstream' C:\tfmig\dryrun\tableau-to-powerbi-migration\_sweep\phase2a_bundle_2126\handover\World_Indicators.json C:\tfmig\dryrun\tableau-to-powerbi-migration\_sweep\phase2a_bundle_2126\handover\RESTAPISample.json
```

shows that handover describes this source as a packaged `hyper`/`dataengine` extract, landed through `Csv.Document`, `storage_mode: Import`, with “No live upstream rebuild is available.” Telling the platform team it needs a credential is the harmful direction.

Recommended fix: assemble each source from all normalized connection legs plus the handover storage decision; emit sanitized `database`/server/warehouse/schema and auth type per leg; and let explicit extract/storage evidence override the engine telemetry wrapper class where no live upstream exists.

#### 3. Parser-tier federated sources can fail open as snapshots

`connections_manifest.py:125-127` calls `classify_source()` once on the parent connection. The actual preflight bundle path deliberately uses every leg (`preflight_source_credentials.py:139-154`). If the first leg is a flat file and a later leg is Snowflake, the parent stamp is `flat_file` and this manifest says “snapshot,” even though the helper finds a live leg.

Reproduction input:

```python
mixed={'name':'Mixed','connection':{
    'class':'excel-direct','mode':'extract','powerbi_target':'flat_file',
    'connections':[
        {'class':'excel-direct','mode':'extract','powerbi_target':'flat_file'},
        {'class':'snowflake','mode':'extract','server':'acct','powerbi_target':'live_source'},
    ],
}}
```

Command: the same in-memory `cm.load_bundle = lambda ...; cm.build(Path('.'))` harness used above, plus `preflight_source_credentials._classify_legs(mixed, 0)`. Observed (exit 0):

```text
MIXED_MANIFEST {'needs_credential': 0, 'snapshots': 1, 'status': 'snapshot (extract - no upstream connection)'}
HELPER_LEGS [('Mixed[0] excel-direct', 'no-creds', ...), ('Mixed[1] snowflake', 'needs-credential', ...)]
```

Recommended fix: use a public leg-classification helper and aggregate fail-safe: any live leg => needs credential; otherwise any review leg => review; only all no-creds legs => snapshot.

#### 4. Dedupe collapses distinct sources because it keys on the lossy display projection

`connections_manifest.py:129-132` says:

```python
safe = safe_connection(connection)
key = (name, json.dumps(safe, sort_keys=True))
```

Two same-named sources on the same server but different normalized `database` values collapse because `database` is dropped. The in-memory reproduction exited 0 with:

```text
DEDUPE {'input_sources': 2, 'manifest_total': 1, 'emitted_connection': {'class': 'postgres', 'server': 'db.example'}}
```

This occurs in the supplied bundle too: `World Indicators` has two different full nested connections (`dataengine`/`.tde` versus `hyper`/`.hyper`) but both reduce to `{"class":"federated"}` and become one row. Whether those ultimately represent one customer job must be decided from a stable source identity, not accidental equality after fields were discarded.

Recommended fix: dedupe on a normalized, secret-free identity that includes all non-secret target dimensions and nested legs (or an authoritative source/binding ID), separately from the display projection.

### Checks without findings

- Single-source status mapping is correct: the focused three classification tests exited 0 (3 passed), including the Snowflake `mode=extract` case.
- Real-bundle blast radius reconciled: 38 handover slices; current and union-of-all-binding-shapes both produced 26 source keys / 60 edges, with 0 missing edges and no zero-consumer manifest entry.
- Empty source list renders without crashing; a zero-slice handover reports impact unknown; a nonexistent bundle exits 1 with `FileNotFoundError`.
- Gates by exit code: Ruff format 0; Ruff check 0; pytest 0 (738 passed); `validate_spec.py --all` 0. The exact `pylint scripts` command exited 8 solely because this working tree already contains untracked `scripts/_probe_deploy.py`; rerunning pylint over `git ls-files 'scripts/*.py'` exited 0, so I do not attribute that failure to PR #104.

**Verdict: request changes**

Guust-Franssens (commented):
## Re-review of `b9d70b4` — fixes attacked

Three of the four previous blockers are substantially fixed: the original seven-marker reproduction is clean, all 4/6 real-bundle servers/databases are now emitted (plus auth/warehouse/schema where present), live legs win over extract legs, and the two different `World Indicators` connections no longer dedupe together. I still found two release-blocking failure modes.

### Blocking Issues

#### 1. The sanitizer still leaks credentials from allowed non-string values and quoted values

`sanitize()` claims it accepts any type, but `safe_connection()` stringifies containers and then applies a regex that only consumes a non-whitespace scalar (`scripts/connections_manifest.py:89,94-105,142`). That leaves ordinary Python/JSON representations untouched:

```text
dict-value LEAK ['DICT_MARKER']
list-dict-value LEAK ['LIST_DICT_MARKER']
dict-reason LEAK ['DICT_REASON_MARKER']
quoted-space-secret LEAK ['SPACE', 'MARKER']
schemeless-userinfo LEAK ['SCHEMELESS_MARKER']
```

Command (exit 0): an in-memory `cm.build()` harness against `b9d70b4`, with:

```python
{'warehouse': {'password': 'DICT_MARKER'}}
{'service': [{'api_key': 'LIST_DICT_MARKER'}]}
{'powerbi_target_reason': {'token': 'DICT_REASON_MARKER'}}
{'server': 'PassWord = "QUOTED SPACE MARKER"'}
{'server': 'user:SCHEMELESS_MARKER@host'}
```

All markers survived into `connections.json`; the quoted and schemeless cases also survive into Markdown when rendered fields are used. A credential split across two fields (`server='password'`, `database='=SPLIT_FIELD_MARKER'`) also survives, although I would treat cross-field reconstruction as a secondary hardening case rather than the main blocker.

Impact: the deliverable still cannot uphold its explicit “never credentials” guarantee for malformed/adversarial-but-serializable input—the exact fail-safe bar for Monday.

Recommended fix: do not stringify structured values. For an allow-listed scalar field, accept only `str`/numeric scalars (or recursively sanitize structure before serialization); otherwise omit/redact it. Parse credential assignments with quoting support, and handle bare `user:pass@host` or reject it as invalid target syntax. Add complete JSON + Markdown regression cases for dict/list, quoted whitespace, and schemeless userinfo.

#### 2. Blast radius is still keyed only by sanitized display name, so distinct same-named sources inherit each other's consumers

`blast_radius()` reduces bindings to `source name -> workbooks` (`scripts/connections_manifest.py:165-190`), and `build()` assigns that same list to every full-connection entry with that name (`scripts/connections_manifest.py:238-253`). The full connection now prevents a collapse, but it does not preserve which workbook used which connection.

This is visible in the supplied bundle. The two handovers are distinct:

```text
RESTAPISample.json    bound=World Indicators   embedded_classes=['dataengine']
World_Indicators.json bound=World Indicators   embedded_classes=['hyper']
```

Yet the generated manifest reports both connection rows as affecting both reports:

```text
dataengine 2 ['RESTAPISample', 'World_Indicators']
hyper      2 ['RESTAPISample', 'World_Indicators']
```

Command (exit 0): `cm.build(r'C:\tfmig\dryrun\tableau-to-powerbi-migration\_sweep\phase2a_bundle_2126')`, printing each `World Indicators` leg class and `used_by`.

Impact: the customer is given a false dependency map and can be sent to remediate the wrong report/source pair. It also double-counts the impact used for ordering.

Recommended fix: carry a stable source identity (or normalized full connection identity) from each handover binding into the blast-radius key. If the engine contract supplies only a name for `bound_datasource`, enrich it from that same handover's `embedded_datasources`; when identity cannot be resolved, mark consumers/impact ambiguous instead of assigning the name-wide union to every row.

### Non-Blocking Issues

- `EXTRACT_ENGINE_CLASSES = {'hyper', 'dataengine', 'tde'}` appears sufficient for observed/public TWB extract engine classes. The mixed live case is safe: a `hyper` leg plus Snowflake returns `needs-credential`; `mode=extract` on Snowflake also returns `needs-credential`. However, all-extract or extract+flat synthetic leg sets are returned as `needs-credential` before the extract review check (`scripts/connections_manifest.py:208-220`). That is conservative rather than fail-open, but the resulting instruction can be misleading. Classify extract-engine legs specially before aggregating ordinary leg verdicts.
- The top-level manifest `bundle` string at `scripts/connections_manifest.py:262` is not sanitized. A path containing `;token=...` leaks to JSON. This is unlikely in the supplied production path, but it contradicts “every emitted string”; sanitize or omit it.
- `monkey_key` and `passwordless.example.com` are preserved unchanged; I did not find the requested over-redaction.

### Real-bundle and gate results

- Real bundle: 27 source rows; 15 need credentials, 10 snapshots, 2 review. Every available server/database was emitted: server 4/4, database 6/6; warehouse 1/1, schema 4/4, auth_method 5/5. No remaining `federated`-only target row.
- Rendered document no longer tells the customer to connect to `hyper`/`dataengine`; both `World Indicators` entries are under **Needs a look**.
- Original leak markers and added `warehouse`, `service`, `port`, `auth_method`, mixed-case/whitespace, bare `?token=`, bare `;PWD=`, int, and `None` cases passed.
- `.venv\Scripts\python.exe -m ruff format --check scripts tests .github/skills` — exit 0.
- `.venv\Scripts\python.exe -m ruff check scripts tests .github/skills` — exit 0.
- `.venv\Scripts\python.exe -m pylint scripts` — exit 0, 10.00/10.
- `.venv\Scripts\python.exe -m pytest -q` — exit 0, **738 passed, 4 skipped**. This checkout does not contain the stated 750-test baseline; no test failed.

**Verdict: request changes**

Guust-Franssens:
Fixed the two re-review blockers on \eat/connections-manifest\ in commit 16b9666. Sanitizer now structurally redacts dict/list values plus quoted and schemeless credentials without redacting \passwordless.example.com\ / \monkey_key\. Blast radius now uses handover connection identity when available and reports ambiguous counts as unknown instead of name-wide unions. Gates: ruff format 0, ruff check 0, pylint 0, pytest 0 (757 passed). Real bundle CLI check exit 0; Multi Source Probe still shows real SQL/Databricks/Snowflake targets and World Indicators dataengine/hyper are routed to review with one report each.

PR #105, opened 2026-08-12
Add pbi-deployer: land an estate in a Fabric landing zone, models first

### LINKED ISSUE #98: pbi-deployer: land a whole estate in a single Fabric landing-zone workspace, verify, then promote by re-import
**Decision (owner, 2026-08-11):** a migration lands **everything** — semantic models and reports — into a **single Fabric landing-zone (LZ) workspace** with no production IAM. Once verified, the content is promoted into a properly-secured target workspace. This issue is the deployer that does it: `pbi-deployer`, which `AGENTS.md` currently calls "phase 2 and does not exist yet".

## Why a landing zone

Today the toolkit's honest answer to "where does the output go?" is *local PBIP only*; getting it into Fabric is a manual `fab import` or a Desktop publish. That is fine for one workbook and untenable for an estate of 38.

The landing zone separates two concerns that are currently tangled:

- **Does it work?** — verified in a throwaway workspace where nobody cares who can see what.
- **Who may see it?** — decided once, in the target workspace, against content already known to be correct.

Getting IAM right *before* you know the content is correct means re-doing it after every fix round.

## Promotion is a re-import, not a move — and that is the key design fact

**Fabric has no move-item-between-workspaces API.** Verified: copy-and-recreate only, and a copied report must be **rebound** to the copied model. The usual workarounds (deployment pipelines, `semantic-link-labs` metadata copy) exist because most people do not have the source artifacts.

**We do.** The PBIP is ours, on disk, under the three-location rule (`migrations/{workbooks,datasources}/<slug>/fabric/` is the deliverable). So promotion is simply **importing the same artifacts into a second workspace**. Consequences worth stating plainly:

- The LZ is **disposable**. We never read content back out of it. Delete it after promotion.
- Promotion cannot drift from what was verified — it is byte-identical input.
- No deployment pipeline, no pairing, no metadata copier.

## Hard limits that shape the design

| limit | value | consequence |
|---|---|---|
| Items per workspace | **1,000**, enforced (folders do not count) | A workbook is ~2 items (SemanticModel + Report) plus 1 per datasource; ~87 items for a 38-workbook estate. One LZ holds **~450–500 workbooks** before the cap bites. |
| API rate limiting | HTTP **429** + `Retry-After` | Creating ~90 items back-to-back will hit it. Batch and honour `Retry-After`; never retry blind. |
| Data-source credentials | do **not** travel | Connections are re-established in the target workspace. Accepted for v1 — see the connections-manifest issue. |

## Scope

**1. `scripts/deploy_estate.py`** (name negotiable):

```
python scripts/deploy_estate.py --bundle <dir> --workspace <LZ name-or-id> [--create] [--dry-run]
python scripts/deploy_estate.py --promote --from <LZ> --to <target> --bundle <dir>
```

- **Model-first ordering.** Datasource models before the workbooks that bind to them, reusing the order `estate_survey.py`/`assess_estate.py` already compute. A report imported before its model binds to nothing.
- **Idempotent.** Re-running updates rather than duplicating; a half-finished run resumes.
- **`--dry-run`** prints the item plan, the folder plan and the projected item count against the 1,000 cap, and changes nothing.

**2. Binding integrity.** The byPath `definition.pbir` that works locally must resolve in Fabric. A shared datasource is referenced by several reports; after import each report must point at the *workspace* model, not a relative path. This is the single most likely thing to break — treat it as the primary test, not an afterthought.

**3. A verification gate before promotion.** Promotion is a deliberate act, so require evidence: every item present, every report bound, every model refreshable (or explicitly recorded as awaiting credentials). Refuse to promote on unmet gates unless overridden, and record the override.

**4. Report what it did.** A deployment manifest — item name, type, folder, workspace, timestamp — so a later finding is traceable to an import.

## Definition of done

- An estate bundle deploys to an empty LZ workspace, and every report opens **bound to the right model**.
- Promotion to a second workspace produces identical content, verified by comparing item inventories rather than by asserting it.
- 429 is handled via `Retry-After` and proven by a test, not by hoping.
- The 1,000-item cap is checked **before** starting, not discovered at item 1,000.
- `--dry-run` output is trustworthy enough to review before a real run.

## Open questions for the implementer

- **Workspace creation**: do we create the LZ ourselves (`--create`) or require it to exist? Creating means owning capacity assignment.
- **`fab` vs `az rest`**: our convention prefers `fab`, but folder support (see the hierarchy issue) appears absent from `fab 0.1.10`. A split implementation is acceptable if documented.
- **Cleanup**: delete the LZ after promotion, or leave it for audit? Suggest leave, with an explicit `--cleanup`.

## Related

- Tableau hierarchy → Fabric folders (companion issue) — the LZ layout.
- Connections manifest (companion issue) — what must be reconnected after promotion.
- #57 (deployer topology) — this supersedes the open question there.
- `AGENTS.md` "Where does the output go?" must stop saying local-PBIP-only once this lands.

Guust-Franssens:
Fixes #98 (partially — the vertical slice; promotion and the multi-workspace pool remain)

The last mile had no tool. Getting a migrated estate into Fabric was a manual `fab import` per item — fine for one workbook, untenable for 33.

## Three facts, each measured against a real tenant

Each silently produces a broken or absent deployment.

**1. A migrated PBIP binds its report to its model BY PATH**, which the service cannot resolve. The report must be rebound to `byConnection` *after* the model exists — which is what makes model-first ordering a requirement, not a preference.

**2. The `byConnection` shape nearly every source gives you is the wrong one.** The five-field form is PBIR schema **1.0.0**; schema **2.0.0** declares `additionalProperties: false` and allows exactly `connectionString`, so the model guid travels *inside* it. Three failed deploys to establish, each with a precise error:

| attempt | result |
|---|---|
| five-field `byConnection` | `Workload_FailedToParseFile`, naming each extra property |
| `connectionString` only | `InvalidConnectionInformation` |
| **+ `semanticModelId=<guid>`** | **Succeeded** |

**3. `202 Accepted` tells you nothing.** Create returns an empty body and a FAILED operation is indistinguishable from success until `/operations/{id}` is polled. My first probe reported the model deployed and the report fine; the report had failed.

## Crash safety, and a disproved assumption

A run journal records **intent before the mutation**, keyed to a hash of the exact definition deployed. Intent-first closes the gap between calling and recording; the hash closes the worse case — an item that exists but uploaded partially, where a name-only check answers *present* and a resume silently ships it broken.

A **run lock** prevents overlapping deploys. Not hypothetical: I believed a run had finished (`Select-Object -First` truncated the pipeline and let the shell exit while Python kept running), started a second, and the workspace ended with **duplicate models and reports**. That also disproved a documented assumption — **Fabric does not reject a repeated Report/SemanticModel name**, so the journal and lock are the only protection.

## Verified end to end against a real landing zone

33 workbooks → 66 items; every report bound to its model. A re-run **skipped every unchanged item by content hash** and deployed only the remainder. Two reports failed with `invalid package content stream` and the run **exited 1 naming them**, rather than exiting 0 with a partial estate.

`--dry-run` reports the plan and the **item count** — the number to agree with a customer before deploying, since each item carries a cost in their capacity/licensing terms.

Targets an **existing** workspace only; creating one means choosing a capacity, which is not our decision.

`ruff` 0 / `pylint scripts` 0 (10.00/10) / `pytest -q` 0 (**721 passed**, 19 new).


Guust-Franssens:
Blocking review of head `04dcd6b` (GitHub does not permit the PR author account to submit a formal “request changes” review).

### 1. Crash resume reissues create instead of resuming the LRO

Quoted lines:

```py
436: journal.intent(item, "create")
437: status, headers, resp = _post_item(...)
438: operation = headers.get("x-ms-operation-id")
...
450: state, op_body = await_operation(location, tok)
454: journal.outcome(item, "Succeeded", item_id, operation)
```

```py
705: if target.journal.unfinished(model):
706:     LOG.info(... "reconciling")
707: state, model_id, detail = create_item(...)
```

Inline monkeypatch probe against this head:

```text
KILL_AFTER_202_JOURNAL [('intent',)]
UNFINISHED_RESUME ['POST create']
```

A kill after the server accepts the POST leaves no operation id in the journal. `Failed`/local `Timeout` outcomes also leave the prior intent pending, so resume can duplicate an operation that later succeeded. Add a flushed/fsynced `submitted` record containing `operationId`/`Location`; resume must poll it. If no operation id was received, reconcile remote truth before mutating. `Timeout` must never mean “create again”.

### 2. Existing, changed, or tampered items are never updated or verified

The only mutation is `POST .../items` (lines 415-425). `already_deployed()` compares only local bytes with the journal (lines 319-328), and lines 731-734 accept `AlreadyExists` for a report without checking its definition/binding.

Commands:

```text
rg 'updateDefinition|getDefinition' scripts/deploy_estate.py  -> no matches
REPORT_ALREADY_EXISTS_RETURNS None
```

A changed artifact creates a duplicate; a remote stale/tampered item is skipped. Implement remote inventory + `getDefinition` hashing and `updateDefinition`; treat `AlreadyExists` as locate-and-verify/update, not success.

### 3. Journal identity omits workspace and placement

Rows at lines 293-315 contain no workspace/folder, the default is one bundle journal (line 801), and `Item.digest` omits `folder_id` (lines 117-121).

```text
JOURNAL_ROW_KEYS ...  # no workspace/folder
FOLDER_CHANGE_DIGEST_EQUAL True
```

The same bundle deployed to a second workspace can skip models using ids from the first. A flat first run followed by a corrected folder plan remains at root. Key/validate journal state by workspace and item identity, and persist/reconcile intended placement.

### 4. The lock acquisition is raceable

```py
353: if self.path.exists() and not force:
...
365: self.path.write_text(...)
```

A two-thread no-disk probe printed `LOCK_RACE [True, True] 2`: both callers acquired. Use atomic exclusive creation (`'x'`/`O_EXCL`/Windows `CREATE_NEW`), an owner nonce, and owner-checked release. The stale-lock message is actionable, and the lock does cover folder creation; acquisition is the blocker.

### 5. Rebind silently tolerates no PBIR and injects raw names

```py
216: "connectionString": (
217:     f"Data Source=.../{workspace_name};"
218:     f"Initial Catalog={model_name};..."
...
225: return [{**part, "payload": payload} if part["path"] == "definition.pbir" else part for part in parts]
```

```text
NO_PBIR_SILENT True
RAW_CONNECTION Data Source=.../WS;Injected=1;Initial Catalog=Model;Other=2;...
```

Require exactly one `definition.pbir` before POST. Microsoft’s current REST-deployment example needs only `connectionString: "semanticmodelid=[SemanticModelId]"`, avoiding display-name escaping entirely: https://learn.microsoft.com/power-bi/developer/projects/projects-report#report-files. The digest is correctly computed after rebinding.

### 6. Preflight inventory is incomplete and resume budgeting is gross, not net-new

```py
392: status, _, items = call("GET", ... "/items", tok)
395: existing = len(items.get("value", []))
396: budget = 900 - existing
```

Fabric List Items is paginated. Probe: `PREFLIGHT_IGNORES_NEXT_PAGE (True, "'LZ': 100 existing item(s), 800 planned, 0 spare")`, despite an explicit `continuationUri`. This can pass and hit item 1,000 mid-run. Conversely preflight runs before the journal is opened and counts the whole bundle, so a near-cap resume can be refused although most items would be skipped/updated. Paginate item/folder inventories and budget/report net-new creates after reconciliation. List Items also requires only Viewer, so listing does not prove Contributor.

The 900-item arithmetic itself is correct. The folder-aware monkeypatched dry run performed reads only: `DRY_RUN 0 ['GET', 'GET'] []`. It creates nothing, but it still reports every item as “would be created” rather than resume skip/update/create state.

### 7. Folder mapping is not collision-safe or stable

- `_slug()` maps `A-B` and `A_B` to `a_b`; `project_map()` silently overwrites that key (lines 543-585).
- `wanted` comes from a set and is sorted only by depth (line 665). With `PYTHONHASHSEED=1..5`, the unsuffixed `R-D` was assigned to different originals, so a rerun can swap projects among existing folders.
- Lines 610-615 flatten by joining with `" - "`; the probe produced distinct depth-11 paths that both became `.../A - B - C` (`FLATTEN_COLLISION True`).
- `ensure_folders()` returns sanitized-keyed `existing` (line 684), while `_resolve_folders()` indexes the original path (line 821). Probe: `SANITIZED_PLACEMENT_KEY [('R-D',)] False`. `R&D` gets a folder created, but its items go to root.

Use stable workbook/project identity and fail on ambiguity; use deterministic identity-derived suffixes; make flattened tails unambiguous and length-bounded; return `resolved`; paginate List Folders. Full-path reuse is otherwise the correct way to distinguish equal names under different parents.

### Error honesty

The model-id guard at lines 720-722 is correct, and item API failures exit 1. However `discover()` warns and skips a workbook with no model (lines 240-247), and folder placement failures degrade to root while the run can still print “all items deployed”. Validate against the bundle manifest and count omissions/placement failures, or a partial estate can still exit 0.

### Gates (exit codes, head `04dcd6b`)

- `python -m ruff format --check scripts tests .github/skills` → 0
- `python -m ruff check scripts tests .github/skills` → 0
- `python -m pylint scripts` → 0 (`10.00/10`, not 24)
- `python -m pytest -q` → 0 (`755 passed, 4 skipped`)

Ignored untracked `scripts/_make_hostile_projects.py` and `scripts/tableau_env_pat.py`; they are not part of the PR.

**request changes**

Issue #77, opened 2026-08-10
Agent emitted a documented-invalid DAX column name ('City, State') while a sibling agent used the compliant name - and the skill's stated reason for the rule is wrong


Guust-Franssens:
## Summary

Two of our agents, migrating in parallel on the same day, chose **different names for the same fix** — and the one that shipped in `book_6-1-Maps` uses a character Microsoft documents as invalid:

| bundle | column | matches guidance? |
|---|---|---|
| `book_8-1-Dashboards` | `column 'Map Location' = Orders[City] & ", " & Orders[State]` | ✅ |
| `book_6-1-Maps` | `column 'City, State' = IF(OR(...), BLANK(), 'Orders'[City] & ", " & 'Orders'[State])` | ❌ |

`.github/skills/powerbi-semantic-model-gotchas/SKILL.md:312-314` already says:

> ⚠️ **A comma is not a legal DAX column-name character**, so the obvious name `'City, State'` is illegal — name it e.g. `'Map Location'` and **tell the report builder the exact name**.

So the guidance existed, one agent followed it, the other did not, and nothing caught the divergence.

**Provenance: this is ours, not the deterministic engine's.** The engine-only screening run (`_screen`, same engine build, same workbook, byte-identical input) emits **no** such column — checked across all 19 workbooks in that run. Only the two agent-migrated bundles have one.

## But the guidance's *reason* is wrong, which matters more

I expected to confirm the model was broken. It is not. Measured against the shipped `book_6-1-Maps` model (Desktop MSIX 2.157.627.0, pid 26660, msmdsrv port 64196, catalog `ce40cf52-…`), via ADOMD:

| DAX | result |
|---|---|
| `DISTINCTCOUNT('Orders'[City])` *(control)* | ✅ `531` |
| `DISTINCTCOUNT('Orders'[City, State])` | ✅ **`604`** |
| `CALCULATE(COUNTROWS('Orders'), 'Orders'[City, State] = "Akron, Ohio")` | ✅ `21` |
| `TOPN(3, SUMMARIZECOLUMNS('Orders'[City, State]))` | ✅ `Los Angeles, California`, … |
| `COUNTROWS(ALL('Orders'[City, State]))` | ✅ `604` |
| `COUNTROWS(VALUES('Orders'[City, State]))` | ✅ `604` |
[...]

PR #120, opened 2026-08-13
Merge appended gotcha corrections into canonical rules (#116, #117, #119)

### LINKED ISSUE #116: powerbi-semantic-model-gotchas: the M culture rule recommends the exact value a later section proves corrupts data
Found by an independent review of the shipped skill bundles, then confirmed by reading the file.

## Problem

`.github/skills/powerbi-semantic-model-gotchas/SKILL.md` §3 gives three mutually exclusive instructions about M culture:

1. **"Always pass an explicit culture to M type-conversion calls"** - and uses `"en-US"` as the worked example.
2. Immediately after: **a hard-coded `"en-US"` SILENTLY CORRUPTS a legacy `.xls`** - the culture must match the MACHINE, not the model. (Measured: SUM 493x too large, 5,952 of 9,994 dates blank, refreshes green.)
3. Later still: do NOT adopt the host-locale solution; convert to CSV instead.

An agent reads these in order and acts on the first one it hits.

## Why it matters

Rule 1 as written names the exact value that rule 2 documents as data-destroying, and the failure is silent and green. This is the single highest-consequence contradiction in the bundle because the output looks correct.

## Fix

Replace all three with ONE decision tree, e.g.:

- the culture must match the **actual textual representation in the source**, not the model and not a fixed constant
- for legacy BIFF `.xls`, the durable answer is conversion to CSV/XLSX; do not present a hard-coded or host-derived culture as a general fix
- keep the measured evidence (it is the proof), but attach it to the single surviving rule

## Acceptance

Grep the section for `en-US` - it should not appear as a recommended value anywhere, only in the worked example of the failure.

### LINKED ISSUE #117: powerbi-semantic-model-gotchas: force-kill Desktop and blanket modal-closing can discard user work and silently dismiss credential prompts
Found by an independent review of the shipped skill bundles.

## Problem

`.github/skills/powerbi-semantic-model-gotchas/SKILL.md` §7.2-7.3 instruct the agent to:

- close Desktop with a **force-kill** (`Stop-Process -Id <pid> -Force`) rather than a graceful close
- enumerate every descendant window and **close every one where `IsModal` is true**

## Why it matters

The force-kill has a documented and legitimate reason (a graceful close prompts to save, and saving writes the stale in-memory model back over corrected TMDL). That reasoning is sound and should be preserved. But neither rule checks whether there is unsaved work, and the modal sweep is a blanket close rather than an allowlist - so it will equally close credential prompts, save prompts, error dialogs, upgrade prompts and security dialogs without reading them.

The bundle elsewhere insists that a missing credential must be surfaced to a human rather than worked around. A blanket modal-closer is capable of dismissing exactly that dialog and turning a stop-and-ask into a silent failure.

## Fix

- [ ] keep the force-kill rationale, but check for unsaved changes first and ask the user rather than deciding for them
- [ ] change the modal sweep from "close everything modal" to "enumerate and REPORT everything modal, close only an explicit allowlist of known-nuisance dialogs" (the Bing deprecation dialog that motivated the section is a perfect allowlist entry)
- [ ] make it explicit that a credential/sign-in modal is never auto-dismissed - it is escalated

### LINKED ISSUE #119: Two unmerged self-corrections leave contradictory rules in the gotcha bundles
Found by an independent review of the shipped skill bundles.

## 1. Small-multiples sorting - the two bundles disagree

- `.github/skills/powerbi-report-gotchas/SKILL.md` (~L125-130): the only recovery candidate is a rank column plus Sort by Column, and "community reports say even that is unreliable under small multiples"
- `.github/skills/powerbi-semantic-model-gotchas/SKILL.md` (~L265-270): a controlled measurement showing `sortByColumn` DOES survive small multiples - "reordered all 17 categories correctly across all 4 panes"

The later controlled measurement should win. As it stands, an agent reading the report bundle first logs an accepted limitation instead of implementing a fix we have proven works.

## 2. Comma in a column name

Already tracked in #77, which cites `powerbi-semantic-model-gotchas/SKILL.md:312-314` ("a comma is not a legal DAX column-name character").

Additional evidence for that issue: the SAME FILE says the opposite ~17 lines earlier (~L289-296) - "A comma **is legal** in a Tabular object name (verified against a live engine: TOM accepted the create and `'Orders'[City, State]` parses - brackets delimit)". So this is not only a wrong rule, it is a self-contradiction, which explains how two sibling agents reached different conclusions on the same day.

## Fix

- [ ] delete the superseded statement in each pair and keep one canonical conclusion with its evidence
- [ ] when a later measurement overturns an earlier claim, MERGE it rather than appending - these files have accumulated self-corrections by appending, and an agent acts on whichever it reads first
- [ ] sweep both gotcha bundles for any other append-style correction that left the original claim in place

Guust-Franssens:
## What this fixes

Three skill-bundle contradictions (#116, #117, #119) that all share one root cause: a correction was
**appended after** the claim it was meant to overturn, leaving both in place. An agent loading the
skill acts on whichever it reads first, so the appended note never actually corrected anything. Each
pair is now **merged into one canonical rule that carries the measured evidence**, and the superseded
claim is **deleted** — not softened, not left standing with an "actually, see below".

Only the two bundles I own are touched: `.github/skills/powerbi-report-gotchas/**` and
`.github/skills/powerbi-semantic-model-gotchas/**`. `pbip-model-refresh` was left alone (owned by a
parallel agent).

---

### #116 — M-culture: three mutually exclusive instructions → one decision tree

`semantic-model-gotchas` §3 said all of: "always pass an explicit culture, e.g. `en-US`" / "a
hard-coded `en-US` silently corrupts a legacy `.xls` (SUM 493× too large, refreshes green)" / "convert
to CSV instead".

**Survivor / evidence:** the measured corruption wins. The rule is now *"pass an explicit culture
**and make its value match the source's actual textual representation, never a fixed constant**"*,
followed by a decision tree: **(1)** land the sheet as CSV/XLSX upstream = the durable fix (values
never become text, no culture can corrupt them); **(2)** stopgap only if you must parse the `.xls` in
place = detect the OS locale (`GetUserDefaultLocaleName`) and emit *that*, explicitly host-dependent.

**Acceptance met:** `en-US` no longer appears anywhere as a recommended value — only inside the worked
failure example (the `en-BE` host round-trip) and the explicit "do **not** hard-code `en-US`"
prohibition. Verified by grep.

### #117 — force-kill + blanket modal-close

`semantic-model-gotchas` §7.2/§7.3 told the agent to force-kill Desktop and to close **every** modal
it enumerates — which can discard unsaved work and silently dismiss the exact credential prompts the
bundle elsewhere (§5) insists must be escalated to a human.

**Survivor / evidence:** the force-kill's documented reason is **preserved** — a graceful close
prompts to save, and saving writes the stale in-memory model back over the corrected TMDL. What
changed is the behaviour *around* it:
- Force-kill is now scoped to *an instance you opened yourself* to refresh a model you are editing on
  disk (state you can reproduce from `_build/`). Added a ⚠️ **unsaved-work check** (`powerbi-desktop
  status` → `currentFilePath`); if it may hold work you did not create, **stop and ask** before
  killing.
- The modal sweep changes from "close everything where `IsModal`" to **"enumerate and REPORT every
  modal, close only an explicit allowlist"** (the `Bing map visuals are going away` deprecation dialog
  is the canonical allowlist entry). **A credential/sign-in modal is NEVER auto-dismissed** — it is
  surfaced and escalated, exactly as §5 requires. Unknown modals are left open for the caller.

### #119 — two unmerged self-corrections

**(a) Small-multiples sorting.** `report-gotchas` said a model-level Sort by Column is "unreliable
under small multiples"; `semantic-model-gotchas` §4 has a **controlled measurement** showing it
**works**.
**Survivor / evidence:** the later measurement wins. `report-gotchas` keeps its own valid finding
(the *report-layer* `sortDefinition` is ignored once the pane field is in the small-multiples well),
but "community reports say even that is unreliable" is **deleted** and replaced with the measured
model-side fix: a model-level `sortByColumn` keyed to a rank column **reordered all 17 categories
correctly across all 4 panes** on the byte-identical `clusteredBarChart`. It is now framed as *route
to the model owner as the real fix*, **not** an accepted limitation.

**(b) Comma in a column name.** `semantic-model-gotchas` said, ~17 lines apart, both that a comma **is
legal** (verified against a live engine) and that it is **not** legal / **is illegal**.
**Survivor / evidence:** the **measured** claim wins. The engine **accepts** the comma — verified
live: TOM accepted the create, `'Orders'[City, State]` parses (brackets delimit), and every
table-qualified reference evaluates — so it does **not** fail. The false "a comma is not a legal DAX
column-name character / is illegal" assertion is **deleted**. The recommendation to avoid the comma is
**kept**, but its reason is corrected to the evidence-based one: the comma is *documented-invalid*
(DAX naming requirements) and works only by bypassing the validation Desktop's UI enforces —
unsupported territory that external tooling may reject. Emitted column renamed `'City, State'` →
`'Map Location'`.

> **#77 interaction (referenced, not closed):** #77 owns the definitive ADOMD evidence that the comma
> works and the shipped-model column rename. This PR resolves the *skill's* internal contradiction in
> the direction #77 argues for ("the comma works, but is unsupported — prefer a comma-free name"). Two
> things are **left for #77** on purpose: the rename of the shipped `book_6-1-Maps` model column, and
> the **Place-vs-Address `dataCategory`** question (the azureMap point-map bullet uses `Address`, the
> Bing bubble-map bullet uses `Place`) — #77 flags that as an unsettled open question, so I did not
> guess.

---

## Append-style-correction sweep (both bundles)

Grepped both bundles for append-style markers ("corrects an earlier", "an earlier revision claimed",
"actually", "but see below", "turns out", "in fact", "was wrong", "overturned", "unconfirmed").
**Result: the only append-style corrections left *standing* were the three above.** Every other
correction-style note is already properly merged — the old claim is cited only as the refuted error
and the current rule is stated as the truth. Examples confirmed already-correct: the aggregation-enum
`Max` value, the "unconfirmed offline" note, the "Do NOT cite Learn" note, and §6's own `en-US`
self-correction (which quotes the wrong interpretation as a refuted strawman, then lands on CSV — now
consistent with the §3 rewrite). The legacy-`.xls` navigation-key note appears in both §3 and §7.1 but
is a benign **duplication** (identical guidance), not a contradiction — left as-is.

## Things I did not resolve from in-file evidence (flagged, not guessed)

- **Place-vs-Address `dataCategory`** for the concatenated map column — genuinely unsettled per #77;
  left both bullets as-is.
- **Out-of-scope `'City, State'` references** in `.github/pbi.kb/visuals/azureMap.md` and
  `docs/tableau-map-to-azuremaps.md` still name the shipped model column. Those describe the
  `book_6-1-Maps` model that #77 (item 2) owns renaming — flagged for reconciliation under #77, not
  edited here.

## Validation

Prose/markdown only — no `.py` touched, so no Python ritual. Verified by re-grep that no superseded
phrase survives (`unreliable under small multiples`, `is not a legal DAX column-name`, `cannot contain
a comma`, `A comma is not`), that `en-US` in §3 is failure-example/prohibition-only, and that the §4
cross-reference quoted from `report-gotchas` matches the semantic-model text verbatim. Diff is surgical
(88 insertions / 34 deletions across the two owned files; no unrelated reflow or retitling).

Fixes #116
Fixes #117
Fixes #119

### DIFF
--- a/.github/skills/powerbi-report-gotchas/SKILL.md
+++ b/.github/skills/powerbi-report-gotchas/SKILL.md
@@ -120,14 +120,18 @@ These pass `validate` but render wrong. Only a live Desktop screenshot catches t
   undocumented**, which is exactly why it can only be found by rendering — cite the measurement.
   **Consequence for migrations:** when a source tool expresses *both* "this discrete pill forms
   panes" *and* "sort this axis by a measure" (a Tableau `<cols>` pane pill + `<computed-sort>` is
-  exactly this pair), Power BI's native small multiples can honour only one. Decide which property
-  carries more meaning, implement that, and log the other as an accepted limitation — do not assume a
-  structurally intact `sortDefinition` means the sort is live. The only recovery candidate is a
-  numeric rank column plus **Sort by Column** on the category, which is a **semantic-model** change
-  (route it to the model owner), and community reports say even that is unreliable under small
-  multiples. Still retarget the sort onto a field the visual actually projects on an axis before
-  logging the limitation: it costs nothing, it is render-neutral, and it leaves the one encoding
-  that is correct on its own terms, so the order appears for free if Desktop ever honours it.
+  exactly this pair), the **report layer** cannot carry the sort — a `sortDefinition` is ignored once
+  the pane field is in the small-multiples well, whether it names an off-axis or an on-axis field
+  (measured above). **This is NOT an accepted limitation — it has a proven model-side fix.** A
+  model-level **Sort by Column** (`sortByColumn`) on the category, keyed to a numeric rank column,
+  DOES survive small multiples: measured in `powerbi-semantic-model-gotchas` §4 ("a model-level sort
+  SURVIVES small multiples where a visual-level sort does not"), it reordered all 17 categories
+  correctly across all 4 panes on the byte-identical `clusteredBarChart` where the report-layer sort
+  was ignored. So **route the sort to the model owner as the real fix, not a fallback** — a rank
+  column plus `sortByColumn` is the mechanism that works, and it is the model's job, not the
+  report's. Keeping the report-layer `sortDefinition` targeted at a projected field is still a
+  harmless, render-neutral adjunct (belt-and-braces), but do not log an accepted limitation and stop:
+  the sort is recoverable, and we have proven it.
 
 - **A measure's `SourceRef.Entity` must be the measure's HOME table, not the table it aggregates.**
   🟢 verified at query level. `Sales (w/o Category)` = `CALCULATE(SUM('Orders'[Sales]),
--- a/.github/skills/powerbi-semantic-model-gotchas/SKILL.md
+++ b/.github/skills/powerbi-semantic-model-gotchas/SKILL.md
@@ -689,16 +689,34 @@ shape as §5, with no credential anywhere in sight; this source was a local file
 **Rules.** Close Desktop **before** editing TMDL, then reopen. **Do not use a graceful close here:**
 it prompts to save, and saving writes the **stale in-memory model back over the corrected TMDL**,
 silently undoing your fix. For an instance *you opened yourself to refresh a model you are editing on
-disk* — whose entire in-memory state you created and can reproduce from `_build/` — a **force-kill**
-(`Stop-Process -Id <literal pid> -Force`) is therefore the correct close, and it also discards
-live-only MCP objects (probe tables) as a free cleanup.
+disk*, a **force-kill** (`Stop-Process -Id <literal pid> -Force`) is the right close **once you have
+confirmed there is no unsaved in-memory state worth keeping** (the cleanliness gate below); it also
+discards live-only MCP *probe* objects (disposable scratch tables) as a free cleanup.
 
-⚠️ **A force-kill discards ALL unsaved state, not only the stale model — so first make sure there is
-nothing to lose.** Confirm via `powerbi-desktop status` that `currentFilePath` is *your* `.pbip` and
-that this is the instance you opened for this refresh cycle. If it is an instance you did **not**
-open, or one that may hold unsaved work you did not create (a report edit, a UI-side model change),
-do **not** decide for the user — **stop and ask** whether to save or discard before killing it.
-Never force-kill a sibling build's instance (the pid-binding rule).
+⚠️ **Authorise a force-kill by a positive cleanliness check, not by provenance.** Knowing you opened
+the instance, and that `currentFilePath` is your `.pbip`, identifies *which* instance you hold — it
+does **not** prove it is clean; an instance you opened can still accumulate unsaved changes. Read the
+per-instance **`hasUnsavedChanges`** flag that `powerbi-desktop status` returns next to
+`currentFilePath`:
+- **`hasUnsavedChanges: false`** — the in-memory model holds nothing the disk lacks; force-kill is
+  safe and your corrected TMDL loads on reopen. (Normal case: editing TMDL *on disk* does not dirty
+  the Desktop document, it only makes the in-memory copy stale.)
+- **`hasUnsavedChanges: true`** — unsaved in-memory state exists; do **not** force-kill on provenance
+  alone, and do **not** scope the question to work *you did not create* — what matters is whether
+  unsaved state exists, not whose it is. Force-kill only if you can *positively confirm* every unsaved
+  change is already reproducible on disk (re-running this cycle's `_build/` script, and/or an
+  `ExportToTmdlFolder` you have already run, rebuilds it exactly). Otherwise **stop and ask** whether
+  to save/export or discard.
+
+**The realistic source of unsaved work you authored yourself is the modeling MCP:** it writes
+tables/measures/relationships **directly into the live Desktop model**, where they exist only in
+memory (and read as `hasUnsavedChanges: true`) until `database_operations ExportToTmdlFolder`
+persists them to the on-disk TMDL — a force-kill silently destroys any you have not exported. Mind
+the interaction with the graceful-close hazard above: here the on-disk TMDL is the *corrected* copy
+and the live model is *stale*, so neither a save nor an export can rescue in-memory work without
+overwriting your fix; if you genuinely hold both an on-disk correction and unexported in-memory
+changes you need, that is a real conflict — **stop and ask.** Never force-kill a sibling build's
+instance (the pid-binding rule).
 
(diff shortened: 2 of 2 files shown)


Guust-Franssens:
## Review response — #117 blocker fixed (commit f7e4218)

The blocker is legitimate and is now fixed. The old guard checked **provenance** ("I opened it" +
`currentFilePath` matches), which proves *which* instance you hold, not that it is *clean*, and its
stop-and-ask was scoped to work *"you did not create"* — which licensed an agent to silently discard
its **own** unsaved, MCP-authored in-memory model changes. Both are addressed:

- **Positive cleanliness check, named.** The Desktop Bridge already exposes a per-instance
  **`hasUnsavedChanges`** flag on `powerbi-desktop status` (documented in
  `pbip-model-refresh/scripts/refresh_pbip_model.py:224` — *"The Desktop Bridge's view of the running
  instances, including `hasUnsavedChanges`"* — and used there as the authoritative save-state signal).
  The gate now reads it: `false` → clean, force-kill safe; `true` → unsaved state exists, do not
  force-kill on provenance alone.
- **"You did not create" carve-out removed.** The rule now says explicitly: *what matters is whether
  unsaved state exists, not whose it is.*
- **`_build/` reproducibility converted from justification to a precondition to confirm.** Force-kill
  with `hasUnsavedChanges: true` is permitted only if you can *positively confirm* every unsaved
  change is already reproducible on disk (re-run this cycle's `_build/`, and/or an `ExportToTmdlFolder`
  you have already run, rebuilds it exactly); otherwise **stop and ask**.
- **MCP case named explicitly.** The modeling MCP writes tables/measures/relationships directly into
  the live Desktop model, where they read as `hasUnsavedChanges: true` until
  `database_operations ExportToTmdlFolder` persists them — a force-kill destroys any unexported
  change. Added the interaction with the graceful-close hazard: because the on-disk TMDL is the
  *corrected* copy and the live model is *stale*, neither a save nor an export can rescue in-memory
  work without overwriting the fix — a genuine conflict that must **stop and ask**.

Preserved unchanged: the graceful-close hazard (a save overwrites corrected TMDL with the stale
model), the force-kill mechanism, the probe-table cleanup, and the pid-binding sibling rule.

This is a **new commit** (not a force-push) so the SHA you reviewed stays intact and the delta is
reviewable on its own. Scope held: only `powerbi-semantic-model-gotchas/SKILL.md` §7.2 changed;
`pbip-model-refresh/**` untouched. #116, #119a, #119b were confirmed satisfied by the review and were
not revisited. `Fixes #116` / `Fixes #117` / `Fixes #119` remain intact.

PR #121, opened 2026-08-13
Make pbip-model-refresh persist safely, bind to the right instance, and gate on real data

### LINKED ISSUE #113: pbip-model-refresh: frontmatter says persisting is opt-in, but the CLI persists by DEFAULT
Found by an independent review of the shipped skill bundles, then verified directly.

## The contradiction

`.github/skills/pbip-model-refresh/SKILL.md` frontmatter (the text the skill registry surfaces, and therefore what an agent reads when deciding whether invoking this is safe):

> Persisting is opt-in - `--save` - because a persisted cache can make the PBIP unopenable.

But `scripts/refresh_pbip_model.py` has **no `--save` flag**. It has `--no-save`, i.e. **persisting is the default**. There is a comment in the script deliberately arguing for that default ("Forgetting a `--save` opt-in hands the NEXT agent an empty model... The second is worse").

So the default was deliberately flipped in code and the description was never updated.

## Why it matters

An agent that trusts the description runs the tool without `--no-save` expecting a read-only operation, and instead gets `database.tmdl` rewritten, `cache.abf` created/replaced, and potentially a UI Save invoked. The description is exactly the surface an agent uses for routing, so this misleads at the point of decision.

The reviewer additionally flagged that `FileMode.Create` truncates an existing cache *before* `ImageSave` succeeds, and that the compatibility level is changed before persistence is proven - so a mid-failure leaves the project in a worse state than either outcome.

## Fix

Pick ONE and make everything agree - the code comment's reasoning for persist-by-default is sound, so the likely fix is to correct the frontmatter and SKILL.md prose rather than the code. Whichever way:

- [ ] frontmatter, SKILL.md run section, and the `image_save()` docstring all state the same default
- [ ] cache write is staged and swapped atomically, so a failure cannot destroy an existing good cache
- [ ] compatibility-level changes roll back if persistence fails

## Acceptance

A reader of the frontmatter alone can correctly predict whether the project is mutated. Add a test that asserts the documented default matches the parsed argparse default, so they cannot drift again.

### LINKED ISSUE #114: pbip-model-refresh: instance-identity check is fail-open, so a wrong port can have its image written into another project
Found by an independent review of the shipped skill bundles.

## Problem

`.github/skills/pbip-model-refresh/scripts/refresh_pbip_model.py` - `same_model()` and its callers are fail-OPEN in several ways:

- returns success when the cache/model cannot be resolved
- accepts any live model whose schema *contains* the on-disk table names - including a sibling model with a superset schema
- reports that weak subset check as "confirmed"
- `--port` bypasses PID-based port discovery entirely
- falls back to the first semantic-model folder when names do not match

The bundled tests in `tests/test_desktop_instance_binding.py` currently *encode* the fail-open behaviour, so they pass while the defect stands.

## Why it matters

The skill's stated purpose includes strict pid binding precisely so one migration cannot write into another's project. As written, a wrong port serving a structurally similar model passes the check, and its image is written into the wrong project's `cache.abf`. Concurrent Desktop instances are explicitly supported, which makes this reachable rather than theoretical.

## Fix

- [ ] fail CLOSED when identity cannot be established
- [ ] remove `--port` from the mutating path, or require it to equal the port discovered from the PID
- [ ] resolve the model from the PBIP binding rather than `models[0]`
- [ ] if a semantic fingerprint is kept, make it exact (tables + columns + measures), not a subset test, and stop calling a subset match "confirmed"
- [ ] update the tests so they assert fail-closed; a test that locks in the old behaviour is the reason this survived

### LINKED ISSUE #115: pbip-model-refresh: DATA_OK is emitted from an arbitrary first table, so it can certify a model whose real sources never loaded
Found by an independent review of the shipped skill bundles.

## Problem

With no `--table`, both `scripts/probe_desktop_query.py` and `scripts/refresh_pbip_model.py` probe the **first non-hidden table** and report a model-level `DATA_OK`.

That first table can be a parameter table, a calculated table, a small CSV-backed table, or just one of several distinct sources.

## Why it matters

`DATA_OK` is used as a readiness signal - the SKILL.md tells the agent that `DATA_OK` means "you are done". A static local table can therefore certify a model whose actual business tables are empty or whose live source never loaded.

This directly contradicts `.github/skills/powerbi-semantic-model-gotchas/SKILL.md`, which insists on proving a REAL read per live source before any table/measure work, and that `PERSISTED` does not prove the live source loaded. The two bundles disagree about what counts as proof.

## Fix

- [ ] accept a list of canary tables, one per distinct source, and report readiness per table/source
- [ ] never emit a model-level `DATA_OK` from an implicit first-table probe - either require explicit tables or downgrade the verdict wording to name the single table actually probed
- [ ] align the wording with the semantic-model-gotchas rule so one bundle does not undercut the other

### LINKED ISSUE #118: Skill bundles are not self-contained: refresh_pbip_model.py tells users at runtime to run a script that is not bundled
Found by an independent review of the shipped skill bundles, then verified.

## Problem

The bundle's portability claim is that copying the skill folder takes the whole procedure with it - and CI asserts this by copying each folder to a temp dir and running its tests with the source repo unimportable.

That claim is false on the failure path:

- `scripts/refresh_pbip_model.py:657` prints, at runtime, `powershell -File scripts/probe_desktop_credential.ps1 -DesktopPid <pid>`
- `SKILL.md:98` calls that script "the arbiter"
- `scripts/refresh_pbip_model.py:471` references it again in a comment

**`probe_desktop_credential.ps1` is not in the bundle.** Contents are: `SKILL.md`, `scripts/probe_desktop_query.py`, `scripts/refresh_pbip_model.py`, `tests/`.

## Why it matters

The recovery instruction fires precisely when a user is stuck on a refresh timeout - the moment the guidance matters most - and once the bundle is installed as a plugin or copied to another repo, it points at a file that does not exist.

Other bundles have the same class of dangling reference to host-repo paths (`.github/pbi.kb/`, `docs/`, `scripts/capture_powerbi_pages.py`), but those are cited as background reading. This one is an executable instruction.

## Fix

- [ ] bundle `probe_desktop_credential.ps1`, or inline the procedure it performs, or degrade gracefully with a self-contained fallback
- [ ] audit the other bundles for references that are phrased as executable instructions vs optional background, and mark host-repo paths explicitly optional
- [ ] extend the existing portability test so it fails when a bundled script prints a path that does not exist inside the bundle

Guust-Franssens:
## What this fixes

Four filed bugs in the `pbip-model-refresh` skill bundle. Scope is strictly
`.github/skills/pbip-model-refresh/**` — the `powerbi-report-gotchas` /
`powerbi-semantic-model-gotchas` bundles are owned by a parallel change and were
not touched.

### #113 — persist-by-default doc drift + non-atomic write + premature compat bump
- **Docs corrected to match the code, not vice-versa.** The frontmatter said
  persisting was "opt-in via `--save`" while the code persists by DEFAULT
  (`--no-save` opts out). The in-code reasoning for persist-by-default is sound
  (it is the script's stated purpose, and the compat-mismatch hazard that once
  justified opt-in is now aligned away), so I fixed the prose.
- **Atomic write.** The cache is staged to `cache.abf.tmp` and swapped with
  `os.replace` only once it exists and is non-empty (`_staged_image_write`), so a
  failed/partial `ImageSave` can no longer truncate an existing good cache.
- **Compat rollback.** The `database.tmdl` `compatibilityLevel` bump is now
  provisional — `_persist_image` snapshots `database.tmdl` (and the generated-edit
  ledger) and restores them if the write does not land, so compat is never raised
  for a cache that was never written.
- **Anti-drift test** asserts the frontmatter's persistence default equals the
  parsed argparse default, so doc and code cannot silently diverge again.

### #114 — fail-open instance identity
- `same_model()` now **fails closed**: no resolvable model folder, or no TMDL
  tables to fingerprint, returns `(False, …)` instead of "unverified → proceed".
- The fingerprint is now an **exact table-for-table match**. A sibling with a
  **superset** schema (all your tables plus more) is refused, not reported
  "confirmed" — that subset match was the core hole.
- `main()` refuses a `--port` that does not **equal** the pid-derived port, before
  any read or write, so `--port` can never bypass PID-based discovery on the
  mutating path.
- `cache_file()` resolves an ambiguous multi-model folder from the
  `.pbip → report → definition.pbir` `byPath` **binding**, and returns `None`
  (fail closed) rather than the old arbitrary `models[0]`.

### #115 — DATA_OK from an arbitrary single table
- A model-level `DATA_OK` is emitted only when explicit canaries (`--tables`, one
  per distinct live source) all return rows. With no canary named, only the first
  queryable table is probed and the verdict is **downgraded** to
  `TABLE_OK '<table>'`, naming the single table actually read. Applied to both the
  read-only probe and the refresh's own data check.

### #118 — dangling reference broke bundle self-containment
- `probe_desktop_credential.ps1` (the arbiter the refresh names on `TIMEOUT`) now
  ships **inside** the bundle, and `refresh_pbip_model` prints its own bundled path
  (`CREDENTIAL_PROBE`) so the runtime instruction always resolves to a file that is
  present wherever the folder is copied.

## Evidence the tests catch the bugs (not just green)

The existing tests encoded the buggy behaviour, so "green" is not proof. I
reverted the two scripts to their pre-change state and re-ran the suite: **30 of
the new/changed tests FAIL** against the old code, while the **38 unchanged tests
still pass**. Every behaviour change is pinned by a test that fails without the fix.
With the fixes in place the full bundle suite is green (**68 passed**); scripts are
`ruff`-clean and pylint **10.00/10** (tests are not held to 10.00/10, matching CI).

## Where I diverged from the issues' proposed fix (pushback with evidence)

- **#114 fingerprint compares tables AND columns AND measures.** An earlier
  revision of this body claimed "exact table-set, NOT columns+measures" — that
  sentence was stale and the blind review was right to flag the contradiction.
  The code has compared all three since round 2: `same_model` calls
  `_first_table_difference` on the table-name set and `_first_schema_difference`
  on the `(table, column)` and `(table, measure)` pair sets, each case-folded and
  each pinned by its own test. The false-negative risk that stale sentence named
  (RowNumber, date-hierarchy columns) is handled where it belongs — those columns
  are filtered out of the TMDL side before comparison — not by dropping the check.
  Prose corrected; no code change was warranted.

## What I deliberately did NOT change (follow-up)

- **`probe_desktop_query`'s read-only `--port` override** is left as-is. #114's
  "don't let `--port` bypass PID discovery" is scoped to the **mutating** path; the
  probe writes nothing, so an explicit port there is a legitimate escape hatch.
- ~~**Two-copies drift** between the bundled `probe_desktop_credential.ps1` and
  the repo-root original.~~ **Done in round 4** — the root copy is now a
  forwarding shim, so there is one arbiter and it is the fail-closed one.
- **Auditing OTHER bundles for their own dangling references** was explicitly out
  of scope; only this bundle's own dangling reference was fixed.

---

## Round 4 — addressing the blind review

The review returned **request changes** with one reproduced blocker. All six
findings are answered below.

### BLOCKER (Findings 1 + 2) — `_is_complete_abf` validated the wrong file format

Confirmed and fixed. Round 3's predicate asserted a persisted `cache.abf` was a
Microsoft Compound File Binary (magic `D0 CF 11 E0 A1 B1 1A E1`). **It is not.**
Measured over 13 real caches (17,478 B – 60,688,851 B) written by Power BI
Desktop *and* by this toolkit's own ImageSave: **0 of 13 match CFBF**. So the
staged write never swapped, `_persist_image` always returned `False`, and every
run fell through to the legacy UI-Automation save — persist-by-default, the
entire subject of #113, was silently dead. The reviewer's diagnosis of the shape
is exactly right: round 3 replaced a loose-but-correct predicate
(exists + non-empty + mtime newer) with a strict-and-wrong one.

The real format, derived from the corpus rather than assumed, and verified
byte-for-byte on every file:

```
0..99    UTF-16LE "This backup was created using XPress9 compression."   (exactly 100 bytes)
100..101 uint16 pad
102..    per block: uint32 uncompressedBytes, uint32 lengthFromTheMagic,
         magic 2A D7 86 4E, payload
         next header = offset + 8 + length ; the chain ends EXACTLY at EOF
```

Detecting a **partial write** is the whole point of the predicate, so the check
is a full chain walk, not a header sniff. Two properties make truncation
detectable: the chain must land exactly on EOF, and all 60 measured non-final
blocks declare exactly 2 MiB uncompressed while every final block declares less —
so a truncation that happens to fall on a chunk boundary (which a chain walk
alone cannot see) is still caught. Known and accepted false-reject: an image
whose uncompressed size is an exact multiple of 2 MiB. That is a deliberate
fail-closed choice and its cost is a fallback to the UI save, not data loss; it
is documented in the module.

**The predicate now returns a rejection *reason*, not a bool**, and
`_staged_image_write` prints it (`save   : staged cache REJECTED - <reason>`).
The CFBF bug survived a whole round precisely because the predicate could only
say "no", and nothing said why.

**Tests rebuilt on the real format (Finding 2).** `_build_minimal_cfbf` and
`_valid_abf_bytes` are gone, along with the false docstring; `_cfbf_bytes`
survives only as a **negative control**. New fixtures **hard-code** the format
constants rather than importing them from the module under test — a fixture
built from the predicate's own constants can only ever confirm the predicate,
never contradict it, which is the structural lesson of this finding. Added: a
golden real-header test, block-boundary truncation, exhaustive prefix rejection,
rejection-reason printing, and — as the reviewer asked — **three tests that feed
real `cache.abf` files through `_staged_image_write` and assert the swap
happens**. They discover the corpus via `PBIP_REFRESH_REAL_ABF` and **skip** when
it is absent, so the bundle stays portable: `tests/test_skills.py` (copied to a
temp dir, repo unimportable) still passes, 23 tests. The reviewer's inverse
mutation now fails **0** tests, was 5.

### Finding 3 (HIGH) — a failed persist invalidated the good cache via mtime

Confirmed and fixed. `_restore_rollback_snapshot` restored `database.tmdl`
byte-for-byte but with a fresh mtime, leaving the definition newer than the
cache — the skill's own documented cache-discard trigger. Reproduced (`cache
newer than definition` flipping `True → False` with bytes unchanged) before the
fix. `_snapshot_rollback_paths` now captures `(bytes, atime_ns, mtime_ns)` and
the restore replays them with `os.utime(..., ns=…)`; pinned by a test.

### Finding 4 (HIGH) — body-vs-code contradiction on #114

**The body was stale; the code is correct** — see the corrected bullet above.
`same_model` has compared tables, `(table, column)` pairs and `(table, measure)`
pairs since round 2, each pinned by a test. No code change was warranted.

### Finding 5 (MEDIUM) — `DATA_OK → TABLE_OK` broke host-repo gates

Fixed by **decoupling, not weakening**. `--tables` is the refresh **scope**;
`--canaries` (new) is the **verification set**.

| invocation | verdict |
|---|---|
| whole-database refresh + `--canaries a b` all return rows | `DATA_OK` |
| refresh narrowed by `--tables` | `TABLES_OK` — never model-level, however many canaries pass |
| no canary named | `TABLE_OK '<table>'`, naming the single table read |
| `--verify-only --tables a b` | `DATA_OK` — nothing was narrowed |
| any canary empty | `NO_DATA`, which outranks everything |

This closes the trap the reviewer identified: an agent chasing the mandated
`DATA_OK` can no longer get it by adding `--tables a b`, because narrowing the
refresh now *caps* the verdict at `TABLES_OK`. `--tables` alone still supplies
canaries, so existing callers keep their probe — they simply cannot certify a
model-level verdict over a partially-refreshed model. Mirrored in
`probe_desktop_query.py`, where `--tables`/`--table` remain plain aliases because
that script never refreshes.

**Host-repo changes this needs, which I do not own and did not make:**

- `.github/agents/pbi-semantic-builder.agent.md:301` mandates `DATA_OK +
  PERSISTED` from a call **without** `--tables`. With no canary named that call
  now yields `TABLE_OK`, so it must add `--canaries <one per distinct live
  source>`. It must **not** add `--tables` — that would cap it at `TABLES_OK`.
- The two other `SKILL.md` files instructing `-> PREFLIGHT: DATA_OK` need the
  same `--canaries` switch.

### Finding 6 (LOW/MEDIUM) — two same-named credential arbiters, one fail-open

Fixed, since the reviewer is right that it was ~20 lines away. Rather than
duplicating the fix into the root copy (which recreates the drift), the root
`scripts/probe_desktop_credential.ps1` is now a **forwarding shim** to the
bundled arbiter — matching this repo's existing shim convention for
`probe_desktop_query.py`. There is one arbiter again, it is the fail-closed one
(`UNKNOWN` when no Refresh was invoked, never falling through to
`CREDENTIAL_PRESENT`), and `scripts/README.md` / `docs/data-source-credentials.md`
stay correct without edits. Two tests pin it: the bundled arbiter never fails
open, and the root file is a shim rather than a re-grown implementation.

### Also in this round

`_verdict.py` extracted from `refresh_pbip_model.py` to stay under the pylint
module-length cap, and the interprocess-lock work for #114 (`_lock.py`, per-run
staging paths, `BaseException` rollback) that was already in the working tree
lands here too.

### Round-4 verification

- **13/13 real caches accepted**; every truncation of the smallest real cache
  rejected; a real block-boundary truncation rejected with the boundary reason.
- Reviewer's inverse mutation: correcting the format fails **0** tests, was 5.
- Bundle suite **125 passed** with the real corpus, **122 passed + 3 skipped**
  without it. `tests/test_skills.py` portability gate: **23 passed**.
- Gates by exit code: `ruff format --check` 0, `ruff check` 0, `pylint` 0
  (10.00/10), `pytest` 0.
- **Ten mutations injected, ten caught**: CFBF magic restored; final-block rule
  dropped; chain walked past EOF and accepted; staged file swapped without
  validation; rejection reason not printed; `os.utime` removed from rollback;
  `--tables` re-coupled to canaries; the `TABLES_OK` branch disabled; the bundled
  arbiter made to fail open again; the root shim re-grown into a UIA
  implementation.

---

Closes: Fixes #113 Fixes #114 Fixes #115 Fixes #118

Fixes #114

### DIFF
--- a/.github/skills/pbip-model-refresh/SKILL.md
+++ b/.github/skills/pbip-model-refresh/SKILL.md
@@ -192,8 +192,11 @@ data can only have come from the cache.
 Two traps:
 
 - The AMO client raises **"The server sent an unrecognizable response"** *while writing the file
-  correctly*. Judge success by the FILE (exists, non-empty, mtime advanced), never by the absence of
-  an exception.
+  correctly*. Judge success by the FILE, never by the absence of an exception — but "the file" means a
+  **complete** ABF: only that specific benign response is tolerated, and the staged `cache.abf.tmp` is
+  swapped in only after it is validated as a real Analysis Services backup (CFBF/OLE2 magic + a full
+  header), so a truncated or disk-full write is rejected instead of replacing a good cache. Any other
+  exception propagates and the compatibility-level alignment is rolled back.
 - `database_operations ExportToTmdlFolder` persists model *definition* changes but carries no rows.
   It cannot substitute for this.
 
@@ -215,22 +218,29 @@ queries that same wrong port.
 So, before refreshing, querying or saving anything, two guards run. First, **if you pass `--port` it
 must EQUAL the port derived from `--pid`** — a mismatch aborts, so `--port` can never bypass PID-based
 discovery to point the read (and the `ImageSave` write) at another instance. Second, `same_model()`
-compares the connected model's tables against the TMDL that owns the destination cache and requires an
-**EXACT table-for-table match**; a mismatch — or an identity it cannot establish at all (no model
-folder resolved, or no TMDL tables to fingerprint) — aborts with `WRONG_MODEL`, **failing closed**
-rather than assuming it is fine. A superset schema (all your tables plus more) is a mismatch, not a
-"confirmed". File metadata fundamentally cannot tell you whose rows are in a blob, only the model's
-own contents can. When a project folder holds several `.SemanticModel` directories the destination is
-resolved from the `.pbip` → report → `definition.pbir` `byPath` **binding**, never an arbitrary first
-match. In a parallel batch **always pass `--pid`** (`powerbi-desktop status` maps pid to open file);
-with several instances open the scripts refuse to guess.
+compares the connected model against the TMDL that owns the destination cache and requires an **exact
+match of tables, columns AND measures** — table names alone are too coarse (a bound sibling with the
+same table names but different columns or measures would pass), so the fingerprint descends into each
+table's columns and the model's measures, filtering only the auto-generated noise that never appears in
+TMDL (RowNumber system columns, `LocalDateTable_*` / `DateTableTemplate_*` auto-date tables). A
+mismatch — or an identity it cannot establish at all (no model folder resolved, or no TMDL to
+fingerprint) — aborts with `WRONG_MODEL`, **failing closed** rather than assuming it is fine. A superset
+schema (all your tables plus more) is a mismatch, not a "confirmed". File metadata fundamentally cannot
+tell you whose rows are in a blob, only the model's own contents can. When a project folder holds
+several `.SemanticModel` directories the destination is resolved from the `.pbip` → report →
+`definition.pbir` `byPath` **binding first**; only if no binding resolves does a single same-named
+sibling act as a last-resort fallback — the binding is authoritative and the name heuristic can never
+short-circuit it. In a parallel batch **always pass `--pid`** (`powerbi-desktop status` maps pid to open
+file); with several instances open the scripts refuse to guess.
 
 **Sweep for orphans before you build, not just siblings while you build.** Measured 2026-08-01: a
 Desktop instance was already running on the *exact `.pbip` path* about to be generated, left over from
-an earlier, since-deleted attempt. `same_model()` would not have caught it — its TMDL tables matched —
-but `INFO.MEASURES()` showed measures the new build never defines, i.e. a **different model held in
-memory on your path**, one Save away from overwriting the files you are generating. So at the start of
-a build, list `Get-Process PBIDesktop`, read each one's command line
+an earlier, since-deleted attempt. The tables-only fingerprint then in force would not have caught it —
+its TMDL tables matched — though the columns+measures fingerprint now would, because `INFO.MEASURES()`
+showed measures the new build never defines. But do not lean on that: an orphan whose schema is
+byte-for-byte your model (the common re-open case) is **identical** to `same_model()`, one Save away
+from overwriting the files you are generating. So at the start of a build, list `Get-Process
+PBIDesktop`, read each one's command line
 (`Get-CimInstance Win32_Process -Filter "ProcessId=<pid>"`), and force-close any instance already bound
 to your own `.pbip` before writing to it. Identify by `MainWindowTitle` + command line, never by "the
 one instance that is running".
--- a/.github/skills/pbip-model-refresh/SKILL.md
+++ b/.github/skills/pbip-model-refresh/SKILL.md
@@ -40,7 +40,7 @@ A migrated model hands over as *TMDL plus a promise*. Two things go wrong:
 ## Run it
 
 ```
-python scripts/refresh_pbip_model.py [--pid <pbidesktop-pid>] [--tables "A" "B"]
+python scripts/refresh_pbip_model.py [--pid <pbidesktop-pid>] [--canaries "A" "B"] [--tables "A" "B"]
                                      [--no-save] [--verify-only] [--ui-save]
 ```
 
@@ -112,22 +112,32 @@ stay byte-identical.
 Read-only preflight (proves credentials + source reachability without changing anything):
 
 ```
-python scripts/probe_desktop_query.py [--pid <pbidesktop-pid>] [--tables "A" "B"]
+python scripts/probe_desktop_query.py [--pid <pbidesktop-pid>] [--canaries "A" "B"]
 ```
 
-**Name one canary table per distinct live source** with `--tables`. A model-level `DATA_OK` is only
+**Name one canary table per distinct live source** with `--canaries`. A model-level `DATA_OK` is only
 emitted when *every* named canary returns rows; with **no** table named, only the first queryable
 table is probed and the verdict is **downgraded** to `TABLE_OK '<table>'` — a static parameter/CSV
 table can return rows while a live source never loaded, so one arbitrary table can never certify the
 whole model. This mirrors the `powerbi-semantic-model-gotchas` rule: *prove a REAL read per live
-source*. `refresh_pbip_model.py` applies the same rule to its own data check (its `--tables` is both
-the refresh scope and the canary set).
+source*.
 
-Both print a machine-readable last line: `REFRESH: DATA_OK + PERSISTED` / `TABLE_OK '<table>'` /
-`NO_DATA` / `NOT_PERSISTED` / `WRONG_MODEL` / `ERROR <msg>`, and `PREFLIGHT: DATA_OK` /
-`TABLE_OK '<table>'` / `NO_DATA` / `ERROR`. Exit 0 is the good outcome — but it covers **both** a
-model verdict (`DATA_OK`) and a single-table verdict (`TABLE_OK`), so a gate that needs model-level
-certainty must require the literal `DATA_OK` (i.e. pass explicit canaries), not merely exit 0.
+⚠️ **`--canaries` verifies; `--tables` narrows the refresh. They are different knobs, on purpose.**
+They used to be one, and that built a trap: the only documented way to earn `DATA_OK` was to name
+tables in `--tables`, which simultaneously refreshed **only** those tables — certifying a *model*-level
+verdict over a *partially* refreshed model, i.e. the "next agent gets an empty model" failure this
+bundle exists to prevent, wearing a green badge. So `refresh_pbip_model.py` now emits
+`TABLES_OK '<a>', '<b>'` — never `DATA_OK` — whenever `--tables` narrowed the refresh, and
+`--canaries` verifies without narrowing anything. `--tables` alone still supplies the canary set (old
+callers keep a probe), it just can no longer certify the whole model. `probe_desktop_query.py` never
+refreshes, so there `--tables`/`--table` remain plain aliases for `--canaries`.
+
+Both print a machine-readable last line: `REFRESH: DATA_OK + PERSISTED` / `TABLES_OK '<a>', '<b>'` /
+`TABLE_OK '<table>'` / `NO_DATA` / `NOT_PERSISTED` / `WRONG_MODEL` / `ERROR <msg>`, and
+`PREFLIGHT: DATA_OK` / `TABLE_OK '<table>'` / `NO_DATA` / `ERROR`. Exit 0 is the good outcome — but it
+covers a model verdict (`DATA_OK`), a scoped verdict (`TABLES_OK`) **and** a single-table verdict
+(`TABLE_OK`), so a gate that needs model-level certainty must require the literal `DATA_OK` — which
+means running a **whole-model** refresh with explicit `--canaries`, not merely exit 0.
 
 ## Order matters, do not refresh before you finish editing
 
@@ -170,7 +180,7 @@ number you read back.
 
 | Path | What it is | Status |
 |---|---|---|
-| **AMO `Server.ImageSave(databaseId, Stream)`** | Writes `cache.abf` directly from the engine. No UI, works headless; staged to `cache.abf.tmp` and swapped in atomically, so a failed/partial write can't destroy an existing good cache. | **Default (opt out with `--no-save`)** |
+| **AMO `Server.ImageSave(databaseId, Stream)`** | Writes `cache.abf` directly from the engine. No UI, works headless; staged to a per-run private file and swapped in atomically, so a failed/partial write (or a concurrent second run) can't destroy an existing good cache. | **Default (opt out with `--no-save`)** |
 | **UI Automation (`InvokePattern` on the "Save" element)** | Drives Desktop's own Save through the Windows *accessibility* tree. | Fallback / `--ui-save` |
 
 ⚠️ **Read the `--no-save` warning above before using either.** Both write the same `cache.abf`. A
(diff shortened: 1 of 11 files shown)


Guust-Franssens:
## Round 2 — addressing the blind review (3 blockers + 2 majors)

New commit `c743e06` on top of the reviewed `b5ae3a4` (no force-push; the reviewed SHA is intact). Every behaviour change below is covered by a test that **fails against the reviewed code** — verified by two revert experiments (bottom of this comment).

### BLOCKER 1 — a partial cache could replace a good one ✅ fixed
`_staged_image_write` no longer suppresses every exception, and no longer swaps on "merely non-empty".
- The inner `image_save` closure now re-raises anything that is **not** the one benign AMO *"unrecognizable response"* (`_is_benign_imagesave_response_error`, matched narrowly).
- The staged `cache.abf.tmp` is swapped in **only after** `_is_complete_abf()` confirms it is a real Analysis Services backup — CFBF/OLE2 magic (`D0 CF 11 E0 A1 B1 1A E1`) **and** a full ≥512-byte header. A truncated / disk-full write is rejected and the existing cache is preserved.
- The completeness check is deliberately conservative (magic + min header): rejecting a *valid* cache would be worse than the disease, so it only rejects what is provably not a backup.

Covering tests (all fail against the reviewed code): `test_a_clean_write_of_incomplete_bytes_is_rejected`, `test_a_raising_write_with_no_output_propagates_and_leaves_the_cache_intact`, `test_a_raising_write_that_produced_partial_bytes_does_not_replace_the_cache`, `test_benign_imagesave_error_is_recognised_but_a_real_failure_is_not`, `test_a_successful_write_swaps_the_new_cache_in`.

### BLOCKER 2 — compat rollback skipped on exception paths ✅ fixed
`_persist_image` now wraps the whole **align → write → replace** sequence in `try/finally` with a `committed` flag; the `compatibilityLevel` snapshot is restored unless `os.replace` **definitively committed**. Alignment, stale-temp deletion, `stat()`, and a Windows `os.replace()` sharing-violation all roll back now.

Covering test: `test_persist_rolls_back_the_compat_bump_when_os_replace_raises` (monkeypatches `os.replace` → `PermissionError`). Against the reviewed code it reproduces your finding exactly — `DID NOT RAISE` and the model left at `1604 -> 1702` with no rollback.

### BLOCKER 3a — name heuristic short-circuited the authoritative binding ✅ fixed
`cache_file` now consults the `.pbip → definition.pbir` `byPath` **binding first**; a same-named sibling is only a **last-resort fallback** when no binding resolves. The binding can no longer be bypassed by a same-named sibling.

Covering test: `test_cache_file_resolves_the_bound_model_when_several_exist` — rewritten per your note so the bound model (`Zed`) sorts **second** and a decoy (`Proj`) matches the `.pbip` stem. It now genuinely fails against the reviewed code (which returned the stem decoy).

### BLOCKER 3c — table-only fingerprint was the only control on the name-match path ✅ fixed (and I concede the #114 pushback)
You were right and my round-1 pushback was wrong. Since 3a proves the binding is **not reached** on the name-match path, the table-only fingerprint was the *only* remaining control there, not defence-in-depth. `same_model()` now compares **tables + columns + measures**.
- Your false-negative concern about auto-generated columns is preserved by **filtering**, not by skipping columns/measures: `RowNumber` system columns (TMSCHEMA `Type == 3`) and `LocalDateTable_*` / `DateTableTemplate_*` auto-date tables are removed on both sides, so a real column/measure drift on a same-named sibling is caught while the auto-gen noise does not cause false `WRONG_MODEL`.

Covering tests: `test_same_model_refuses_when_a_column_differs`, `test_same_model_refuses_when_a_measure_differs`, `test_tmdl_columns_measures_parses_quoted_and_calculated_names`, `test_tmdl_columns_measures_skips_auto_date_tables`, `test_column_names_filters_rownumber_and_auto_date_columns`, `test_measure_names_filters_auto_date_measures`.

### MAJOR 4 — recovery command broke on paths with spaces ✅ fixed
The printed command now quotes the arbiter path: `powershell -File "<abs path>" -DesktopPid <pid>`.

Covering test: `test_timeout_recovery_command_quotes_a_path_with_spaces` — copies the bundle under a `Program Files`-style directory and asserts `-File "<path>"`.

### MAJOR 5 — credential arbiter could report PRESENT without probing ✅ fixed
`probe_desktop_credential.ps1` now enumerates **every** top-level window for the pid (`Get-PidWindows` / `FindAll`, replacing the single-window `FindFirst`), and returns **`VERDICT: UNKNOWN` (exit 3)** when no Refresh control was ever invoked, instead of waiting and asserting `CREDENTIAL_PRESENT`.

Covering test: `test_credential_arbiter_enumerates_all_windows_and_fails_closed` — **structural** (asserts `Get-PidWindows`/`FindAll` present, `FindFirst` absent, the `-not $invoked` guard, and the `UNKNOWN` verdict). ⚠️ **Honest limitation:** this script needs live UIAutomation + Desktop, so there is no CI path that *executes* it; the fix is pinned by source assertions only.

### #115 — untouched
Confirmed satisfied in round 1; not revisited.

---

### What I did **not** change, and why
- **Root `scripts/probe_desktop_credential.ps1`** (the repo-root copy, outside the bundle) still carries the old fail-open logic. It is **out of scope** for this bundle-scoped PR and no CI enforces byte-identity with the bundled copy. → **Follow-up:** sync or delete the root copy.
- **#118 audit of *other* bundles** for dangling arbiter references — explicitly out of scope per the task; only this bundle's own reference was fixed. → **Follow-up.**
- `test_cache_file_uses_the_stem_only_when_no_binding_resolves` is a **design-documentation guard** for the retained last-resort stem fallback; it passes against both baselines because that fallback exists before and after. I flag it as *not* blocker coverage rather than dress it up as such.

### Revert-experiment results (per your explicit requirement)
Two baselines. `origin/master` made no changes to `pbip-model-refresh` vs the branch base, so its copy of these scripts **is** the pristine pre-fix baseline (matching your round-1 "30 fail / 38 pass" reproduction).

| Baseline (scripts reverted to) | Result | Meaning |
|---|---|---|
| **Round-1 reviewed code (`b5ae3a4`)** | **16 failed, 64 passed** | The 16 are *exactly* the round-2 behaviour-change tests (5×B1, 2×B2, 1×B3a, 6×B3c, 1×M4, 1×M5). Amended acceptance tests correctly **pass** — they cover round-1 behaviour present in that baseline. This isolates the round-2 fixes. |
| **Pristine (`origin/master`)** | **42 failed, 38 passed** | Up from round-1's 30/38; the **38 passers are identical** (no accidental coverage loss), failures grew by exactly the 12 new round-2 tests. Every new **and** amended test is in the failure list. |

Restored to round-2 afterwards: **80 passed**. `ruff` clean on all four touched `.py`; `pylint` 10.00/10 on `scripts/`.

Issue #154, opened 2026-08-14
P1: minimizing Desktop permanently destroys credential-dialog evidence - UNKNOWN is not latched, verdict degrades to SLOW_SOURCE


Guust-Franssens:
## Summary

If the Power BI Desktop window is **minimized** while a probe is waiting, the open credential dialog
becomes permanently invisible to the detector. Restoring the window does **not** bring it back. The
detector then reports `none` — indistinguishable from a genuinely healthy Desktop — so the probe waits
the full 330 s and finishes with **`SLOW_SOURCE`**, blaming the warehouse for being slow when the real
cause is a missing credential.

A correct, fast `NO_CREDENTIAL` in **74 s** degrades into a wrong verdict at **330 s**, triggered by an
entirely natural operator action (minimizing a window).

## The good news first — the existing `UNKNOWN` guard is correct, and its assumption is now verified

`_credential_modal.py:277-285` degrades to `UNKNOWN` when no candidate is found *and* the main window
is iconic, with the reason *"Power BI Desktop owner window is minimized; owned credential dialogs are
hidden"*. **That assumption is empirically true.** Measured 2026-08-14 against a real Snowflake
credential dialog:

```
---------- BASELINE - dialog up, nothing minimized ----------
enumerated windows: 3   tier2 candidates: 1
  [other] hwnd=4528190 702x257 minimized=False title='' class=WindowsForms10.Window.20008.app.0.2a59b4f_r10_
  [MAIN] hwnd=33559522 2011x1298 minimized=False title='Probe' class=WindowsForms10.Window.8.app.0.2a59b4f_r10_ad1
  [other] hwnd=2232864 0x0 minimized=False title='' class=Internet Explorer_Hidden
  => PRODUCTION VERDICT: BLOCKED_BY_DIALOG (tier2/geometry)

>>> minimizing MAIN window hwnd=33559522 (SW_MINIMIZE)

---------- MINIMIZED - owner window iconic, dialog still logically open ----------
enumerated windows: 2   tier2 candidates: 0
  [MAIN] hwnd=33559522 159x27 minimized=True title='Probe' class=WindowsForms10.Window.8.app.0.2a59b4f_r10_ad1
  [other] hwnd=2232864 0x0 minimized=False title='' class=Internet Explorer_Hidden
  => PRODUCTION VERDICT: UNKNOWN (Power BI Desktop owner window is minimized; owned credential dialogs are hidden)
```

[...]

PR #155, opened 2026-08-14
fix(#152,#153): make the credential-gate verdict seam observe a successful refresh

### LINKED ISSUE #152: P0: probe_live_source can never observe a successful refresh - the credential gate cannot be lifted by any source
## Summary

`scripts/probe_live_source.py` **can never observe a successful refresh**. Every live source — however
healthy — is reported as a failure, and (because of #B, filed separately) as a confident
`NO_CREDENTIAL` with a "A HUMAN MUST ACT" directive.

Measured live 2026-08-14 against a **working** Databricks source: the child refreshed successfully
(`data   : 1 row(s) in 'shipment'`, child exit code **0**) and the probe still exited 1 with the full
credential-stop banner.

## Root cause — a verdict-vocabulary desync across the parent/child seam

Three facts that are individually correct and jointly fatal:

1. `.github/skills/pbip-model-refresh/scripts/_verdict.py:88-99` — when `--tables` is passed (and not
   `--verify-only`), the child prints **`REFRESH: TABLES_OK '<name>'`** and returns 0. This was
   deliberate and correct, per #115 ("a model-level `DATA_OK` over a partially refreshed model is a
   false certificate").
2. `scripts/probe_live_source.py` — `_has_data_ok_verdict` (L429-440) is an **anchored regex that
   accepts only the literal `DATA_OK`**.
3. `scripts/probe_live_source.py` — the probe **always** passes `--tables <table>` and **never**
   passes `--canaries`.

So the only verdict the probe can ever *elicit* is the one verdict it cannot *accept*. There is no
source, no credential, and no configuration that lifts the gate.

Aggravating factor: `_refresh_and_classify` **ignores the child's exit code entirely** and classifies
on stdout text alone. That is what let the two modules' vocabularies drift apart silently — the child
was already telling the truth in a channel nobody read.

## Reproduction (offline — no Desktop, no network, no credentials)

Feed the production classifier the verbatim captured child output:

```
child printed : REFRESH: TABLES_OK 'shipment'   (child exit code was 0)
_has_data_ok_verdict(happy output) -> False
expected      -> True   (a row came back: "data   : 1 row(s) in 'shipment'")
```

## Why CI did not catch it

`tests/test_probe_live_source_verdict.py::test_accepts_genuine_data_ok_verdict` hand-writes its
fixture:

```python
stdout="  data   : 1 row(s) in 'Orders'\nREFRESH: DATA_OK\n",
```

That is output **the production caller cannot produce**, because it always passes `--tables`. Both
sides of the seam are individually tested and individually correct; the seam itself is untested, with
a fixture that contradicts the real emission. The suite is green (1343 passing) while the feature is
non-functional.

## Regression window

`_verdict.py`'s `TABLES_OK` branch arrived in **9b69f93 (#121, 2026-08-13)**. The last recorded green
live-probe result predates it (2026-08-05, Databricks `DATA_OK` in 21 s). So the gate has been
un-liftable since #121 and nobody re-ran a happy path.

## Suggested fix

Not prescriptive, but the shape that removes the class of bug rather than this instance:

1. **Accept the verdict family, scoped.** `_has_data_ok_verdict` should accept
   `TABLES_OK '<probed table>'` and `TABLE_OK` as well as `DATA_OK` — scoping `TABLES_OK` to the table
   the probe actually asked for, so #115's guarantee is preserved (a `TABLES_OK` naming some *other*
   table must still not count).
2. **Honour the child's exit code** instead of, or in addition to, parsing prose. The child already
   returns 0 on success.
3. **Add a contract test at the seam.** Assert that the argv `probe_live_source` actually builds,
   run against `_verdict.py`'s real emission logic, produces output `_has_data_ok_verdict` accepts.
   A hand-written fixture cannot catch this class; a test that derives the fixture from the real
   caller can.

Please use `Fixes #<this issue>` in the commit message (a bare `(#N)` reference does not close it).

### LINKED ISSUE #153: P1: the no-dialog banner contains its own sentinel tokens, so a healthy refresh classifies as NO_CREDENTIAL
## Summary

The **reassuring** banner printed when there is **no** blocking dialog contains the literal sentinel
tokens `CREDENTIAL_MISSING` and `BLOCKED_BY_DIALOG`. `probe_live_source.py`'s classifier substring-
matches those tokens against the whole stdout blob, so **the detector's own "everything is fine"
message is what fabricates a `NO_CREDENTIAL` verdict.**

Filed separately from #A because it has a different owner and a different fix. #A alone would fail a
healthy source as an unclassified `ERROR`; #B upgrades that into a *confident, wrong, human-directed*
credential stop — telling an operator to go and sign in to a source that just returned data.

## Root cause

`.github/skills/pbip-model-refresh/scripts/_credential_modal.py:333-342` — `print_refresh_banner()`,
which by construction runs **only on the no-dialog path**, emits prose containing:

> ... at the deadline it re-checks and reports **CREDENTIAL_MISSING**, **BLOCKED_BY_DIALOG**, or SLOW_SOURCE.

`scripts/probe_live_source.py:354-427` — `_classify_failure` does a naive lowercase substring test:

```python
if "blocked_by_dialog" in low:
```

`low` is the entire child stdout, so the banner matches itself.

## Reproduction (offline — no Desktop, no network, no credentials)

```
input         : the happy-path banner ONLY (literally says 'No blocking dialog')
_classify_failure -> NO_CREDENTIAL
expected      -> ERROR at worst; NEVER NO_CREDENTIAL
```

**Isolating control** — same captured output with only the banner line deleted:

```
same output, banner line deleted -> _classify_failure=ERROR
```

which confirms the banner is the sole cause of the `NO_CREDENTIAL` label, and that #A alone would
still wrongly fail the run, just as `ERROR`.

## Regression window

Introduced in **97a033e (#151, 2026-08-14)** — "Detect a blocking Desktop dialog before waiting 330s
for it".

## Suggested fix

Reword-only is a patch, not a fix — the landmine (free prose being scanned for control tokens) stays
armed for the next person who writes a helpful message.

1. **Classify structurally.** Restrict sentinel matching to a machine-readable verdict line
   (`^(PREFLIGHT|REFRESH): <TOKEN>`), not an unanchored substring scan of the whole blob. Note the
   codebase already learned this lesson once for `DATA_OK` —
   `test_requires_data_ok_verdict_token` exists precisely to stop a stray `DATA_OK` substring from
   clearing the gate. The same anchoring was never applied to the *failure* tokens.
2. **And** make the banner not contain the tokens it is describing (belt and braces).
3. Add a regression test using the **verbatim captured banner text**, asserting it does not classify
   as `NO_CREDENTIAL`.

Please use `Fixes #<this issue>` in the commit message (a bare `(#N)` reference does not close it).

Guust-Franssens:
## What & why

Two defects in the **live-source credential-gate verdict seam** (parent
`scripts/probe_live_source.py` ↔ child `pbip-model-refresh` skill). Both block the
Monday 2026-08-17 workshop. 1343 tests were green the whole time because the one seam
test hand-wrote a verdict line the production caller **cannot** elicit.

### Defect A — #152 (P0): the gate can never clear
- The probe **always** shells out with `--tables <table>` and **never** `--canaries`,
  so `_verdict._emit_data_verdict` emits `REFRESH: TABLES_OK '<table>'` (exit 0) — but
  `_has_data_ok_verdict` accepted **only** the literal `DATA_OK`, which that argv can
  never produce. No source/credential/config could ever lift the gate.
- `_refresh_and_classify` ignored the child's **exit code** and classified on stdout
  prose alone.

**Fix:** `_has_data_ok_verdict(text, table)` now accepts the whole success family the
child emits — `DATA_OK` (model-level, unconditional) and `TABLE_OK` / `TABLES_OK`
**scoped** to the probed table (a verdict naming some *other* table is a false
certificate — preserves #115). Matching stays **anchored** to a verdict line, never a
substring, so `DATA_OK_FROM_WORKER` still doesn't count. The table name is threaded
from `_refresh_and_classify`, which now requires **both** `returncode == 0` **and** the
scoped success verdict.

### Defect B — #153 (P1): a healthy refresh is labelled `NO_CREDENTIAL`
`_classify_failure` did `if "blocked_by_dialog" in low` over the whole stdout blob, and
the child's own reassuring banners *named* the sentinel tokens in prose, so the banner
matched itself → `NO_CREDENTIAL` + "A HUMAN MUST ACT". There are **two** such banners:

- `print_refresh_banner` — the healthy **no-dialog** path (the reported instance).
- `print_refresh_unknown_banner` — the **minimized-owner UNKNOWN** path (a **second
  instance**, confirmed by #154). Besides naming the tokens, its old `"Credential
  dialog check"` prefix contained the bare word **"credential"**, which is itself a
  `CREDENTIAL_MARKER`.

**Fix (belt-and-braces, both halves):**
1. **Structural** — credential-stop tokens are matched only on a
   `^(REFRESH|PREFLIGHT|PROBE):\s+<TOKEN>` verdict **line**.
2. **Reword both banners** so they carry no control tokens (and not the bare word
   "credential"). `print_refresh_unknown_banner`'s prefix is now `"Blocking-dialog
   check … UNKNOWN"`, so a generic `ERROR` routed through the UNKNOWN path is no longer
   mislabelled. Both banners keep their operational prose (`"No blocking dialog"` /
   `"UNKNOWN (<reason>)"`); the UNKNOWN `reason` arg is preserved untouched.

`CREDENTIAL_MARKERS` stays a **separate, deliberately-unanchored** free-text scan — a
revoked Databricks PAT returns a 403/socket-reset with **no** modal and **no** verdict
line, so those markers are the *only* signal that catches it. Only the verdict *tokens*
got the structural treatment.

## Tests (the part CI was missing)
- **Seam/contract test** derives its fixture from **both real sides**: capture the exact
  argv the probe builds → parse with `refresh_pbip_model`'s **real** `_build_arg_parser`
  → run `_verdict`'s **real** `_emit_data_verdict` on a one-row refresh → assert the
  parent accepts the emitted line. This catches a vocabulary desync across the seam; the
  old hand-written `REFRESH: DATA_OK` fixture could not.
- **Verbatim 2026-08-14 live-Databricks** success (exit 0 + `TABLES_OK 'shipment'`) now
  classifies `DATA_OK` (it returned `NO_CREDENTIAL` at exit 1 before).
- **Negative controls:** non-zero exit is never success; `TABLES_OK` for a *different*
  table doesn't clear the gate (#115).
- **#153 banner regression is PARAMETRIZED over BOTH banner functions**
  (`print_refresh_banner` + `print_refresh_unknown_banner`) rather than hard-coding one
  string, so a third banner that reintroduces a sentinel token in prose is caught the
  moment it is added to the list.
- The unanchored `CREDENTIAL_MARKERS` free-text path still stops a revoked-PAT socket
  reset with no verdict line.
- The anchoring guard `test_requires_data_ok_verdict_token` stays green.

## Relationship to #154 (NOT fixed here — separate owner)
#154 (minimized Desktop destroys credential-dialog evidence; `UNKNOWN` isn't latched, so
the verdict degrades to `SLOW_SOURCE`) is a **separate** fix in
`join_with_credential_poll`, which this PR **leaves untouched**. This PR only reworded
the banner *prose* on that path; the structured `unknown_reason`/`reason` and the poll
loop are unchanged, so #154's fix surface is intact. As a courtesy to that fix,
`CREDENTIAL_STOP_VERDICT_RE` carries a one-line forward-compat comment naming the exact
parse point to extend if a latched-`UNKNOWN` outcome ever earns its own verdict token
(e.g. `CREDENTIAL_UNKNOWN`). No behavioural change for #154 is included.

## Validation
Bare `pytest` (the `testpaths` invocation) is green — **1356 passed, 7 skipped**.

> **Pre-existing test-harness note (not caused by this change):** the explicit two-path
> form `pytest tests/ .github/skills/pbip-model-refresh/tests/` hits an
> `importmode=prepend` **name clash** between the repo's forwarding shims in `scripts/`
> (`refresh_pbip_model`, `probe_desktop_query`) and the skill's real modules — 3
> collection `ImportError`s. **Reproduced on clean `master`** (`97a033e`) with the same
> command, so it predates this PR. Validated instead by (a) bare `pytest` and (b) running
> the two suites **separately**: `tests/` → 1202 passed / 4 skipped, skill tests → 143
> passed / 3 skipped.

## Files changed
| File | Why |
|---|---|
| `scripts/probe_live_source.py` | verdict-family + scoped acceptance, exit-code honouring, structural credential-stop classification, two module regexes (+ #154 forward-compat comment) |
| `.github/skills/pbip-model-refresh/scripts/_credential_modal.py` | reword **both** refresh banners to carry no control tokens |
| `tests/test_probe_live_source_verdict.py` | seam/contract test + regressions; banner regression parametrized over both banners |
| `.github/skills/pbip-model-refresh/tests/test_credential_modal_detection.py` | one assertion updated for the reworded UNKNOWN banner prefix |

Fixes #152
Fixes #153

### DIFF
--- a/.github/skills/pbip-model-refresh/scripts/_credential_modal.py
+++ b/.github/skills/pbip-model-refresh/scripts/_credential_modal.py
@@ -331,25 +331,39 @@ def source_hint_from_model(model_dir: Path | None) -> str | None:
 
 
 def print_refresh_banner(pid: int, timeout_sec: int, grace_sec: int | float) -> None:
-    """Print the no-dialog, self-bounded refresh warning before the wait starts."""
+    """Print the no-dialog, self-bounded refresh warning before the wait starts.
+
+    Deliberately does NOT name the verdict tokens it may later print. This banner is emitted on the
+    HEALTHY no-dialog path and is captured verbatim by ``probe_live_source``'s classifier; naming
+    ``CREDENTIAL_MISSING`` / ``BLOCKED_BY_DIALOG`` here (or even the bare word "credential") made the
+    reassuring message classify a successful refresh as ``NO_CREDENTIAL`` (issue #153). The classifier
+    now matches verdict LINES structurally, but keeping control tokens out of free prose is the
+    belt-and-braces half of that fix - so a future helpful edit cannot re-arm the landmine.
+    """
     total = timeout_sec + grace_sec
     print(
         f"No blocking dialog on PID {pid}. Refreshing, bounded at {timeout_sec}s XMLA + "
         f"{grace_sec}s grace ({total}s total); a long wait here is expected for a serverless cold start. "
-        "DO NOT kill this process - at the deadline it re-checks and reports CREDENTIAL_MISSING, "
-        "BLOCKED_BY_DIALOG, or SLOW_SOURCE. Killing it early yields NO verdict.",
+        "DO NOT kill this process - at the deadline it re-checks and prints one final machine-readable "
+        "verdict line. Killing it early yields NO verdict.",
         flush=True,
     )
 
 
 def print_refresh_unknown_banner(pid: int, timeout_sec: int, grace_sec: int | float, reason: str) -> None:
-    """Print the bounded-refresh warning when the t=0 modal check is indeterminate."""
+    """Print the bounded-refresh warning when the t=0 modal check is indeterminate.
+
+    Same rule as :func:`print_refresh_banner`: no control tokens - and not the bare word "credential" -
+    in prose the classifier scans (issue #153). The prefix reads "Blocking-dialog check ... UNKNOWN"
+    rather than the old "Credential dialog check", so a non-``DATA_OK`` refresh through this path is not
+    mislabelled ``NO_CREDENTIAL`` by ``CREDENTIAL_MARKERS``' free-text "credential" marker.
+    """
     total = timeout_sec + grace_sec
     print(
-        f"Credential dialog check on PID {pid} is UNKNOWN ({reason}). Refreshing, bounded at "
+        f"Blocking-dialog check on PID {pid} is UNKNOWN ({reason}). Refreshing, bounded at "
         f"{timeout_sec}s XMLA + {grace_sec}s grace ({total}s total); a minimized owner can hide owned "
         "dialogs from enumeration. DO NOT kill this process - at the deadline it re-checks with the "
-        "dialog arbiter and reports CREDENTIAL_MISSING, BLOCKED_BY_DIALOG, or SLOW_SOURCE. Killing it early "
+        "dialog arbiter and prints one final machine-readable verdict line. Killing it early "
         "yields NO verdict.",
         flush=True,
     )
--- a/.github/skills/pbip-model-refresh/scripts/_credential_modal.py
+++ b/.github/skills/pbip-model-refresh/scripts/_credential_modal.py
@@ -279,7 +279,7 @@ def inspect_credential_modal(
     ]
     if minimized:
         return CredentialDetection(
-            unknown_reason="Power BI Desktop owner window is minimized; owned credential dialogs are hidden",
+            unknown_reason="Power BI Desktop owner window is minimized; owned modal dialogs are hidden from enumeration",
             windows=windows,
         )
     return CredentialDetection(windows=windows)
@@ -333,12 +333,14 @@ def source_hint_from_model(model_dir: Path | None) -> str | None:
 def print_refresh_banner(pid: int, timeout_sec: int, grace_sec: int | float) -> None:
     """Print the no-dialog, self-bounded refresh warning before the wait starts.
 
-    Deliberately does NOT name the verdict tokens it may later print. This banner is emitted on the
-    HEALTHY no-dialog path and is captured verbatim by ``probe_live_source``'s classifier; naming
-    ``CREDENTIAL_MISSING`` / ``BLOCKED_BY_DIALOG`` here (or even the bare word "credential") made the
-    reassuring message classify a successful refresh as ``NO_CREDENTIAL`` (issue #153). The classifier
-    now matches verdict LINES structurally, but keeping control tokens out of free prose is the
-    belt-and-braces half of that fix - so a future helpful edit cannot re-arm the landmine.
+    Deliberately does NOT name the verdict tokens it may later print, and interpolates NO caller-supplied
+    text - so unlike :func:`print_refresh_unknown_banner`, this banner's marker-free guarantee is total
+    and self-contained. This banner is emitted on the HEALTHY no-dialog path and is captured verbatim by
+    ``probe_live_source``'s classifier; naming ``CREDENTIAL_MISSING`` / ``BLOCKED_BY_DIALOG`` here (or
+    even the bare word "credential") made the reassuring message classify a successful refresh as
+    ``NO_CREDENTIAL`` (issue #153). The classifier now matches verdict LINES structurally, but keeping
+    control tokens out of free prose is the belt-and-braces half of that fix - so a future helpful edit
+    cannot re-arm the landmine.
     """
     total = timeout_sec + grace_sec
     print(
@@ -353,10 +355,13 @@ def print_refresh_banner(pid: int, timeout_sec: int, grace_sec: int | float) -> 
 def print_refresh_unknown_banner(pid: int, timeout_sec: int, grace_sec: int | float, reason: str) -> None:
     """Print the bounded-refresh warning when the t=0 modal check is indeterminate.
 
-    Same rule as :func:`print_refresh_banner`: no control tokens - and not the bare word "credential" -
-    in prose the classifier scans (issue #153). The prefix reads "Blocking-dialog check ... UNKNOWN"
-    rather than the old "Credential dialog check", so a non-``DATA_OK`` refresh through this path is not
-    mislabelled ``NO_CREDENTIAL`` by ``CREDENTIAL_MARKERS``' free-text "credential" marker.
+    This banner's OWN static prose names no control tokens and not the bare word "credential" - but it
+    interpolates ``reason``, a caller-supplied string from ``inspect_credential_modal`` that the
+    classifier scans along with everything else. So the marker-free guarantee for the UNKNOWN path is
+    owned by the DETECTOR, not by this banner alone: the minimized-owner ``unknown_reason`` was reworded
+    off the word "credential" (a ``CREDENTIAL_MARKER``) for exactly this reason (issue #153) - otherwise
+    a slow/timeout refresh here was mislabelled ``NO_CREDENTIAL``. The prefix reads "Blocking-dialog
+    check ... UNKNOWN" rather than the old "Credential dialog check" as the belt-and-braces static half.
     """
     total = timeout_sec + grace_sec
     print(
--- a/.github/skills/pbip-model-refresh/scripts/_credential_modal.py
+++ b/.github/skills/pbip-model-refresh/scripts/_credential_modal.py
@@ -279,7 +279,9 @@ def inspect_credential_modal(
     ]
     if minimized:
         return CredentialDetection(
-            unknown_reason="Power BI Desktop owner window is minimized; owned modal dialogs are hidden from enumeration",
+            unknown_reason=(
+                "Power BI Desktop owner window is minimized; owned modal dialogs are hidden from enumeration"
+            ),
             windows=windows,
         )
     return CredentialDetection(windows=windows)
--- a/.github/skills/pbip-model-refresh/scripts/probe_desktop_credential.ps1
+++ b/.github/skills/pbip-model-refresh/scripts/probe_desktop_credential.ps1
@@ -171,7 +171,7 @@ if ($blocker) {
 }
 foreach ($w in $windows) {
   if ($w.Minimized -and $w.ClassName.StartsWith('WindowsForms10.Window.8')) {
-    Write-Output "Power BI Desktop owner window is minimized; owned credential dialogs are hidden"
+    Write-Output "Power BI Desktop owner window is minimized; owned modal dialogs are hidden from enumeration"
     Write-Output "VERDICT: UNKNOWN"
     exit 3
   }
--- a/.github/skills/pbip-model-refresh/tests/test_credential_modal_detection.py
+++ b/.github/skills/pbip-model-refresh/tests/test_credential_modal_detection.py
@@ -286,7 +286,7 @@ def test_unknown_refresh_banner_does_not_claim_no_dialog(monkeypatch, parked, ca
         refresh(port=1234, tables=["Orders"], timeout_sec=0.1, desktop_pid=111)
 
     out = capsys.readouterr().out
-    assert "Credential dialog check on PID 111 is UNKNOWN" in out
+    assert "Blocking-dialog check on PID 111 is UNKNOWN" in out
     assert "No blocking dialog on PID 111" not in out
     parked.set()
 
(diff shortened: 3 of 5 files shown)


Guust-Franssens:
## Follow-up (commit `45a79a6`): fixed the DYNAMIC half of #153 the review caught

You were right — #153 was only half fixed. The first commit reworded the banners' **static** prose but `print_refresh_unknown_banner` interpolates a caller-supplied `{reason}`, and the production reason for a minimized owner literally contained the word **"credential"** — so `_classify_failure` still returned `NO_CREDENTIAL`. Verified on the **real detector output** (no hand-written fixtures), both before and after:

| detector `unknown_reason` (harvested by driving the real branch) | before | after |
|---|---|---|
| `window enumeration failed: …` (L269) | `markers=[]` → **ERROR** | `markers=[]` → **ERROR** |
| `… owned credential dialogs are hidden` (L282) | `markers=['credential']` → **NO_CREDENTIAL** 🐞 | *(reworded)* `markers=[]` → **ERROR** ✅ |

The two UNKNOWN reasons classified **differently**, purely by wording — the tell that the fix belongs at the reason source, not the banner.

### What changed
1. **`_credential_modal.py` `inspect_credential_modal`** — reword the minimized-owner `unknown_reason` off the word "credential": `"… owned modal dialogs are hidden from enumeration"`. **Truthiness and semantics unchanged** (still a non-empty description of the same condition), so **#154's latching fix is unaffected** — I did **not** touch `join_with_credential_poll` (`git diff master` confirms it's not in the diff), and the value still flows through `if state.unknown_reason:` exactly as before.
2. **`probe_desktop_credential.ps1`** — same string in the human-run PowerShell **mirror** of the detector. It is **not** in the automated classifier path (`refresh_pbip_model` only prints its path in an *operator hint* string; `probe_live_source` shells out only to `refresh_pbip_model.py`/`probe_desktop_query.py`), so this is a consistency + defensive change, **not** a second live bug — called out so the two detectors don't drift.
3. **Both banner docstrings softened** to claim only what each *enforces*: `print_refresh_banner` interpolates nothing caller-supplied (guarantee is total); `print_refresh_unknown_banner`'s marker-free guarantee is owned by the **detector**, since it interpolates a reason it doesn't control.

### Test is now structural, not enumerated
Replaced the hand-written `parametrize` with `test_no_refresh_banner_and_detector_reason_classifies_as_no_credential`, which **discovers both inputs** so it catches a new banner **and** a new reason with no list to edit:
- **banners** — introspect `_credential_modal` for `print_refresh_*banner`;
- **reasons** — **drive the real detector** down each UNKNOWN branch and harvest the actual `unknown_reason` strings (the old test hand-wrote a reason the detector never emits — the exact #152/#153 blind-spot class, one layer down), plus a source-level backstop that trips if a **third** `unknown_reason` branch is added without a driver.

**Proven to FAIL against pre-fix code and PASS after** (a real regression test, not green-both-ways):
```
# against the unfixed reason string:
AssertionError: banner(s) fabricated a credential stop from their own prose:
  print_refresh_unknown_banner('… owned credential dialogs are hidden',): 'Blocking-dialog check … NO verdict.\n'
FAILED … 1 failed
# after the fix:
1 passed
```

### Judgement call — recorded, **not** fixed (both reviewers flagged it)
`_has_data_ok_verdict`'s `re.findall(r"'([^']*)'", …)` mis-parses a table name containing a literal apostrophe (`O'Brien Sales`). **Declining**, with reasoning documented inline: it **fails safe** (only ever keeps the gate *shut*, never lifts it wrongly), the false-lift is **unreachable** in the single-table probe flow (the probe asks for exactly the table it names), the ambiguity originates in `_verdict`'s **unescaped** emission, and the #115 scoping logic is verified-correct / on your do-not-touch list — so I left the matcher untouched rather than risk regressing it for a fail-safe, unreachable edge. Happy to fix in a follow-up if you'd rather.

### Validation
Bare `pytest` → **1355 passed, 7 skipped** (the two-case parametrized banner test collapsed into one structural test, net −1 instance). Detection + credential-gate suites: 90 passed. `ruff` clean.

Guust-Franssens:
## Independent verification of `45a79a6` — ✅ all checks pass

Verified with a harness that harvests the reason strings by a **different mechanism** than this PR's
own test (AST-parsing `_credential_modal.py` for `unknown_reason` assignments, rather than driving
the detector), so this is a genuine second opinion and not a re-run of the same logic.

```
harvested 2 unknown_reason string(s) from _credential_modal.py:
  - 'Power BI Desktop owner window is minimized; owned modal dialogs are hidden from enumeration'
  - 'window enumeration failed: <exc>'
discovered 2 banner function(s): ['print_refresh_banner', 'print_refresh_unknown_banner']

PASS  print_refresh_banner             -> ERROR   markers=[]  reason='<no reason argument>'
PASS  print_refresh_unknown_banner     -> ERROR   markers=[]  reason='...owned modal dialogs are hidden...'
PASS  print_refresh_unknown_banner     -> ERROR   markers=[]  reason='window enumeration failed: <exc>'

PASS  unknown_reason stays truthy (#154 depends on it)  x2
PASS  #152 live happy stdout lifts gate            -> True
PASS  #115 same stdout, wrong table stays shut      -> False
PASS  revoked-PAT 403 free text -> ACCESS_DENIED (gate-holding, specific)
```

### The fails-before/passes-after claim is confirmed by controlled experiment

Not taken on trust — I ran the **new** test against the **old** production code:

```
$ git checkout c479fb4 -- .github/skills/pbip-model-refresh/scripts/_credential_modal.py
$ pytest tests/test_probe_live_source_verdict.py -q
FAILED ...::test_no_refresh_banner_and_detector_reason_classifies_as_no_credential
1 failed, 9 passed

E  AssertionError: banner(s) fabricated a credential stop from their own prose:
E  print_refresh_unknown_banner('Power BI Desktop owner window is minimized; owned credential
E  dialogs are hidden',): 'Blocking-dialog check on PID 63824 is UNKNOWN (...)'

$ git checkout HEAD -- .github/skills/pbip-model-refresh/scripts/_credential_modal.py
$ pytest tests/test_probe_live_source_verdict.py -q
10 passed
```

It fails on the **exact production string**, which is the property that was missing before. This is a
real regression test, not a green-both-ways one.

### One correction to my own earlier report

My first run flagged the revoked-PAT case as a failure. **That was my harness's bug, not the code's**
— `"403"` is in `ACCESS_DENIED_MARKERS`, which is checked *before* `CREDENTIAL_MARKERS`, and
`ACCESS_DENIED` is the more precise verdict for a 403. Both hold the gate. Likewise my first
enum-failure result was poisoned by my own placeholder text (`<OSError: access denied>`) rather than
by anything in the reason. No defect in either case.

### Residual, non-blocking — recorded so it is not rediscovered

The enum-failure reason is an f-string: `f"window enumeration failed: {exc}"`. That `{exc}` is
**uncontrolled OS text** flowing into prose the parent classifier scans — the same injection class as
#153, one level further out. Measured against realistic OS error strings:

| interpolated `{exc}` | verdict |
|---|---|
| `[WinError 5] Access is denied.` | clean (`ERROR`) |
| `[WinError 1400] Invalid window handle` | clean (`ERROR`) |
| `OSError: [Errno 13] Permission denied` | mislabels as `ACCESS_DENIED` |
| `access denied enumerating top-level windows` | mislabels as `ACCESS_DENIED` |

The canonical Win32 message survives only by luck — the marker is the contiguous `"access denied"`
and Windows says "Access **is** denied". This fails in the **gate-holding** direction (a wrong
*reason*, never a false clear) and the POSIX-style variants are unlikely on this Win32 path, so it is
**not a blocker**. The durable fix would be to stop concatenating uncontrolled text into
classifier-scanned prose at all (e.g. emit the reason on its own non-scanned line).

### Verdict

Both blocking findings resolved. `Fixes #152` / `Fixes #153` present. #154's `unknown_reason`
semantics untouched. The apostrophe decline is well-reasoned and I accept it.

**Ready to merge.**

PR #162, opened 2026-08-15
Stop indeterminate Desktop states degrading to a wrong SLOW_SOURCE verdict (Fixes #154, Fixes #158)

### LINKED ISSUE #154: P1: minimizing Desktop permanently destroys credential-dialog evidence - UNKNOWN is not latched, verdict degrades to SLOW_SOURCE

### LINKED ISSUE #158: P1: a dead or window-less Power BI Desktop reports as HEALTHY - zero enumerated windows falls through to 'none', degrading to SLOW_SOURCE
## Summary

`inspect_credential_modal` (`.github/skills/pbip-model-refresh/scripts/_credential_modal.py`) has **no
zero-window guard**. When window enumeration returns an **empty** list, every branch falls through and
it returns `CredentialDetection(windows=())` — all fields `None`, which the callers read as
**`none` = "healthy Desktop, no dialog, keep waiting"**.

But zero windows is not evidence of health. It is evidence that **the Desktop process is gone,
crashed, or has not created its window yet**. A live, working Power BI Desktop always has a main
window. So the single strongest signal that something is wrong is currently reported as the
all-clear.

Consequence: the same wrong-and-confident verdict as #154 — the poll loop keeps waiting the full
330 s and then reports **`SLOW_SOURCE`**, blaming the warehouse for being slow when the truth is
*"Power BI Desktop died"*.

## Measured, not theorised

Observed 2026-08-15 00:50 local against two real Power BI Desktop PIDs that had just exited (they
belonged to a long-running migration pipeline that cycles Desktop instances between workbooks):

```
===== pid 8480 =====
  windows enumerated : 0
  modal              : None
  blocking_dialog    : None
  unknown_reason     : None
  => VERDICT         : none (no dialog detected)
===== pid 63308 =====
  windows enumerated : 0
  ...
  => VERDICT         : none (no dialog detected)
```

Both processes were confirmed **`GONE`** by `Get-Process -Id <pid>` at the same moment. The detector
reported them as healthy.

## Why the existing UNKNOWN guard does not catch it

The minimized guard requires a window to *exist* and be iconic:

```python
minimized = [
    window for window in windows
    if window.minimized and window.class_name.startswith(DESKTOP_MAIN_CLASS_PREFIX)
]
if minimized:
    return CredentialDetection(unknown_reason="...owner window is minimized...", windows=windows)
return CredentialDetection(windows=windows)   # <-- empty list lands here
```

With `windows = ()` the comprehension is empty, so the `UNKNOWN` degrade never fires and the function
falls through to the healthy return. The `Win32EnumerationError` path is also not hit — enumeration
**succeeded**, it just found nothing.

## Suggested fix

Distinguish two genuinely different cases rather than collapsing both into `none`:

1. **Process is not alive** — definitive. Report a distinct terminal state (Desktop exited/crashed).
   This must never degrade to `SLOW_SOURCE`; the warehouse is not implicated at all.
2. **Process is alive but enumerated zero windows** — indeterminate. Degrade to **`UNKNOWN`** with a
   reason, and let it latch (see #154).

⚠️ Do **not** simply treat "zero windows" as terminal-crashed: a Desktop that is still **starting up**
can legitimately enumerate zero windows for a short period. That is exactly why the liveness check
above is what separates the definitive case from the transient one.

⚠️ Any new `unknown_reason` text must be free of every token in `CREDENTIAL_MARKERS`
(`credential`, `sign in`, `signed in`, `authentication`, `unauthorized`, `access token`, `login`,
`oauth`, `10054`, `forcibly closed`, `unrecognizable response`) — the parent classifier scans banner
prose, and a stray marker misclassifies the failure. That was #153.

## Related

- #154 — `UNKNOWN` is not latched; same "indeterminate reported as healthy, degrades to
  `SLOW_SOURCE`" class. A fix for this issue should latch through the same mechanism.
- #153 — banner prose poisoning the classifier (constrains the wording of any new reason string).

Please use `Fixes #<this issue>` in the commit message — a bare `(#N)` reference does not close it.

Guust-Franssens:
Fixes #154

Fixes #158

Two P1 bugs in the same credential-modal detector subsystem, **same defect class** — an indeterminate

Desktop state reported as healthy (`none`) that degrades to a wrong `SLOW_SOURCE`/timeout at the

deadline — fixed together because #158 wants the **same latch mechanism** #154 introduces and lives in

the same function's immediate neighbourhood.

## #154 — minimize→restore permanently destroys credential-dialog evidence

P1: minimizing Desktop while a credential probe waits **permanently destroys** the credential-dialog

evidence. The detector correctly degrades to `UNKNOWN` when the owner window is iconic (a minimized

owner hides its owned modal dialogs from enumeration — empirically verified against a real Snowflake

702×257 dialog), but `join_with_credential_poll` **ignored `unknown_reason`**. On `minimize → restore`

the dialog provably never returns, the detector reports `none`, that `none` overwrote the earlier

`UNKNOWN`, and at the deadline the run returned `False` → the caller reported **`SLOW_SOURCE`**. Net

effect measured in the issue: a correct **74 s `NO_CREDENTIAL`** became a **wrong 330 s timeout**

blaming the warehouse when nobody ever signed in — exactly the wrong-and-confident verdict class this

probe exists to eliminate.

## What changed and why

**Core fix — latch the observation (`_credential_modal.py`)**

- `join_with_credential_poll` now **latches** the first `unknown_reason` (owner minimized) *or* blocking

  dialog it ever sees during a run into a sticky `latched_unknown`. A later `none` can no longer clear

  it. At the deadline, a run that ever went `UNKNOWN` raises the new **`CredentialUnknownError(pid,

  reason)`** instead of returning `False`.

- New `print_indeterminate_state_notice(pid, reason)` logs the iconic transition loudly. It is

  **marker-free** (contains no token in `CREDENTIAL_MARKERS`), per the #153 constraint. (It was renamed

  from `print_iconic_transition_notice` when #158 generalised it to cover the zero-window cause too.)

- The window is **deliberately NOT auto-restored** (see "Deliberately not done").

**Wiring (`refresh_pbip_model.py`)**

- New `_emit_credential_unknown(pid, reason)` prints `REFRESH: CREDENTIAL_UNKNOWN pid=…; {reason}` plus

  marker-free guidance pointing at the `CREDENTIAL_PROBE` arbiter.

- The `except` block in `_refresh_and_save` catches `CredentialUnknownError` and **returns exit 3** — the

  mandatory distinct UNKNOWN state, never collapsed into blocked's exit 1 or a healthy path.

**Parent classifier (`probe_live_source.py`)**

- `CREDENTIAL_STOP_VERDICT_RE` now recognises `CREDENTIAL_UNKNOWN` alongside `CREDENTIAL_MISSING` /

  `BLOCKED_BY_DIALOG`, so the child's structural verdict maps to **`NO_CREDENTIAL` at step 5** (the

  structural check, *before* free-text marker scanning). The emitted reason stays marker-free so it can

  never be misclassified by the step-6 free-text scan.

## Decision: latch-and-keep-waiting, **NOT terminal**

The brief invited me to consider making a mid-wait `UNKNOWN` **terminal** (short-circuit the wait). I

chose **latch-but-keep-waiting**, and here is the evidence for that call:

- **A terminal short-circuit has a false-positive path; latch-and-wait does not.** The only run that

  ever *surfaces* the latched verdict is one that reaches the deadline anyway — because a worker that

  **finishes** clears the wait via `return True` **regardless of the latch**. So a healthy refresh that

  merely happened to be minimized for a moment is **never** overridden: it completes, returns `True`,

  and the latch is discarded unused. A terminal short-circuit would instead **end the wait early** the

  first time enumeration hiccuped — turning a transient blip during a healthy-but-slow refresh into a

  premature `UNKNOWN`. That is a new failure mode this fix must not introduce.

- **The cost of waiting is nil in the failure case.** When the run really is stuck behind a hidden

  dialog, the verdict is now correct either way; latch-and-wait simply reports it at the deadline

  instead of ~250 s earlier. Since a stuck run produces no useful signal in that window regardless, the

  only thing "terminal" buys is speed on a path that was already going to fail — at the price of a

  false-positive path on the healthy path. Not worth it.

If a future measurement shows the ~250 s matters operationally, "terminal" can be added **on top of**

the latch safely by also requiring the main window to still be iconic at the deadline. I did not do that

here because it is unverifiable without launching Desktop, and latch-and-wait is the strictly safer

default.

## Fails-before / passes-after evidence (controlled experiment)

The regression test drives the **real** detector and the **real** `join_with_credential_poll` loop with

an injected fake `WindowEnumerator` replaying the exact measured sequence: `dialog visible` → `owner

minimized (dialog gone, main window iconic)` → `owner restored (dialog still gone, no candidates)` →

deadline.

1. Backed up the fixed `_credential_modal.py`, reverted **only** the latch logic (kept the tests), ran

   the three poll-loop tests:

   ```

   test_poll_loop_latches_minimize_restore_and_raises_unknown  FAILED

       Failed: DID NOT RAISE <class '_credential_modal.CredentialUnknownError'>

   test_poll_loop_without_unknown_returns_false                 PASSED

   test_poll_loop_raises_blocked_on_visible_dialog              PASSED

   ```

   The latch test fails exactly as the bug predicts (loop returns `False` instead of raising), while the

   two guard tests still pass — confirming the reproduction is specific to the defect and introduces no

   false-positive.

2. Restored the fix (hash-verified identical to pre-experiment), re-ran: **all 3 PASS**.

**Blind-spot guard (#152/#153):** every asserted string is **harvested from production**, never

hand-written. `harvested_minimized_reason()` / `_harvest_minimized_reason()` drive the real

`inspect_credential_modal` to obtain the actual reason string, and `_ScriptedEnumerator` feeds real

window-lists to the real detector inside the real loop. No test asserts a string production cannot emit.

## Validation

- **ruff** `format` + `check` — clean on all 5 files (repo lint select is `["E4","E7","E9","F"]`; I also

  hand-verified no added line exceeds **120** chars, since that select excludes E501 and pylint C0301

  would otherwise red-lit CI).

- **pylint** per-directory, exactly as `.github/workflows/checks.yml` invokes it (run via the repo venv

  interpreter so `lxml` resolves — the global interpreter lacks it and spuriously reports `E0401` on the

  untouched `trace_customer_text.py`):

  - `pylint scripts` → **10.00/10** (`probe_live_source.py` contributes zero messages)

  - `pylint .github/skills/pbip-model-refresh/scripts` → **10.00/10**

- **Full suites** `tests/` + `.github/skills/pbip-model-refresh/tests/`: **1362 passed, 7 skipped**

  (1347 pre-#154 → +15 total: #154 added 6, #158 added 9 (6 child, 3 parent); the two target files went

  29 → 44 tests).

## Deliberately NOT done (scope discipline)

- **Did not auto-restore the minimized window.** The issue notes the probe owns the Desktop, but

  auto-restoring is a side effect the brief explicitly did not want smuggled into this PR. The fix only

  *observes and latches*; it never manipulates the window.

- **Did not align the PowerShell arbiter** (`probe_desktop_credential.ps1`). Its text tier scans every

  window unfiltered by design; parity with the Python side is intentionally guaranteed by the shared

  `credential_modal_signature.regex`, not by code duplication. Out of scope.

- **Did not touch `probe_desktop_query.py`'s separate inline poll loop.** #154 is about

  `join_with_credential_poll`; the exit-3-for-UNKNOWN contract in `probe_desktop_query.py` is consumed,

  not modified.

- **Did not collapse the four exit states.** `UNKNOWN` remains a distinct third state (exit 3), verified

  by `test_refresh_and_save_wires_credential_unknown_to_exit_3`.

- **Left the pre-existing `t=0` pre-check load-bearing behaviour intact** — it prevents a false clear

  when the worker finishes before the loop's first check.

## Unverified (honest gap)

The empirical claim that a restored owner's dialog *never* re-enters enumeration is taken from the

issue's measurement; I did **not** re-launch Power BI Desktop to re-measure it (a live pipeline owns

Desktop instances on this machine — the durable rule is never to launch Desktop or `Stop-Process` any

PID I did not personally launch, and the brief forbids launching Desktop). The fix is

correct **given** that measured premise, and the design is safe **even if** the premise were weaker —

because a worker that finishes still clears the wait via `return True`, so a dialog that *did* come back

and let the refresh complete would never hit the latched path.

---

## #158 — a dead / window-less Desktop was also reported healthy

Filed mid-work and folded in here rather than a conflicting second PR: it is the **same defect class**

(an indeterminate state reported as healthy, degrading to a wrong `SLOW_SOURCE` at the deadline), wants

the **same latch mechanism**, and lives in the **same function**. `inspect_credential_modal` had **no

zero-window guard**: when enumeration returned an **empty** list, every branch fell through to

`return CredentialDetection(windows=windows)` — all fields `None`, which callers read as `none` =

"healthy Desktop, keep waiting". **Measured 2026-08-15:** two Desktop PIDs that had *just exited*

enumerated **0 windows** and the detector called them healthy; both were confirmed `GONE` by

`Get-Process` at the same moment.

The #154 UNKNOWN guard misses it because the minimized degrade needs a window to **exist** and be

iconic — with `windows = ()` that comprehension is empty and it falls straight through; the

`Win32EnumerationError` path is not hit either, because enumeration **succeeded**, it just found nothing.

### Split by process liveness — the two cases need opposite handling

| Case | Meaning | Handling |

|---|---|---|

| **Process ALIVE, zero windows** | starting up or wedged — indeterminate | `unknown_reason` → **latched** through the *same* #154 machinery → `CredentialUnknownError` → exit 3 → `NO_CREDENTIAL` |

| **Process GONE, zero windows** | Desktop exited/crashed — definitive | `process_gone` → **`DesktopGoneError`, raised immediately (terminal)** → exit 2 → `ERROR` |

⚠️ Not collapsed: treating "zero windows" as *terminal-crashed* would turn a **still-starting** Desktop

(which legitimately enumerates zero windows briefly) into a false failure. The **liveness check is

exactly what separates** the definitive case from the transient one — an alive-but-window-less startup

reads as `UNKNOWN` (latched, keeps waiting), never as `gone`.

### Decision: `process_gone` is **terminal**, `unknown_reason` stays latch-and-wait

Justified by the same false-positive analysis as #154, but the *opposite* conclusion, and the liveness

check is why: the only false-positive a terminal short-circuit could have had (ending a healthy startup

early) has **already been removed** — a booting Desktop reads as `unknown_reason` and is latched, not

raised. A **confirmed-dead** PID cannot recover and the data source was never contacted, so running out

the remaining ~250 s buys nothing. So `process_gone` raises immediately, while the alive/indeterminate

case keeps #154's latch-and-wait (a startup that grows a window and finishes still clears the wait via

`return True`, never overridden).

### What changed

- **`_credential_modal.py`** — new `DesktopGoneError`; the zero-window liveness split **reuses the

  existing `_lock._process_alive`** (Win32 `OpenProcess`/`GetExitCodeProcess`) rather than a second

  ctypes copy — its *assume-alive-on-any-ambiguity* bias is exactly what #158 needs, routing an

  uncertain Desktop to UNKNOWN-latched and never to a false crash (see the review-feedback note below); a

  `process_gone` field on `CredentialDetection`; the **zero-window guard** in `inspect_credential_modal`

  (kept *inline* so the `_detector_unknown_reasons` backstop still counts it). `join_with_credential_poll`

  raises `DesktopGoneError` on `process_gone` (mid-loop **and** at the deadline); `_raise_if_blocked`

  covers it at the load-bearing **t=0 pre-check**.

- **`refresh_pbip_model.py`** — `_emit_desktop_gone(pid, reason)` prints `REFRESH: DESKTOP_GONE pid=…`

  plus marker-free guidance; `_refresh_and_save` returns **exit 2** (the ERROR family) — distinct from

  UNKNOWN's exit 3 and blocked's exit 1.

- **`probe_live_source.py`** — `DESKTOP_GONE_VERDICT_RE` + `_has_desktop_gone_verdict` map the structural

  verdict to **`ERROR`** (a LOCAL failure — the source was never contacted — **never** `UNREACHABLE` or

  `NO_CREDENTIAL`), checked first so it can't be shadowed. Both new detector strings are **marker-free**

  (#153), *verified by harvesting them from the module and testing against the real `CREDENTIAL_MARKERS`

  + `BAD_TABLE_MARKERS` + `ACCESS_DENIED_MARKERS` tuples* — not by eyeballing.

### Review feedback (post-#158): reuse the existing liveness primitive, don't reimplement it

Two review findings on the #158 liveness code, both correct and now folded in:

1. **`worker.is_alive()` is the query *thread*, not the Desktop *process*.** Confirmed — every

   `is_alive()` in this package refers to the worker thread running the query. This PR's liveness check is

   a **separate** `process_is_alive(pid)` call on the Desktop PID, so this finding did **not** bite the

   implementation; but it is the right diagnosis of *why* a dead Desktop could read as healthy — there was

   no process-liveness anywhere in the detection path before #158.

2. **Don't reimplement liveness — the primitive already exists.** My first cut added a parallel ctypes

   `_process_is_alive` in `_credential_modal.py`. `_lock.py` already ships `_process_alive` /

   `_process_alive_windows` (`OpenProcess`/`GetExitCodeProcess`, deliberately **not** `os.kill(pid, 0)`,

   which on Windows would *terminate* the target). I deleted my copy; the detector now imports

   `_lock._process_alive` as the default liveness check. That primitive is **strictly better** than mine

   was — it adds the `GetExitCodeProcess`/`STILL_ACTIVE` dead-detection and a `pid <= 0` guard my version

   lacked — and its *assume-alive-on-any-ambiguity* bias is precisely the safe direction here (uncertain →

   UNKNOWN-latched, never a false crash), so it needed no behavioural change. `_lock.py` is

   **byte-identical** (import only, no edit) and its stale-lock reclaim tests still pass. Net effect:

   **−18 lines**, one liveness implementation instead of two, and no test churn (all tests inject their

   own liveness fake, so the full suite stays **1362 passed / 7 skipped** and pylint stays **10.00/10**).

### #158 fails-before / passes-after (controlled experiment)

Backed up the fixed `_credential_modal.py`, reverted **only** the detector's zero-window guard (the exact

#158 bug locus), kept the tests:

```

# FAILS-BEFORE (guard reverted, sha F4806D8…):

test_poll_loop_raises_desktop_gone_when_process_dead          FAILED  Failed: DID NOT RAISE DesktopGoneError

test_zero_windows_dead_reports_process_gone_not_no_dialog     FAILED  assert None is not None  (…process_gone=None)

```

The primary test fails exactly as the bug predicts — the loop **returns `False`** (the degraded

`SLOW_SOURCE`/timeout path) instead of raising the terminal `DesktopGoneError`, and the detector reports

the healthy-`none` state for a dead process. Restored the fix (hash-verified identical to

pre-experiment, sha `3BCC98D…`), re-ran: **both PASS**, and `DesktopGoneError` is raised on the **first

poll** (`enumerator.calls == 1` — terminal, no waiting).

Nine #158 tests total, all driving the **real** detector / poll loop through an injected **empty**

enumerator + injected liveness, with every asserted string **harvested from production**:

`test_zero_windows_alive_reports_unknown_not_no_dialog`,

`test_zero_windows_dead_reports_process_gone_not_no_dialog`,

`test_direct_refresh_raises_desktop_gone_fast_at_t0`,

`test_poll_loop_raises_desktop_gone_when_process_dead`,

`test_poll_loop_latches_unknown_when_process_alive_zero_windows`,

`test_refresh_and_save_wires_desktop_gone_to_exit_2` (child); and

`test_desktop_gone_verdict_classifies_as_error`,

`test_desktop_gone_full_child_transcript_classifies_as_error`,

`test_desktop_gone_and_zero_window_detector_strings_are_marker_free` (parent). The

`_detector_unknown_reasons` backstop gained a **third driver scenario** (empty enumerator + alive), so a

future `unknown_reason` branch still forces a matching driver.

### #158 — deliberately NOT done / honest gaps

- **Did not treat "zero windows" as terminal-crashed** — the liveness split is the whole point; a

  starting Desktop must not become a false failure.

- **Did not auto-restore** anything, and **did not touch the PowerShell arbiter** — same scope discipline

  as #154.

- **Did not fix `probe_desktop_query.py`'s separate inline poll loop.** It reads the detection fields via

  its own loop; the new `process_gone` field defaults to `None`, so its behaviour is unchanged (a

  window-less/dead Desktop still reads as `none` there). Bringing `probe_desktop_query.py` onto the same

  `process_gone` handling is a **documented follow-up**, not smuggled into this PR.

- **`_lock._process_alive`'s real Win32 `OpenProcess`/`GetExitCodeProcess` call is UNVERIFIED by this

  PR's unit tests** — the brief forbids launching Desktop, so *every* test injects a fake liveness check

  and the real function is exercised only in production. This is an honest gap: the branch *logic* is

  fully driven by injected fakes, but the actual syscall behaviour on a live PID is not asserted here.

  (The primitive itself is load-bearing for lock reclaim and covered by the existing `_lock` tests, which

  still pass unchanged.)

---

## Rebase onto advanced `master` + a C0302 line-cap fix (extraction)

While this PR was open, `master` advanced by three squash-merges — **#161** (docs), **#159** (probe
wall-clock supervision) and **#160** (dead-code removal, which deleted four symbols: `POLL_SECONDS`,
`detect_credential_modal`, and `_credential_missing` ×2). I rebased onto the new `master` (`8824175`);
the branch is now exactly one commit on top of it.

**Deleted-symbol check (the clean-merge trap): my fix references none of the four.** A clean git merge
would have hidden a reference to a deleted name as a runtime `NameError`, not a conflict. My fix already
uses the live equivalents — `inspect_credential_modal(...)`, `_credential_state(pid)`, and the
`PREFLIGHT_`/`REFRESH_CREDENTIAL_POLL_SECONDS` constants — so there is no latent `NameError`. Verified by
grep against the branch and by the full suite passing post-rebase. No deleted symbol was re-added.

**The C0302 the rebase surfaced.** #159 had *deliberately* pinned `probe_live_source.py` at **1196 /
1200** lines (four under pylint's `max-module-lines`). This PR's parent-classifier additions (the
`CREDENTIAL_UNKNOWN` branch for #154 and the whole `DESKTOP_GONE` branch for #158) add a net **+37**,
taking the module to **1233 lines** — over the cap. pylint prints `10.00/10` right up to the boundary and
only then fails with **`C0302 Too many lines in module (1233/1200)`, exit 16** — exactly the trap #159's
commit message warned about. `ruff` never catches this: this repo's ruff lint selection is
`["E4","E7","E9","F"]`, which excludes the line-length / too-many-lines families.

> Measurement gotcha worth recording: `git show <ref>:<path> | Measure-Object -Line` under PowerShell
> mis-counts (a pipe artifact) — it reported 1067/1101 when byte analysis (`data.count(b"\n")`) proves the
> real counts are 1196 (origin/master) and 1233 (my commit). The repo is `core.autocrlf=true` with no
> `.gitattributes`, so files are CRLF on disk but LF-committed; the line count is identical either way.

**Why extraction, not prose-trimming.** #159 got under the cap by trimming comments, which left the
module at *zero* durable headroom — the very fragility this PR then tripped over. Trimming my own
additions plus adjacent comments only reaches ~1200–1206, recreating that fragility and diluting the
institutional-knowledge comments (the #115/#152/#153 rationale). The proportionate root-cause fix is to
extract the **cohesive verdict-LINE matching layer** — the three `*_VERDICT_RE` constants and the pure
`_has_*_verdict` helpers — into a new module **`scripts/_verdict_lines.py`**. They are pure (`import re`
only), Desktop-independent, and already a distinct concern from `_classify_failure`'s mapping logic. The
knowledge-rich comments move *with* the code (zero loss).

**Behaviour-preserving, tests unchanged.** `probe_live_source` imports the helpers, so they remain module
attributes and the seam tests that reference `probe_live_source._has_desktop_gone_verdict` /
`._has_credential_stop_verdict` / `._has_data_ok_verdict` keep resolving through the re-export — **no test
edits were needed**, which is itself evidence the refactor changed no behaviour.

| extraction | before | after |
|---|---|---|
| `probe_live_source.py` | 1233 lines (**C0302 fail, exit 16**) | **1144 lines** ✅ |
| `scripts/_verdict_lines.py` | — | **112 lines** (new) |
| `pylint scripts` | C0302 fail | **10.00/10, exit 0** ✅ |
| `pylint .github/skills/pbip-model-refresh/scripts` | 10.00/10 | **10.00/10** ✅ |
| `ruff format --check` + `ruff check` (whole dirs) | pass | **pass** ✅ |
| full `pytest -q` | 1373 passed / 7 skipped | **1373 passed / 7 skipped** ✅ (unchanged) |

The #154/#158 fails-before/passes-after evidence above is unaffected — the refactor only relocates the
verdict matchers; the detector, the latch, the exit codes and their regression tests are byte-identical.


Guust-Franssens:
Squash-merged to `master` as `c31be4c`; #154 and #158 closed by that commit.

GitHub refused both the GraphQL and REST merge endpoints (`Merging stacked PRs via this endpoint is not supported`), so this was merged locally with `git merge --squash` and pushed. The full diff landed unchanged: **+1338/-135 across 9 files**, identical to this PR's head `ba8383d`.

Verified before pushing, because the branch had drifted twice: master gained #169 and #180 (branch was 3 behind, merged in, full suite **1392 passed** locally, CI green at `ba8383d`), then gained #167 (targeted suites **209 passed** on the final combination).

The three blocking findings from the blind review were fixed in #168 and re-reviewed by the same reviewer -> **approve**, with production behaviour proven unchanged via docstring-stripped AST comparison across all five changed modules. Platform-guard limitation tracked as #182.

PR #180, opened 2026-08-15
Name declare_generated_edit.py in the personas that must run it (Fixes #166)

### LINKED ISSUE #166: The --tamper sign-off gate is structurally unpassable: personas are told to run it but never told declare_generated_edit.py exists
A mandatory pre-sign-off quality gate is **structurally impossible to pass** for an agent that follows
the documentation faithfully. Found by a blind end-to-end "cold run" of this toolkit (stumble S28).

## The defect

`tableau-migrator`'s step *"Supervise what you delegate"* ends with:

> *"Before sign-off: `check_migration_progress.py --bundle <b> --tamper`; drift blocks"*

and the shared conventions (`AGENTS.md`, synced into every persona) require:

> *"Declare generated edits. TMDL/PBIR/`.pbip`: file/change/why + replay script + hash record."*

Run against a bundle built **exactly as the personas prescribe**: **exit 1, 82 UNDECLARED vs 19
DECLARED.**

The 19 declared are **all self-declarations** written by `refresh_pbip_model.py`, which maintains its
own ledger entry. Every human/agent-authored `_build/fix_*.py` script — which **both builder personas
explicitly instruct you to write** — is undeclared, because **nothing in any agent-facing document
names the tool that declares them.**

`scripts/declare_generated_edit.py` exists and does exactly this. Measured:

- `git grep -l declare_generated_edit -- .github docs AGENTS.md` → **no matches**
- named only in `scripts/README.md` L40 — one row of a ~100-row table — whose "Used by" column reads
  ***"`_build/fix_*.py` authors and `tableau-migrator` sign-off"***. Both of those audiences are told
  to run the gate; **neither is told the tool exists.**

So the loop is closed on paper and open in practice: personas → write `_build/fix_*.py` → gate demands
declarations → the declaration mechanism is documented only where neither audience reads. Any agent
following the docs faithfully arrives at sign-off with a red gate and no instruction on how it could
have avoided it.

**Why this matters more than a missing doc line:** a subagent receives **only its own persona file**.
`AGENTS.md`, `.github/copilot-instructions.md` and user-global instructions do **not** reach a custom
subagent (verified with a sentinel experiment; see the note at the top of `AGENTS.md`). There is no
`include`/`extends`. So "it's in `scripts/README.md`" is equivalent to "it does not exist" for the
agents that need it.

---

## What to deliver

**1. Name the tool in both builder personas.** `.github/agents/pbi-report-builder.agent.md` and
`.github/agents/pbi-semantic-builder.agent.md`, at the exact point each is told to declare its
generated edits. Give the **actual invocation**, not a pointer to another file — a stateless agent must
be able to run it from the persona text alone.

**2. Prove the invocation you document actually works.** Read `scripts/declare_generated_edit.py` and
the gate that consumes its output (`scripts/check_migration_progress.py`), and confirm: required
arguments, where the declaration is written, and how `--tamper` then finds it. **Run it once against a
throwaway example.** Do not document a command you have not executed.

**3. Judge whether S4 shares a root cause.** A sibling finding from the same run: `python
scripts/credential_gate.py verify <bundle>` reports
`GATE VERIFY: VIOLATION - 19 artifact(s) exist while the gate is applied`, where **all 19** are
`reports/<Workbook>.Report/definition.pbir` — i.e. the engine's own pristine report root, which nobody
touched. `reports/` is missing from the allowed-roots list in both `docs/credential-gate.md` and
(evidently) the code, even though the same doc's bundle table calls `reports/` engine-pristine output
that must never be edited. **On any engine-path bundle this makes the authoritative pre-ship check
return a red verdict that is 100% false positives.**

If it shares a root cause with the `--tamper` defect (both gates being blind to the engine's own output
roots), fix both. If it does not, **leave S4 alone and say so explicitly** — it will be filed
separately.

---

## The constraint that makes this hard — read carefully

The agent personas sit at roughly **99% of a hard 30,000-character cap**, and
`python scripts/sync_agent_conventions.py --check` **must exit 0** (it is a CI gate). Adding text may
break the cap.

- Run `uv run python scripts/sync_agent_conventions.py --check` **before** you start to record the
  baseline, and again after.
- The shared-conventions block inside each persona is **GENERATED** from `AGENTS.md`, between
  `<!-- BEGIN:shared-conventions -->` and `<!-- END:shared-conventions -->`. **Never hand-edit that
  block inside a persona** — CI fails on drift. If the right home is genuinely the shared block, edit
  `AGENTS.md` and run `uv run python scripts/sync_agent_conventions.py` to regenerate. But **prefer the
  persona-specific section**: only the two builders need this, and the shared block is duplicated into
  all four personas, so putting it there costs 4x the characters for 2x the benefit.
- If it does not fit, **make room**. Find text in the persona that is redundant, superseded, or now
  lives in one of the `.github/skills/` bundles the persona already invokes, and remove it. Report
  exactly what was removed and why it was safe. **Do not silently truncate anything load-bearing.**

---

## Definition of done

- `uv sync --all-extras`, then from the repo root:
  - `uv run python scripts/sync_agent_conventions.py --check` — **must exit 0**
  - `uv run pytest tests/test_check_migration_progress.py -q`
  - `uv run pytest -q` if the change reaches further
- **Prove the gate is now passable AND still catches real drift** — this is the acceptance test:
  1. a scenario with a properly declared generated edit → `--tamper` **passes**
  2. a scenario with an undeclared edit → `--tamper` **still fails**

  A fix that makes the gate pass unconditionally is a **regression, not a fix**. Paste both runs in the
  PR body.
- If Python changed, the full CI ritual: `uv run ruff format --check scripts tests .github/skills`,
  `uv run ruff check scripts tests .github/skills`, then **all three** pylint roots
  (`uv run pylint scripts`, `uv run pylint .github/skills/pbip-model-refresh/scripts`,
  `uv run pylint .github/skills/powerbi-ai-readiness/scripts`). Ruff alone does **not** predict CI —
  this repo's ruff selection excludes `E501` while pylint enforces a 120-char line.
- PR body reports: before/after character-cap numbers for **both** personas, exactly what was removed to
  make room (if anything), the verdict on S4, and the two `--tamper` runs.

## Out of scope

~27 other findings from the same cold run are being fixed in parallel by other agents. Do **not** edit
`.github/skills/powerbi-report-gotchas/` or `.github/skills/powerbi-semantic-model-gotchas/` — active
work is in flight in both.

Guust-Franssens:
Closes #166. **PR #171 should be closed as superseded** — the cloud agent bailed after 117s with zero
code changes ("this environment prohibits access to `.github/agents/`"), which is exactly where this
fix has to land. This PR was authored locally, where no such restriction applies.

## The defect

A mandatory pre-sign-off gate was structurally unpassable for an agent that followed the docs
faithfully. Both builder personas say *"declare your generated edits"*; `tableau-migrator` blocks
sign-off on `check_migration_progress.py --tamper`; and `scripts/declare_generated_edit.py` — the
**only** thing that writes a declaration — was named nowhere an agent would read it. A custom
subagent receives ONLY its own persona file, so `scripts/README.md` is equivalent to "does not exist".

## What changed

| file | change |
|---|---|
| `.github/agents/pbi-report-builder.agent.md` | new ⚠️ block right after the `_build/fix_*.py` rule: the working invocation + the three measured traps; DoD #2 now gates on `--tamper` exiting 0 |
| `.github/agents/pbi-semantic-builder.agent.md` | new `### Declare every TMDL edit you make` after the workflow, incl. what is *already* covered (`refresh_pbip_model.py`'s own `database.tmdl` bump, `.pbi/` sidecars) vs. what is yours; new DoD #12 |
| `.github/agents/tableau-migrator.agent.md` | step 9 now tells the report builder to run its fixes through the wrapper; the sign-off line says `UNDECLARED` routes back rather than being the orchestrator's own fix |
| `scripts/README.md` | the `declare_generated_edit.py` row now carries both traps |

No Python changed.

## Character cap — before / after

`python scripts/sync_agent_conventions.py --check` exits **0** before and after.

| persona | before | after | Δ |
|---|---|---|---|
| `pbi-report-builder` | 29210 (97%) | **29841 (99%)** | +631 |
| `pbi-semantic-builder` | 25850 (86%) | **28635 (95%)** | +2785 |
| `tableau-migrator` | 29790 (99%) | **29809 (99%)** | +19 |
| `pbi-migration-validator` | 28352 (94%) | 28352 (94%) | — |

### What I removed to make room, and why it was safe

`pbi-report-builder` had only 790 chars of headroom and the insert needed ~1400. Five removals, all
redundant/superseded rather than load-bearing — **nothing was silently truncated**:

1. *"Do not skip straight to authoring — these three skills are explicitly designed as a chained
   handoff…"* — redundant with the section heading (*"Skills you use, **in this order**"*), the
   frontmatter `description`, and the explicit *"If a page must be built from scratch … fall back to
   the full authoring chain"*. It was also **stale and mildly contradictory**: it said "these three
   skills" in a six-item list, and told you not to skip to authoring in a persona whose whole premise
   is that the engine already built the report.
2. The **Model reference** sub-block in *Mandatory validation* (cross-tree `byPath` into a shared
   datasource) — the same ground is covered in more depth by `powerbi-report-gotchas` §3, which the
   persona invokes as step 0 and gates on in DoD #1. Compressed to a one-liner that keeps the one
   detail the skill does **not** carry (the cloud `byConnection` form) and cites §3.
3. The illustrative `reason` quote in the `viz_fidelity[]` table row — the very next paragraph quotes
   a real, longer `reason` verbatim, so it was a duplicate example.
4. Skills item 5 (`preview-*`) compressed — the same CLI verbs are listed again under §(B) precedence 1.
5. *"When the encoding is genuinely unknown"* steps 2–3 compressed; the load-bearing human-round-trip
   procedure is kept in full, only the retrospective framing was dropped.

`tableau-migrator` was already at 99%, so its two small additions were paid for by removing a
duplicated *"Name the mode explicitly"* (step 7 already says it, and step 10's own bold lead names
the mode), the redundant `Proceed only once it reports "Ready to migrate."` (the preceding sentence
already gates on the exit code — and this repo's own rule is to judge by exit code, not printed
text), and two reflows. Net +19 chars.

## Acceptance test — passable AND still catches real drift

Both runs against a throwaway bundle built by `run_estate.write_generated_artifact_manifest` (the
real production function), two generated `visual.json` files, one idempotent `_build/fix_axis_title.py`.

**1. Properly declared generated edits → `--tamper` PASSES**

```
=== SCENARIO 1: declared generated edits ===
patched pbip/ShippingKPIs/ShippingKPIs.Report/definition/pages/p1/visual.json
DECLARE: RECORDED pbip/ShippingKPIs/ShippingKPIs.Report/definition/pages/p1/visual.json -> ...\_build\generated-edit-declarations.json
patched pbip/ShippingKPIs/ShippingKPIs.Report/definition/pages/p2/visual.json
DECLARE: RECORDED pbip/ShippingKPIs/ShippingKPIs.Report/definition/pages/p2/visual.json -> ...\_build\generated-edit-declarations.json
TAMPER [DECLARED_DRIFT] demo-bundle
  DECLARED changed: pbip/ShippingKPIs/ShippingKPIs.Report/definition/pages/p1/visual.json via _build/fix_axis_title.py
  DECLARED changed: pbip/ShippingKPIs/ShippingKPIs.Report/definition/pages/p2/visual.json via _build/fix_axis_title.py
TAMPER EXIT=0
```

**2. One undeclared hand-edit on top → `--tamper` STILL FAILS**

```
=== SCENARIO 2: one undeclared hand-edit on top ===
TAMPER [DRIFT] demo-bundle
  DECLARED changed: pbip/ShippingKPIs/ShippingKPIs.Report/definition/pages/p1/visual.json via _build/fix_axis_title.py
  UNDECLARED changed: pbip/ShippingKPIs/ShippingKPIs.Report/definition/pages/p2/visual.json - record target, baseline hash and expected post-fix hash in _build/generated-edit-declarations.json
TAMPER EXIT=1
```

The gate is passable by following the docs, and is not passable by drifting. Note the second run also
shows the declaration is bound to an exact resulting hash: touching a target *after* declaring
invalidates its own declaration.

## The cloud agent's four claims — confirmed vs corrected

| # | claim | verdict |
|---|---|---|
| 1 | the wrapper records hashes in `_build/generated-edit-declarations.json` | ✅ **confirmed** — `declare_generated_edit.py:23,86-92` |
| 2 | declared drift passes, undeclared drift fails | ✅ **confirmed** — measured above; `check_migration_progress.py:542` maps `CLEAN`/`DECLARED_DRIFT`→0, `DRIFT`→1 |
| 3 | `--bundle <b> --target <rel> --script <b>/_build/fix_*.py` | ❌ **CORRECTED — incomplete, and incomplete in the exact way that reproduces the original bug at scale.** Run verbatim against a fix script that touches two files, it declares **one** and the gate stays red at exit 1 (`TAMPER [DRIFT] … UNDECLARED changed: …/p2/visual.json`). A second wrapper run cannot rescue it: the script is idempotent, so it prints `DECLARE: NO_CHANGE` and records nothing. Documenting that invocation alone would have left the 82-undeclared case at 81. The working form uses the wrapper's `-- <script-args>` passthrough with a `--only <bundle-relative path>` scope argument on the fix script, once per target. |
| 4 | S4 is a separate defect (missing `reports/` exemption) | ✅ **confirmed** — see below |

Two further traps I found that nobody had claimed, both now documented: **you cannot retro-declare a
hand-edit** (the wrapper hashes the target *before* running your script and the gate only accepts a
declaration whose `baseline_sha256` is the engine's — measured: `DECLARE: NO_CHANGE`, exit 0, gate
still exit 1), and **`DECLARE: NO_CHANGE` is a silently useless success**.

## S4 — independent verdict: SEPARATE. Leaving it alone.

I agree with the cloud agent, and the evidence is a one-token omission in a *different* file, gate
and mechanism:

- `scripts/migration_bundle.py:24` — `ENGINE_OUTPUT_DIRS = frozenset({"pbip", "semantic_models", "data"})`.
  **`reports` is absent.** That constant feeds `engine_artifact_records()` (so the receipt never
  records `reports/**` at all) and `credential_gate._is_engine_output()` (so even a receipt entry
  would not exempt it). `reports/<WB>.Report/definition.pbir` has a `.pbir` suffix ∈
  `DEFINITION_SUFFIXES`, is therefore audited, has no receipt record, and lands in `gate_artifacts` →
  `GATE VERIFY: VIOLATION`. 19 workbooks = 19 false positives.
- The issue hypothesised a **shared** root cause — *"both gates being blind to the engine's own output
  roots"*. That half is wrong: `--tamper` is **not** blind to `reports/`.
  `check_migration_progress._is_generated_artifact` matches on any path component ending `.report` /
  `.semanticmodel` **anywhere** in the tree, so `reports/` is fully inside the tamper baseline. My
  demo bundle exercised `pbip/`, but the same matcher covers both.
- Root causes: #166 is **documentation** (a tool nobody was told about; zero code). S4 is **code** (a
  missing set member). Different file, different gate, different fix, and fixing one does not move the
  other. File S4 separately — it is a one-word change plus a test.

## Definition of done

| gate | exit |
|---|---|
| `python scripts/sync_agent_conventions.py --check` | **0** (all four personas ≤ 99% of cap) |
| `pytest tests/test_check_migration_progress.py -q` | **0** — 42 passed |
| `pytest tests/test_sync_agent_conventions.py tests/test_repo_layout.py tests/test_scripts_help.py -q` | **0** — 177 passed |
| `ruff check scripts tests .github/skills` | **0** |
| `pylint scripts` | **0** — 10.00/10 |

No Python was changed, so the other two pylint roots are untouched. Scratch files used for the
acceptance test were deleted; `git status` is clean apart from the four committed files.


Guust-Franssens:
Pushed `834006d`. **F1 accepted in full — the reviewer is right, and I reproduced it independently
before changing anything.** All three items fixed; nothing declined.

## F1 — verified myself, not taken on faith

On a real bundle (`C:\tfmig\coldrun2b\bundle`), the first three `definition.pbir` pairs:

| | `reports/…/definition.pbir` | `pbip/…/definition.pbir` |
|---|---|---|
| book_1-1-Introduction | `../book_1-1-Introduction.SemanticModel` | `../Orders (Sample - Superstore).SemanticModel` |
| book_3-1-Scatter | `../book_3-1-Scatter.SemanticModel` | `../Orders (Sample - Superstore) (2).SemanticModel` |
| book_3-2-Bar-Chart | `../book_3-2-Bar-Chart.SemanticModel` | `../Orders (Sample - Superstore).SemanticModel` |

So "restore from `reports/`" does not restore anything — it substitutes a **different file** that
binds the report to a model that is not there. My line was worse than no line, and it contradicted
`powerbi-report-gotchas` §3, which the same persona invokes at step 0. The `scripts/README.md` twin
survived only because it happened to carry `(or re-run the engine)`; that is luck, not design, so I
removed the `reports/` pointer there too rather than leaving a second copy to be re-imported later.

**Root cause worth naming:** I wrote a restore step I had never executed. Every *other* claim in that
block is measured — which is exactly what made the one unmeasured sentence blend in.

```diff
-  only accepts a declaration whose baseline is the engine's hash, so retro-declaring is a
-  `NO_CHANGE` no-op. Restore the file from `<bundle>/reports/`, then declare.
+  only accepts a declaration whose baseline is the engine's hash, so a retro-declaration is never
+  accepted. Restore the target by re-running the engine — `reports/` is a **different file**, not a
+  copy of `pbip/` (gotchas §3).
```

## `NO_CHANGE` label — fixed in BOTH personas

You flagged it on `pbi-report-builder`; `pbi-semantic-builder:234` carried the same wrong label, so
both now say **"a retro-declaration is never accepted."** That phrasing is true under both mechanisms
— the `NO_CHANGE` path (hand-edit identical to the script's output) and the `RECORDED`-but-rejected
path (hand-edit differs → the wrapper records the *hand-edited* hash as `baseline_sha256`, the gate
compares against the **manifest** baseline, no match → `UNDECLARED`).

**Disclosure:** I confirmed that second path **from source** (`declare_generated_edit.py` hashes
before/after and writes the on-disk pre-run hash; `check_migration_progress._matching_declaration`
requires `baseline_sha256` to equal the manifest baseline), **not** by re-running it — your
reproduction stands as the measurement. The new wording is correct either way, which is why I did not
spend the bundle rebuild to re-derive a result you already have.

## Ellipsis — fixed

The runnable example now spells the `--only` path in full, matching the `pbi-semantic-builder` twin:

```bash
python scripts/declare_generated_edit.py --bundle <b> \
  --target pbip/<WB>/<WB>.Report/definition/pages/<p>/visuals/<id>/visual.json \
  --script <b>/_build/fix_axis_title.py \
  -- --only pbip/<WB>/<WB>.Report/definition/pages/<p>/visuals/<id>/visual.json
```

(The `# DECLARE: RECORDED pbip/.../visual.json` line below it keeps its ellipsis — that is echoed
*output*, not something anyone types.)

## Cosmetic — taken, since I was in that exact region

Blank line restored between the `--tamper` self-check and `### Visual encoding`.

## Gates (by exit code)

| gate | exit |
|---|---|
| `python scripts/sync_agent_conventions.py --check` | **0** |
| `pytest tests/test_check_migration_progress.py tests/test_sync_agent_conventions.py tests/test_repo_layout.py -q` | **0** — 102 passed |

Cap after: `pbi-report-builder` 29941 (99%, +100), `tableau-migrator` 29809, `pbi-semantic-builder`
**28621 (95%, −14)**, `pbi-migration-validator` 28352. No Python touched, so ruff/pylint are
unchanged from the previous run. Diff is 3 files, +9/−6 — deliberately small, as asked.

⚠️ `pbi-report-builder` now has **59 characters** of headroom. That is a real hazard for whoever edits
it next, but buying headroom means trimming unrelated prose, which is exactly the larger diff you
asked me not to take two days out from the workshop. Flagging rather than acting.

PR #232, opened 2026-08-19
Add provenance-tagged row-count hints to migration specs

Guust-Franssens:
## Summary
- Adds table-level `row_count` hints with `{value, source, as_of}` for `hyper`/`live` and explicit `{source: "unknown"}` for unknown.
- Keeps `parse_tableau.py` offline/fast by emitting unknown table hints during parse; adds `extract_hyper_data.py --enrich-spec` as the opt-in Hyper-reading enrichment step.
- Updates schema and tests so bare integers, unknown sources, and `source: "unknown"` with a value are rejected while old specs without `row_count` remain valid.

## Design decisions
1. **Placement:** `data_sources[].tables[].row_count`. The Hyper count is inherently per Hyper relation/table, and the spec already exposes parsed physical/logical tables there. I did not add a datasource total because joins/unions make a datasource aggregate easy to misread.
2. **How parse learns it:** parse does not open Hyper files. It writes `{source: "unknown"}` and the separate `extract_hyper_data.py --enrich-spec` path replaces matching table hints with Hyper counts. This avoids making every parse slower and dependent on tableauhyperapi.
3. **Backwards compatibility:** `row_count` is optional in the schema. Existing specs without it still validate.

## Validation
- `powershell -ExecutionPolicy Bypass -File scripts/preflight.ps1 -Update -CheckUpstream` (exit 0)
- `...\.venv\Scripts\python.exe -m ruff format ...` and `-m ruff check ... --fix` (exit 0)
- `...\.venv\Scripts\python.exe -m pytest tests\test_spec_contract.py tests\test_parse_tableau.py tests\test_extract_hyper_data.py -q` (67 passed)
- `...\.venv\Scripts\python.exe -m pylint scripts` (exit 0, 10.00/10)
- `...\.venv\Scripts\python.exe scripts\validate_spec.py --all` (16 specs valid)
- `...\.venv\Scripts\python.exe -m pytest -q` (1629 passed, 7 skipped, 4 warnings)

## Mutation results
Deliberately broke each new behavior and confirmed the relevant test failed:
- Allowed bare integer `row_count` in the schema -> `test_schema_rejects_bare_integer_row_count` failed.
- Made parser emit unknown as `{"source":"unknown","value":0}` -> unknown round-trip test failed.
- Made `row_count` required on tables -> existing-spec back-compat test failed.
- Changed Hyper enrichment to add 1 to the count -> enrichment persistence test failed.

## What I could not verify
- No live-source row count path was implemented or verified; `live` is schema-supported for future credentialed enrichment only.
- I did not validate ambiguous multi-table Hyper/spec name matching against a real customer TWBX; unmatched/ambiguous tables remain `unknown` by design.

### DIFF
--- a/docs/migration-spec.schema.json
+++ b/docs/migration-spec.schema.json
@@ -294,16 +294,20 @@
   },
   "definitions": {
     "rowCountHint": {
-      "description": "Provenance-tagged row-count hint. Unknown deliberately has no value so consumers cannot mistake absence for zero/small.",
+      "description": "Provenance-tagged row-count hint. Unknown deliberately has no value so consumers cannot mistake absence for zero/small. observed_at is when WE READ the count, not when it became true - the underlying extract may be far older, and nothing here records its refresh time.",
       "oneOf": [
         {
           "type": "object",
-          "required": ["value", "source", "as_of"],
+          "required": ["value", "source", "observed_at"],
           "additionalProperties": false,
           "properties": {
             "value": { "type": "integer", "minimum": 0 },
             "source": { "type": "string", "enum": ["hyper", "live"] },
-            "as_of": { "type": "string", "format": "date-time" }
+            "observed_at": {
+              "type": "string",
+              "format": "date-time",
+              "description": "RFC 3339 timestamp of when this count was READ. format:date-time is annotation-only under jsonschema without the format-nongpl extra, so parse_tableau._schema_contract_errors enforces it with datetime.fromisoformat instead."
+            }
           }
         },
         {
--- a/docs/migration-spec.schema.json
+++ b/docs/migration-spec.schema.json
@@ -80,7 +80,11 @@
                 "id": { "type": "string" },
                 "name": { "type": "string" },
                 "source_relation": { "type": "string", "enum": ["table", "custom-sql", "join", "text", "pivot", "collection", "union"] },
-                "custom_sql": { "type": ["string", "null"] }
+                "custom_sql": { "type": ["string", "null"] },
+                "row_count": {
+                  "$ref": "#/definitions/rowCountHint",
+                  "description": "Best-effort volume hint for this physical/logical table. Hyper counts are credential-free but reflect Tableau's cached extract as of its refresh time; live counts are current but credentialed; unknown is a normal outcome for sources with no credential-free count."
+                }
               }
             }
           },
@@ -289,6 +293,30 @@
     }
   },
   "definitions": {
+    "rowCountHint": {
+      "description": "Provenance-tagged row-count hint. Unknown deliberately has no value so consumers cannot mistake absence for zero/small.",
+      "oneOf": [
+        {
+          "type": "object",
+          "required": ["value", "source", "as_of"],
+          "additionalProperties": false,
+          "properties": {
+            "value": { "type": "integer", "minimum": 0 },
+            "source": { "type": "string", "enum": ["hyper", "live"] },
+            "as_of": { "type": "string", "format": "date-time" }
+          }
+        },
+        {
+          "type": "object",
+          "required": ["source"],
+          "additionalProperties": false,
+          "properties": {
+            "source": { "type": "string", "const": "unknown" }
+          },
+          "not": { "required": ["value"] }
+        }
+      ]
+    },
     "shelfField": {
       "type": "object",
       "properties": {
@@ -320,3 +348,5 @@
     }
   }
 }
+
+
--- a/docs/migration-spec.schema.json
+++ b/docs/migration-spec.schema.json
@@ -306,7 +306,7 @@
             "observed_at": {
               "type": "string",
               "format": "date-time",
-              "description": "RFC 3339 timestamp of when this count was READ. format:date-time is annotation-only under jsonschema without the format-nongpl extra, so parse_tableau._schema_contract_errors enforces it with datetime.fromisoformat instead."
+              "description": "RFC 3339 timestamp of when this count was READ. format:date-time is annotation-only under jsonschema without the format-nongpl extra, so parse_tableau._schema_contract_errors enforces it with a shape check plus datetime.fromisoformat instead. KNOWN EXCEPTION: a seconds field of 60 (a leap second, which RFC 3339 permits) is REJECTED. Python's datetime cannot represent one at all, so accepting it would mean recording a different instant than the one written; fixing it faithfully needs the rfc3339-validator dependency this fallback exists to work without. Disclosed here so the schema does not advertise a format it knowingly does not fully accept."
             }
           }
         },
--- a/scripts/extract_hyper_data.py
+++ b/scripts/extract_hyper_data.py
@@ -177,8 +177,10 @@ def extract_data_sources(
             connection_info = ds["connection"]
             if connection_info["mode"] != "extract":
                 continue
-            hyper_file_name = Path(connection_info.get("hyper_file", "")).name
-            hyper_path = _resolve_hyper_path(hyper_files, connection_info.get("hyper_file") or "")
+            # Normalise once: the spec schema allows a null hyper_file, and Path(None) raises.
+            hyper_file_value = connection_info.get("hyper_file") or ""
+            hyper_file_name = Path(hyper_file_value).name
+            hyper_path = _resolve_hyper_path(hyper_files, hyper_file_value)
             if hyper_path is None:
                 logger.warning("Hyper file not found for %s: %s", ds["id"], hyper_file_name)
                 manifest[ds["id"]] = {"error": f"hyper file not found: {hyper_file_name}"}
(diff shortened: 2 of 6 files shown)


Guust-Franssens:
REQUEST CHANGES: observed_at is documented as JSON Schema date-time / RFC 3339, but the fallback validator accepts non-JSON-Schema timestamps because it uses datetime.fromisoformat() directly.

Reproduction on this branch:

``powershell
Push-Location _scratch\review-docs\wt232
@'
import json
from pathlib import Path
import sys
sys.path.insert(0, str(Path('scripts').resolve()))
from parse_tableau import collect_spec_validation_errors
schema=Path('docs/migration-spec.schema.json')
spec={
 'migration_spec_version':'1.0','source':{'file_name':'x.twb'},
 'data_sources':[{'id':'ds','connection':{'class':'hyper','mode':'extract'},'tables':[{'id':'tbl','name':'T','source_relation':'table','row_count':{'value':1,'source':'hyper','observed_at':'2026-08-19 10:04:00+00:00'}}],'fields':[]}],
 'worksheets':[],'dashboards':[]}
print(collect_spec_validation_errors(spec, schema))
'@ | C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\.venv\Scripts\python.exe -
Pop-Location
``

Output:

``text
[]
``

Ground truth in the diff:
- docs/migration-spec.schema.json:307-310 says observed_at has ormat: "date-time" and is an RFC 3339 timestamp.
- scripts/parse_tableau.py:1672-1701 says it enforces values with datetime.fromisoformat().

A space separator (2026-08-19 10:04:00+00:00) is accepted by romisoformat(), but it is not the JSON Schema/RFC3339 date-time shape the schema advertises. This matters because the fallback is meant to make the contract deterministic when jsonschema lacks ormat-nongpl; it should not silently weaken the documented contract.

Guust-Franssens:
REQUEST CHANGES: the fix closes the space/basic-format over-acceptance and the tests mutation-check correctly, but it now over-rejects RFC 3339 leap-second timestamps.

Ground truth:
- RFC 3339 section 5.6 defines `time-second = 2DIGIT ; 00-58, 00-59, 00-60 based on leap second rules`, and section 5.8 gives `1990-12-31T23:59:60Z` as an example. Source: https://www.rfc-editor.org/rfc/rfc3339.txt
- This branch rejects that shape after the regex, because `datetime.fromisoformat()` cannot parse second 60.

Reproduction on this branch:

```powershell
Push-Location C:\tfmig\wt-volume
<run a spec with observed_at values below through collect_spec_validation_errors>
```

Observed output:

```text
2026-08-19t10:04:00z ACCEPTED []
2026-08-19T10:04:00-00:00 ACCEPTED []
2026-08-19T10:04:00+14:00 ACCEPTED []
2026-08-19T23:59:60Z REJECTED ["$.data_sources[0].tables[0].row_count.observed_at: '2026-08-19T23:59:60Z' is not a valid RFC 3339 timestamp (second must be in 0..59)"]
```

Other checks passed:
- Targeted timestamp tests: `9 passed`.
- `python -m pylint scripts`: exit 0, `10.00/10`.
- Mutation tests: removing the regex shape gate fails the two new bad cases (`space` and `basic` format); removing `.upper()` fails lowercase `t`/`z` acceptance. Score for these targeted mutations: 2/2 caught.
- I could not construct a `.upper()` meaning-change counterexample after the regex: the accepted alphabetic characters are only `T/t/Z/z`, so uppercasing only normalizes those RFC3339 case-insensitive markers.

Guust-Franssens:
Partial pushback on the leap-second finding, with evidence — the finding is correct, but the framing changes what the right fix is.

**It is not a regression from this PR.** The previous check called `datetime.fromisoformat(raw)` directly, which rejects a leap second identically:

```
>>> datetime.fromisoformat('2026-12-31T23:59:60Z')
ValueError: second must be in 0..59
>>> datetime.fromisoformat('2026-12-31T23:59:60+00:00')
ValueError: second must be in 0..59
```

So the shape regex did not introduce it, and removing the regex would not fix it.

**It is a stdlib ceiling, not a choice.** `datetime` has no leap-second representation at all. Even if parsing succeeded there would be nothing to store or compare — accepting the value would mean silently recording a *different instant* than the one written, which is worse than rejecting it. Fixing it faithfully needs a dependency (`rfc3339-validator`), and the entire point of this fallback is that it works when that dependency is absent.

**What WAS fixable is the message**, and that is now done (`e5dab75`): a generic *"not a valid RFC 3339 timestamp"* sends the reader hunting for a typo in a timestamp that is genuinely valid. It now names the leap second and says why it cannot be represented, with a test pinning it as a documented limitation rather than an accident.

Happy to be argued out of this — if you think the deviation belongs in the schema docs instead, or if you can find a stdlib-only way to accept it faithfully, say so.

Guust-Franssens:
REQUEST CHANGES after adjudication.

Ruling on the disagreement:

1. **Not a regression from the regex PR: confirmed.** Checking out the parent parser and running `collect_spec_validation_errors()` on `2026-12-31T23:59:60Z` produced the previous generic error: `second must be in 0..59`. So the regex did not introduce the leap-second rejection.

2. **Not faithfully fixable with `datetime` alone: confirmed.** Python `datetime` has no `second=60` representation; accepting it by coercing to `59` or next-second would record a different value/instant. If you want full RFC3339 leap-second acceptance, that needs a parser/validator that models RFC3339 rather than bare `datetime`.

3. **But two changes are still needed.**

   **A. The schema/contract should disclose the exception.** `docs/migration-spec.schema.json` still advertises `format: date-time` / RFC3339 semantics for `observed_at`, while the repo knowingly rejects a valid RFC3339 subset (leap seconds). The code/test comments explain the limitation, but schema consumers see the schema contract first. Please add the limitation to the schema description (for example, RFC3339 date-time except leap-second `:60` values, which are rejected because Python datetime cannot represent them).

   **B. The new leap-second branch is over-broad.** `raw[17:19]` is the correct index for the seconds field for every string the regex accepts (`YYYY-MM-DD[Tt]hh:mm:ss...`), including fractional seconds and both offset forms. But checking only that slice mislabels ordinary invalid timestamps as leap seconds.

   Reproduction on this branch:

   ```text
   2026-12-31T23:59:60Z REJECTED "... is a leap second; Python's datetime cannot represent one ..."
   2026-12-31T23:59:60.123Z REJECTED "... is a leap second; Python's datetime cannot represent one ..."
   2026-12-31T23:59:60+00:00 REJECTED "... is a leap second; Python's datetime cannot represent one ..."
   2026-12-31T23:59:60.123+14:00 REJECTED "... is a leap second; Python's datetime cannot represent one ..."
   2026-12-31t23:59:60z REJECTED "... is a leap second; Python's datetime cannot represent one ..."
   2026-12-31T23:58:60Z REJECTED "... is a leap second; Python's datetime cannot represent one ..."
   2026-12-31T23:59:59Z ACCEPTED []
   ```

   `23:58:60Z` is not a valid leap-second timestamp under RFC3339 section 5.7; it is just an invalid seconds value. The branch should not call every `ss == 60` value "a leap second" unless it also scopes the wording to "seconds field is 60" or verifies the valid leap-second position.

Other checks:
- New leap-second test mutation: removing the leap-specific branch makes `test_a_leap_second_is_rejected_with_an_HONEST_reason` fail. Mutation caught.
- Baseline targeted tests pass: `10 passed` for the observed_at tests.
- `python -m pylint scripts`: exit 0, `10.00/10`.
- CI on `feat/spec-row-count`: latest two runs are `completed success` (`32443445889`, `32443405701`).

Guust-Franssens:
APPROVE: the final fix is correct; no further changes needed from me.

Checks performed:
- Reviewed `git diff HEAD~1 HEAD` for `3824f9a`.
- Probed regex-accepted second-60 shapes:
  - `2026-12-31T23:59:60Z`
  - `2026-12-31T23:59:60.123Z`
  - `2026-12-31T23:59:60+00:00`
  - `2026-12-31T23:59:60.123+14:00`
  - `2026-12-31t23:59:60z`
  - `2026-12-31T23:58:60Z`

  All rejected with the scoped message `has a seconds field of 60...`, not with the over-specific `is a leap second`. The fixed index `raw[17:19]` is the seconds field for every regex-accepted shape because the regex fixes `YYYY-MM-DD[Tt]hh:mm:ss` before any optional fraction/offset.
- Control value `2026-12-31T23:59:59Z` is still accepted.
- Schema disclosure is acceptable: it now tells schema consumers that `observed_at` is RFC3339 except for the known stdlib leap-second/seconds-60 ceiling, and explains why accepting it faithfully needs a non-stdlib validator/representation.
- Mutation-tested both new assertions:
  - Removing the positive `seconds field of 60` phrase fails the test.
  - Reintroducing the forbidden `is a leap second` phrase fails the negative assertion.
- CI: latest two `gh run list --branch feat/spec-row-count --limit 2` entries are both `completed success` (`32445252761`, `32445249437`).

Verdict: APPROVE.

PR #243, opened 2026-08-20
feat(handover): a reader for the engine work queue - dedupes guidance, surfaces emptied visuals

Guust-Franssens:
> ⚠️ **The original framing of this PR was wrong and has been retracted in [`2c77f09`](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/commit/2c77f09).** The body below is the corrected version. See "What I got wrong" at the bottom.

## What this adds

`scripts/read_handover.py` — a reader for the engine's per-workbook handover slice.

A slice is large (**347 KB** for a 60-stub workbook) and most of that bulk is redundant. Reading it by hand means recovering from a file-read refusal, parsing it programmatically, then filtering 9 fields × 60 requests down to the one category you are about to author. Every agent pays that round trip on every workbook.

## Why it earns its place

**1. It surfaces emptied visuals, which nobody had looked at.** `pbip_ref_drops[].emptied` marks visuals whose *every* field binding was dropped — they render blank on a report that validates clean. There are **15** in the worked example, sitting unremarked beside a 170-item `remediation_worklist`. Nothing else ranked them. `--severity` deliberately never hides them.

**2. It de-duplicates `category_guidance`**, which the engine emits **per request** rather than per category. Verified across all 38 handovers in `_bundle-208`: exactly **one distinct guidance string per category, estate-wide** — so a 60-request file carries 60 copies of an 886-character block, about **53 KB of pure repetition**, and most of why the file is awkward. All seven categories together cost 4,481 bytes.

**3. `--list` gives estate-wide triage** of both queues in one view.

**4. It never truncates silently** — over `--max-bytes` it names every omitted item and the command that prints it. Not because a silent truncation was ever observed, but because a tool whose job is fitting a large queue into a small window is exactly where that would be easy to introduce.

**5. It forces UTF-8 stdout** — a real Tableau formula containing `U+25B2` crashed the tool under cp1252.

1.6 KB of output replaces a 347 KB file.

## Persona and skill edits

Both builder personas point at the reader and note that `requests[]`, not `needs_review[]`, is the list carrying `formula`. Detail lives in the **uncapped skills** (`powerbi-semantic-model-gotchas` §8, `powerbi-report-gotchas` §10) rather than the personas, which sit at 97–99% of their 30,000-char cap.

§8 also closes a real content gap: `_GUIDANCE[MODEL_OBJECT_PARAMETER]` routes to **field parameters**, but when every branch is already a working measure a plain **`SWITCH`** is the lighter faithful fix — as proven by the working 8-branch `SWITCH` sibling sitting ~400 lines below the stub in the same file.

## What I got wrong

The first version of this PR claimed the reader fixed a **silent** failure: that a plain read returned ~20 KB with no error, landed on `needs_review[]`, and left the consumer believing it had the whole queue — and that this **caused** three customer workbooks to ship bare `BLANK()` dispatchers.

Falsified two ways, within hours:

1. **The read fails loudly.** The file-read tool refuses oversized files outright — *"File too large to read at once (346.9 KB)"* — with a suggested remedy. Measured at both 20.7 KB and 346.9 KB. **There is no silent-truncation band.**
2. **A controlled A/B refuted the consequence.** Two fresh agents, same task (quote the full Tableau formula for the stubbed calc `Group Sort`), one given the **old** *"read `handover/<workbook>.json`"* instruction and one given the reader. **Both succeeded.** The control hit the size error, recovered by parsing with `ConvertFrom-Json`, and returned the complete 3-branch `CASE`, the right `target_table` and correct queue totals — describing the recovery as *"reactive, triggered by a size error"*.

The **measurements** were all correct — byte offsets, the `needs_review`/`requests` field asymmetry, the duplication. What was wrong was the **consequence inferred from them**, and then a **cause asserted on top of that inference**. There was never any evidence of what the field agents actually did; what caused those three dispatchers is **unknown and now recorded as unknown**.

The retraction is kept in `powerbi-semantic-model-gotchas` §8 rather than deleted, because how it was wrong is more instructive than the entry was: *a byte offset past a documented read limit looks like proof of unreachability, and is only proof of a byte offset.*

Upstream issue [`#170`](https://github.com/Yarbrdab000/tableau-fabric-skills/issues/170) has been corrected and retitled to the one ask that survives (de-duplicate `category_guidance`).

## Verification

- `ruff format --check` / `ruff check` on `scripts tests .github/skills` — clean
- `pylint` **10.00/10, exit 0** on all three CI roots
- **1687 tests pass**; 23 new
- `sync_agent_conventions.py --check` OK, all four personas under cap
- Smoke-tested across **all 38 real handovers × 2 views, 0 failures**
- **New tests mutation-tested 12/12.** One initial "escape" was traced to a bad harness anchor rather than a test gap, and re-verified after fixing it.


Guust-Franssens:
Validation note before merge: this PR edits .github/skills/powerbi-semantic-model-gotchas/SKILL.md, so installed plugin copies need a content refresh with `python scripts/sync_installed_skills.py`; otherwise preflight can report STALE in plugin and subagents may read the old bundle copy. Local validation on fix/handover-reader-land / feat/handover-reader: sync_agent_conventions.py --check passed, check_navigation_index.py passed, ruff format/check + pylint scripts passed, pytest passed (1963 passed, 7 skipped), and read_handover mutation spot-check killed 5/5 mutants.

PR #260, opened 2026-08-21
fix(gates): check that a visual's fields agree on a joinable table set (#258)

### LINKED ISSUE #258: check_field_bindings.py is silent when a field name exists on BOTH the bound table and the correct one - surfaces later as InvalidUnconstrainedJoin
**Source:** customer field report, from live use on a real estate. This is a gap in a tool that is otherwise working well - it replaced an afternoon of manual Desktop archaeology per workbook.

## The defect

Found on `IA_Aircraft_Installs`. Verbatim:

> a field existing under the identical name on BOTH the referenced table and the correct table produces no gate warning - it only surfaces later as `InvalidUnconstrainedJoin` ("Can t determine relationships between the fields"), and only after the genuinely-missing fields are fixed. Manual cross-check required: verify every field in a visual agrees on one table, not just that each field resolves somewhere.

The checker current question is *"does each field resolve somewhere in the model?"* A field bound to the **wrong table** still resolves - because a column of that name also exists there - so the checker stays silent, and the report fails at render time in Desktop.

## The ordering property is what makes it dangerous

The failure is **masked** until the genuinely-missing fields are fixed. So the sequence is: run the tool, fix the real errors it found, re-run, **get a clean bill of health**, and only then discover the report still does not render. A silent pass here is actively misleading, not merely incomplete.

## Ask

Per visual, verify the fields **agree on a coherent table set**, not merely that each resolves. The rule needs care or it will produce false positives that destroy trust in the tool:

- single table - fine
- several tables joined by **active relationships** - normal and correct in a star schema (fact measure + dimension attribute is the most common visual there is); must **not** warn
- no relationship path, or an inactive one - the real defect
- **measures** are not table-bound the way columns are; how they participate needs deciding deliberately rather than treating their home table as a hard constraint
- where a name resolves on **more than one** table, say so and name the candidates - the ambiguity is itself the finding, even when the intent cannot be determined

Likely needs relationship parsing out of TMDL, which the tool may not do today. A narrower honest check ("this name is ambiguous across X and Y - verify it binds to the one you intend") is an acceptable fallback if the full version is disproportionate.

Companion to #236/#247.

Guust-Franssens:
# The defect

`check_field_bindings.py` asked one question: **"does each field resolve somewhere in the model?"**

A field bound to the **wrong table** still resolves, when a column of that name also exists there.
So the gate stayed silent and the report failed later, in Desktop, as `InvalidUnconstrainedJoin` —
*"Can't determine relationships between the fields"*. Verbatim from the field report on
`IA_Aircraft_Installs`:

> a field existing under the identical name on BOTH the referenced table and the correct table
> produces no gate warning — it only surfaces later as `InvalidUnconstrainedJoin`, and only after
> the genuinely-missing fields are fixed. Manual cross-check required: verify every field in a
> visual agrees on one table, not just that each field resolves somewhere.

**The ordering property is what makes it dangerous.** The failure is *masked* until the
genuinely-missing fields are fixed. Fix the real errors the tool found, re-run, get a **clean bill
of health** — and only then discover the report still does not render. A silent pass is actively
misleading, not merely incomplete. Both questions are now answered in the **same pass**, so nothing
is masked behind anything else.

# What "coherent table set" means here

The crux, and deliberately narrow — a naive rule produces false positives that destroy trust in a
tool people already rely on. A visual's **grouping columns** must fall inside one connected
component of the **active** relationship graph. Five exclusions, each of which would otherwise be a
false positive:

| decision | why |
|---|---|
| **Measures are excluded entirely** | A measure aggregates across the model; its home table is an organisational choice (a `_Measures` table is disconnected *on purpose*). In `SUMMARIZECOLUMNS` it is a **value**, never a grouping key — only grouping keys can raise an unconstrained join. Treating a measure's home table as a hard constraint would flag every model with a measure table. |
| **Transitive reachability, not adjacency** | `Installs → Aircraft → Manufacturer` is ordinary snowflake shape. Union-find over active relationships, so a fact + dimension visual — the most common visual in any star schema — stays silent. |
| **Field parameters and calculation groups are exempt** | Substituted at query generation; never joined. Detected by `extendedProperty ParameterMetadata` (Desktop) **or** `= calculated` + `NAMEOF(` (this engine's emit), or `calculationGroup`. |
| **A plain disconnected slicer table is NOT exempt** | `superstore-sales-performance` ships **both** shapes — `X-Axis` is a field parameter, `Region Parameter` is a plain list table. Exempting everything disconnected would swallow the real defect. |
| **Only `visual.json` `queryState` projections count** | A visual-level filter, sort key or conditional-formatting `FillRule` input on an unrelated table renders fine — it simply filters nothing. |

Two more, smaller: **a reference that does not resolve does not vote** (it is already reported
`missing`; speculating about its table bills the user twice for one fix, and the run is red either
way), and **ambiguity candidates are restricted** to tables the rest of the visual can already
reach — a global census would list `Turbine Id` on four tables and drown the finding.

# ⚠️ Contract change

New top-level status **`INCOHERENT`**, exiting **1**, like `UNRESOLVED`.

- A caller testing `status == "OK"` is **unaffected**.
- A caller testing `status != "UNRESOLVED"` would now **let a non-rendering report through**.

**Gate on the exit code, not the status string.** `main()` now uses `FAILING_STATUSES =
frozenset({"UNRESOLVED", "INCOHERENT"})`. Nothing in `run_estate.py` or `preflight.ps1` consumes
this script today, so no in-repo caller had to change; `scripts/README.md` documents the change.

# Reproduction (written first, watched fail)

`test_a_field_bound_to_the_wrong_table_of_the_same_name_is_caught` builds the customer's exact
shape — `Serial Number` on **both** `Installs` (bound) and `Aircraft` (intended), joined to a table
the rest of the visual cannot reach. Against the old code it failed with:

```
assert 'OK' == 'INCOHERENT'
```

Old code: silent, exit 0. New code: `INCOHERENT`, exit 1, naming both candidate tables.

# False-positive evidence

Over the committed corpus — **357 visuals, 16 example models** — the new check reports **0
findings**. On its own that is **weak evidence**, and I would rather say so than reassure: only
**one** shipping visual groups by two or more tables, so 355 of them cannot exercise the rule at
all.

What actually proves the rule is live is two **controlled experiments**, both now pinned as
regression tests against the committed corpus:

| experiment | result |
|---|---|
| baseline | **0** findings |
| relationships ignored | **1** — `airline-alliance-activity` pivotTable `dfd1e77a…` (`Date` + `Flight Activity`), silent in reality only because an active relationship joins them |
| field-parameter/calc-group exemption removed | **1** — `broadway-stage-to-screen` pivotTable `19cadb01…` becomes a false positive |
| both removed | **2** |

Also measured: **11 of the 16 example models already have the same column name on more than one
table** (`wind-energy-utilization` carries `turbine id` on four; `sales-commission-model` has
`sales` and `order date` on two tables with **zero** relationships). The ingredients of the defect
are everywhere in the corpus — no committed visual happens to mis-bind, so the reproduction is
**fixture-only**, which the report states plainly rather than implying live coverage.

# Mutation testing

20 mutations of the new behaviour, judged by exit code. First pass **17 caught / 3 survived**
(ambiguity-candidate restriction, the `visual.json`-only glob, the missing-field exclusion) — each
survivor was a real untested decision, and each got a killing test. Second pass **20/20 caught**.

Because a perfect sweep deserves suspicion rather than celebration, the harness was calibrated with
negative controls: two semantically-neutral edits correctly reported **SURVIVED**, and a deliberate
syntax error reported **CAUGHT**.

# Known scope limit

A **single** wrong-table grouping column plus measures produces **wrong numbers silently** rather
than a render failure — and is structurally indistinguishable from the legitimate
disconnected-parameter idiom. Deliberately not flagged; documented rather than hidden.

Fixes #258


Guust-Franssens:
REQUEST CHANGES: table-agreement misses alias-shaped visual projections.

Reproduction (run from the PR worktree):

```powershell
cd _scratch\review-bindings
..\..\.venv\Scripts\python.exe alias_query_probe.py
```

`alias_query_probe.py` creates a minimal PBIP where the visual query declares aliases in `visual.query.From` and projections refer to them with `SourceRef.Source`:

- `c` -> disconnected `Config`, field `Serial_Number`
- `i` -> disconnected `Installs`, field `Install_Count`

Every field resolves, but the visual groups by two unrelated tables, so this is exactly the issue #258 class. Actual output:

```json
{
  "status": "OK",
  "references": 2,
  "missing": 0,
  "near_misses": 0,
  "visuals": 0,
  "incoherent_visuals": 0,
  "coherence": []
}
```

and the rendered verdict is:

```text
FIELD BINDING CHECK: OK - every PBIR field reference in 1 report(s) resolves in its model.
```

Why: `iter_references()` walks the full JSON and sees the `query.From` alias scope, but `iter_visual_queries()` calls `_walk(state, {}, ...)` on only `queryState`, so `SourceRef.Source` projections cannot resolve and are dropped from the coherence check. This moves the failure boundary from direct `SourceRef.Entity` projections to alias-shaped query projections.

Guust-Franssens:
Blind review verdict: REQUEST CHANGES.

Finding already posted above: alias-shaped visual projections are dropped by table-agreement. The checker resolves the fields in the per-field pass but reports `visuals: 0` / `incoherent_visuals: 0` and prints OK for a visual grouping by two unrelated aliased tables.

Validation I ran:

```powershell
cd _scratch\review-bindings
..\..\.venv\Scripts\python.exe -m pytest tests\test_check_field_bindings.py -q
# 49 passed in 40.54s

..\..\.venv\Scripts\python.exe scripts\check_field_bindings.py examples --quiet
# binding-examples-exit=0
```

Mutation testing (harness classified only pytest exit 1 with real FAILED output as caught): 8/8 sampled mutants caught, 0 harness survivors.

Caught mutants:
- coherence disabled
- measures treated as grouping constraints
- inactive relationship treated active
- field-parameter/calculation-group exemption disabled
- top-level relationship blocks swallowed into preceding table
- alias scope disabled for the per-field walker
- engine baseline outside `pbip/` scanned
- `INCOHERENT` no longer exits non-zero

The surviving failure boundary is not a mutation of existing code; it is an uncovered input class: `query.From` aliases used by `queryState` projections.

Guust-Franssens:
Re-review verdict: APPROVE / no further changes needed.

I re-reviewed only commit `ac5559c` and re-ran the original repro plus additional scope-boundary probes.

Validation:

```powershell
C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\.venv\Scripts\python.exe -m pytest tests\test_check_field_bindings.py -q
# 50 passed in 7.06s
```

Committed-bundle controls:

```text
examples: 16 committed report(s) with definition.pbir
migrations: 0 committed report(s) with definition.pbir
check examples: exit=0
visuals=869 query_from=0 queryState_Source=0
```

Scope probes I ran under `_scratch\re-review-260`:

```text
parent_alias_unrelated -> status=INCOHERENT, references=2, visuals=1, incoherent_visuals=1
parent_alias_related   -> status=OK,          references=2, visuals=1, incoherent_visuals=0
nested_shadow_no_leak  -> status=INCOHERENT, references=2, visuals=1, incoherent_visuals=1
```

That covers the main boundary questions: seeding from the parent `query` is the right scope boundary for `query.From`; nested `From` scopes still shadow locally via `_walk`; and the shadowed alias did not leak to a sibling projection.

Mutation-test of the new regression:

```text
revert_fix_exit 1
FAILED ... assert result["reports"][0]["visuals"] > 0
E AssertionError: an aliased visual must be JUDGED, not silently skipped
E assert 0 > 0

revert_fix_with_direct_entity_helper_exit 0
1 passed
```

So the new test is not passing incidentally: reverting only the production fix fails, while mutating the helper to non-aliased `SourceRef.Entity` makes the reverted implementation pass, proving the alias shape is load-bearing.

Caveat: I could not prove Power BI Desktop emits this exact aliased `queryState` shape in committed report visuals; the committed corpus has 0 occurrences. I still think the fix is a correctness win rather than unnecessary complexity because the per-field walker already supports `query.From`/`SourceRef.Source`, and this one-line change makes the agreement pass use the same valid scope model without changing the direct-entity path.

PR #261, opened 2026-08-21
docs(agents): ask who drives data refreshes in the intake, not mid-run

### LINKED ISSUE #253: refresh_pbip_model.py: REFRESH_TIMEOUT_SECONDS is hardcoded at 300s, but measured Snowflake refreshes take 700-750s
**Source:** customer field report, live estate. Directly benchmarked by them, not estimated.

## The defect

`.github/skills/pbip-model-refresh/scripts/refresh_pbip_model.py` hardcodes `REFRESH_TIMEOUT_SECONDS = 300` (330 s with grace) with **no way to override it**. Measured against their Snowflake source:

| table family | measured refresh |
|---|---|
| `Flight_Level_Query` | ~700-750 s |
| `Tail_Query` | ~452 s |

Both exceed the ceiling, so the wrapper cannot refresh these models at all.

## Two things that make this more than "raise the number"

1. **`--tables` narrowing does not help** when the bottleneck is *one table own duration* rather than cumulative cost. Proven twice by the customer, identical timeout both times. So a per-invocation override is required, not a smarter default.
2. **The manual XMLA/TOM fallback has no ceiling, but has a scope constraint:** it must be **whole-database** scope if a calculated table depends on a refreshed table - e.g. a `Date` table built with `CALENDAR(MINX(...), MAXX(...))` over the fact. Refresh the fact alone and the calculated object does not recompute in the same transaction.

## Ask

- Make the timeout configurable (CLI flag and/or env var), keeping the current value as the default.
- **Keep the self-bounding behaviour.** The script announcing its own deadline is load-bearing: agents are told to time-box external calls at ~2 min *unless the tool is itself the timer*, and one previously killed this script at 120 s and recorded no verdict at all.
- Document the whole-database-scope rule for the XMLA/TOM fallback.

Guust-Franssens:
Adds a fifth question to the dispatcher intake: **who drives the data refreshes?**

## Why this is a question and not a default

A refresh is the one routine step with **no progress signal** — nothing streams row counts, so *"still working"* and *"hung"* are indistinguishable from the outside. That forces the decision to land at the worst possible moment: **mid-flight, under uncertainty, on the agent.**

Worse, the two rules that govern it **point in opposite directions**:

- the general rule says time-box an unresponsive external system at ~2 minutes or 3 attempts;
- the carve-out says *don't* — if the tool announces its own deadline. And `refresh_pbip_model.py` is exactly such a tool.

An agent resolving that tension wrongly fails in one of two measured ways: it kills a legitimately-running refresh at 120 s and records **no verdict at all**, or it waits indefinitely (**129 minutes / 298 tool calls** before a human intervened).

Asking up front removes the tension entirely — the agent no longer has to infer which case it is in.

## The default's ceiling is known to be too low

`REFRESH_TIMEOUT_SECONDS = 300` (330 s with grace) is hardcoded and non-configurable. Measured against a customer's Snowflake source:

| table family | measured |
|---|---|
| `Flight_Level_Query` | ~700–750 s |
| `Tail_Query` | ~452 s |

Both exceed the ceiling. And `--tables` narrowing does **not** rescue it when a *single table* is the bottleneck rather than cumulative cost — proven twice, identical timeout both times. See #253.

## What's added

| strategy | who drives it | ceiling | progress visible |
|---|---|---|---|
| **`scripted`** (default) | `refresh_pbip_model.py` | hard 300 s / 330 s | ❌ banner, then silence |
| `operator` | agent prepares, stops, **you** hit Refresh in Desktop | none | ✅ per-table row counts, live |
| `xmla` | manual XMLA/TOM | none | ⚠️ partial |

`operator` trades start-latency for a refresh that cannot silently time out and that you can actually watch. Choosing it costs one line in the brief; discovering you needed it costs the refresh.

Also documents the constraint `xmla` carries: it must be **whole-database** scope when a calculated table depends on a refreshed table (e.g. a `Date` table built with `CALENDAR(MINX(...), MAXX(...))` over the fact), or the calculated object won't recompute in the same transaction.

## Scope

- Lands in the **dispatcher** section, which is **outside** the `<!-- BEGIN:shared-conventions -->` block — so **no persona regeneration**. `scripts/sync_agent_conventions.py --check` exits **0**, reporting all 4 personas still current. That matters here: `tableau-migrator` is at **99%** of the 30,000-char cap, so this deliberately does *not* add persona text. The brief carries the answer to the executing agent, which is the documented mechanism.
- Docs-only. No behaviour change, and it does **not** fix #253 — it makes the limitation a decision you take knowingly instead of one you discover at 330 s.

Origin: customer field report — *"the refreshes are just a pain point… there is just no proper progress bar."*

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -626,15 +626,18 @@ source was reachable.
 
 | strategy | who drives it | ceiling | you can see progress |
 |---|---|---|---|
-| **`scripted`** (default) | `refresh_pbip_model.py` | **hard 300 s** (330 s with grace), not configurable | ⚠️ an elapsed-time heartbeat only — `still refreshing, 42s / 300s` |
+| **`scripted`** (default) | `refresh_pbip_model.py` | **hard 300 s** (330 s with grace), not configurable | ⚠️ an elapsed-time heartbeat only — `still refreshing, 42s / 330s` |
 | `operator` | the agent prepares everything, stops, and asks **you** to hit Refresh in Desktop | none | ✅ per-table row counts, live in the UI |
 | `xmla` | manual XMLA/TOM against the live instance | none | ⚠️ partial |
 
 **Why it earns a slot in the intake instead of being discovered mid-run.** A refresh is the only
 routine step where *"still working"* and *"hung"* produce an identical signal — the scripted path's
 heartbeat reports **elapsed time, not work done** (`print_refresh_heartbeat` is documented as
-printing "an elapsed/total countdown *without claiming progress*"), so it ticks the same whether the
-engine is reading rows or wedged on a credential prompt — so the decision lands at the worst possible
+printing "an elapsed/total countdown *without claiming progress*"), so a slow refresh and a stuck one
+print the same line. Be precise about what it does catch: a **detected** credential modal does not
+heartbeat on, it aborts with a specific error. What survives is the case the detector cannot see —
+and the script says so itself on timeout: *"CAUSE UNKNOWN - this script cannot distinguish these two,
+and they need opposite responses"* — so the decision lands at the worst possible
 moment: mid-flight, under uncertainty, on the
 agent. Worse, the two governing rules **point in opposite directions** there. The general rule says
 time-box an unresponsive external system at ~2 minutes or 3 attempts; the carve-out says *don't*, if
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -626,13 +626,16 @@ source was reachable.
 
 | strategy | who drives it | ceiling | you can see progress |
 |---|---|---|---|
-| **`scripted`** (default) | `refresh_pbip_model.py` | **hard 300 s** (330 s with grace), not configurable | ❌ a start banner, then silence |
+| **`scripted`** (default) | `refresh_pbip_model.py` | **hard 300 s** (330 s with grace), not configurable | ⚠️ an elapsed-time heartbeat only — `still refreshing, 42s / 300s` |
 | `operator` | the agent prepares everything, stops, and asks **you** to hit Refresh in Desktop | none | ✅ per-table row counts, live in the UI |
 | `xmla` | manual XMLA/TOM against the live instance | none | ⚠️ partial |
 
 **Why it earns a slot in the intake instead of being discovered mid-run.** A refresh is the only
-routine step where *"still working"* and *"hung"* produce an identical signal — nothing streams
-progress — so the decision lands at the worst possible moment: mid-flight, under uncertainty, on the
+routine step where *"still working"* and *"hung"* produce an identical signal — the scripted path's
+heartbeat reports **elapsed time, not work done** (`print_refresh_heartbeat` is documented as
+printing "an elapsed/total countdown *without claiming progress*"), so it ticks the same whether the
+engine is reading rows or wedged on a credential prompt — so the decision lands at the worst possible
+moment: mid-flight, under uncertainty, on the
 agent. Worse, the two governing rules **point in opposite directions** there. The general rule says
 time-box an unresponsive external system at ~2 minutes or 3 attempts; the carve-out says *don't*, if
 the tool announces its own deadline — and `refresh_pbip_model.py` is exactly such a tool. An agent
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -587,7 +587,7 @@ not exist yet. Getting a finished model/report into a Fabric workspace is a manu
 Desktop publish. So the target question is not "local or Fabric?" but *"note the destination
 workspace in the brief so the manual publish is unambiguous."* Do not imply an automated deploy.
 
-### Step 2 — four questions, asked ONCE, in one message
+### Step 2 — five questions, asked ONCE, in one message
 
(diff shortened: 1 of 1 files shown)


Guust-Franssens:
REQUEST CHANGES: the new intake text says the scripted refresh gives "a start banner, then silence" and later says "nothing streams progress", but the actual scripted path emits periodic heartbeat output while the worker is alive.

Ground truth on this branch:
- `AGENTS.md:629` documents scripted as `a start banner, then silence`.
- `.github/skills/pbip-model-refresh/scripts/refresh_pbip_model.py:399-405` calls `join_with_credential_poll(... heartbeat_seconds=REFRESH_HEARTBEAT_SECONDS ...)` for PID-scoped refreshes.
- `.github/skills/pbip-model-refresh/scripts/_credential_modal.py:458-461` defines `print_refresh_heartbeat()` as `print(f"still refreshing, {int(elapsed)}s / {int(total)}s", flush=True)`.
- `.github/skills/pbip-model-refresh/scripts/_credential_modal.py:560-562` emits that heartbeat while the worker is still alive.

This is a documentation bug, not just wording: agents follow AGENTS.md literally, and the real limitation is "no table/source progress" or "no proof of work/progress beyond elapsed-time heartbeats", not silence.

Guust-Franssens:
REQUEST CHANGES: the fix is directionally right about elapsed-time heartbeats, but the revised wording is still inaccurate in two places.

1. Visible credential-modal blocks do **not** keep heartbeating. The branch says the heartbeat "ticks the same whether the engine is reading rows or wedged on a credential prompt" (`AGENTS.md:634-637`). Ground truth:
   - `refresh_pbip_model.py:1354-1362` checks `_credential_state(pid)` before refresh and returns `REFRESH: CREDENTIAL_MISSING` / `REFRESH: BLOCKED_BY_DIALOG` immediately for a visible prompt/dialog.
   - `refresh_pbip_model.py:355-357` repeats the t=0 block check inside `refresh()` before starting the worker.
   - `_credential_modal.py:552-555` polls during the wait and `_raise_detection()` raises immediately on `state.modal` or `state.blocking_dialog` (`_credential_modal.py:486-491`).
   - The indeterminate path is different: `_credential_modal.py:557-562` prints `print_indeterminate_state_notice()` once for `unknown_reason`, then heartbeat can continue. That covers a hidden/unreadable/indeterminate window state, not a visible credential prompt.

   Suggested scope: say the heartbeat ticks the same for real row reads vs a wedged/indeterminate wait; do not name a visible credential prompt as the example.

2. The quoted heartbeat example uses the wrong denominator for the default scripted path. The literal format is exactly:

   ```python
   print(f"still refreshing, {int(elapsed)}s / {int(total)}s", flush=True)
   ```

   (`_credential_modal.py:458-460`), including comma and spacing. But `refresh_pbip_model.py:354` passes `total_timeout = timeout_sec + REFRESH_WALL_CLOCK_GRACE_SECONDS`, with constants `REFRESH_TIMEOUT_SECONDS = 300` and `REFRESH_WALL_CLOCK_GRACE_SECONDS = 30` (`refresh_pbip_model.py:177,184`). So the default heartbeat denominator is `330s`, not `300s`. The table's example `still refreshing, 42s / 300s` is misleading; use `... / 330s` or avoid a concrete denominator.

Verified OK: the ceiling claim on this branch still holds as written: constants are 300 and 30, and the argparse options around `refresh_pbip_model.py:1296-1337` expose no timeout override. Sync gate also passes by exit code and reports `pbi-report-builder.agent.md` at 29,941 chars; the edit is above `<!-- BEGIN:shared-conventions -->` (`AGENTS.md:701`), so it is outside the synced block.

Guust-Franssens:
APPROVE: the third-pass wording matches the source behavior I checked.

Verification:
- Diff reviewed with `git diff HEAD~1 HEAD -- AGENTS.md`.
- Sync gate: `python scripts\sync_agent_conventions.py --check` exited 0 and reported all 4 agents current; `pbi-report-builder.agent.md` is 29,941 chars. The edit is outside the synced block (`<!-- BEGIN:shared-conventions -->` is at `AGENTS.md:704`).
- Detected credential modal behavior: `refresh_pbip_model.py:1354-1362` and `refresh_pbip_model.py:355-357` abort on visible modal/blocking dialog; `_credential_modal.py:552-555` + `486-491` also raises during polling.
- Indeterminate/unknown behavior: `_credential_modal.py:557-562` prints `print_indeterminate_state_notice()` once, then `print_refresh_heartbeat()` can continue.
- Heartbeat format: `_credential_modal.py:458-460` is exactly `still refreshing, {int(elapsed)}s / {int(total)}s`, and `refresh_pbip_model.py:354` passes `timeout_sec + REFRESH_WALL_CLOCK_GRACE_SECONDS`; constants are 300 + 30 (`refresh_pbip_model.py:177,184`), so the documented `... / 330s` example is correct.
- Timeout text: `refresh_pbip_model.py:1190-1191` prints `CAUSE UNKNOWN - this script cannot distinguish these two, and they need` followed by `opposite responses:`; the prose quotes the substantive sentence correctly.

No further changes needed from me.

PR #267, opened 2026-08-21
feat(gates): census = BLANK() stub measures, split actionable vs dead end

### LINKED ISSUE #257: No systematic check for = BLANK() stub measures, and an ad-hoc grep gets it wrong
**Source:** customer field report, live estate.

## The gap

When the engine cannot translate a Tableau calculated field it emits a placeholder measure, `= BLANK()`. Hand-authoring those is the **single largest remaining body of work** on the live estate - roughly **29.3% of ~1,067 measures**. There is no reliable check for it, so it gets tracked far less rigorously than field-binding correctness, which *does* have a tool.

A disk-verified sweep found previously-invisible ratios:

| model | stubs |
|---|---|
| `ACMU_CRO_Tail_Monitoring_Dashboard` | 64/89 (**72%**) |
| `AAL_Free_Monitoring_Dashboard` | 32/50 (**64%**) |
| `IA_Airborne_Services_Dashboard` | 69/129 (**53%**) |

## The part that makes this urgent

An ad-hoc automated sweep **gave a wrong answer** - ACMU `Selected Measure` was reported still-stubbed when it was not - caught only by manual file re-verification, never by the tool.

That failure mode is predictable: a naive grep for `BLANK()` is wrong, because legitimate authored DAX frequently *contains* `BLANK()` without *being* a stub (`IF(ISBLANK([x]), BLANK(), [y])`, `DIVIDE(a, b, BLANK())`). The detector must ask **"is the entire expression nothing but `BLANK()`"**, not "does this text appear". Get it wrong and the tool is worse than nothing, because its numbers get quoted.

## Ask

A `scripts/check_stub_measures.py` analogous to `check_field_bindings.py`:

- robust detection (spacing, case, multi-line, comments) - not a substring match
- per-table and model-wide ratios, in the `64/89 (72%)` shape already in use
- **split stubs by whether the original Tableau formula survived as an annotation**: *actionable* (translatable now) vs *dead end* (must be recovered from the workbook). That split turns one scary number into a work queue plus a smaller escalation list.
- same CLI shape, output style and exit-code semantics as `check_field_bindings.py`, since agents will run them together

Companion to #236/#247.

Guust-Franssens:
Implements #257. A customer measured **29.3% of ~1,067 measures estate-wide** still `= BLANK()` placeholders, with no reliable way to check — and an ad-hoc sweep that **gave a wrong answer**, caught only by manual file re-verification.

## The detection rule, and why it cannot produce the reported false positive

A substring search for `BLANK()` is wrong: legitimate authored DAX frequently *contains* `BLANK()` without *being* a stub — `IF(ISBLANK([x]), BLANK(), [y])`, `DIVIDE(a, b, BLANK())`. The rule here asks whether a measure's **entire effective expression** is a single `BLANK()` call, across all three TMDL expression forms (inline, `=`-prefixed, and multi-line block), tolerating spacing (`BLANK ( )`), case, and comments.

The false-positive case the customer hit is a committed regression test, written before the detector.

## Numbers — real models, not fixtures

| corpus | stubs |
|---|---|
| `examples/*/fabric/` (16 models, 526 measures) | **0 / 526 (0%)** |
| a real 38-workbook estate bundle | **135 / 264 measures (51%)**, plus 97 / 857 calculated columns |

The 0% on committed examples is the false-positive control: 526 pieces of real authored DAX, zero flagged. The 51% is the problem the customer described.

**Cross-validated:** an independently written PowerShell line-scanner agreed exactly at **232** actionable stubs.

## The actionable / dead-end split — and a finding inside it

232 actionable / **0 dead end**: *every* stub still carries its `TableauFormula` annotation. Nothing is lost for want of source information — the engine preserves the original Tableau formula and then emits `BLANK()` anyway. That turns one alarming number into a translatable work queue, and it is now [reported upstream](https://github.com/Yarbrdab000/tableau-fabric-skills/issues/171) as evidence for automatic translation of parameter-driven selectors.

## Mutation testing

| pass | result |
|---|---|
| 1st (26 mutations, judged by exit code) | 22 caught / **2 genuine survivors** |
| survivors | both defensive guards, provably unreachable at verdict level |
| 2nd, after documented helper-level assertions | **26 / 26** |

Reported honestly rather than as a clean sweep — the survivors were real, and are named.

## ⚠️ Separate upstream defect found while building this

Four engine resource docs still document the stub placeholder as **`= 0`**, but the engine emits **`= BLANK()`**:

- `resources/calc-to-dax.md:263` — *"`dax is None` → emits `measure '<name>' = 0`"*
- `resources/feature-parity.md:33` and `:52`
- `resources/migration-gotchas.md:33`
- `resources/migration-orchestrator.md:114`

Ground truth: `scripts/tmdl_generate.py` `generate_measure_tmdl` — `expr = dax if dax else "BLANK()"`, with a docstring dated 2026-08-06 explaining the change. Anyone building a detector from those docs finds **zero stubs on a fully-stubbed model**. Reported upstream.

## Gates (exit codes, not printed text)

`ruff format` **0** · `ruff check --fix` **0** · `pylint scripts` **0** (10.00/10) · `pylint .github/skills/pbip-model-refresh/scripts` **0** · `pylint .github/skills/powerbi-ai-readiness/scripts` **0** · `pytest -q` **0**

CLI shape, output style and exit-code semantics deliberately mirror `check_field_bindings.py`, since agents run them together.

Fixes #257


Guust-Franssens:
Blind review verdict: APPROVE / no changes needed.

What I ran:

```powershell
cd _scratch\review-stubs
..\..\.venv\Scripts\python.exe -m pytest tests\test_check_stub_measures.py -q
# 54 passed in 3.91s

..\..\.venv\Scripts\python.exe scripts\check_stub_measures.py examples --quiet
# stub-examples-exit=0

..\..\.venv\Scripts\python.exe scripts\check_stub_measures.py examples --strict --quiet
# stub-examples-strict-exit=0
```

Mutation testing (harness classified only pytest exit 1 with real FAILED output as caught, not import/setup failures): 8/8 sampled mutants caught, 0 genuine survivors.

Caught mutants:
- substring match instead of full expression match
- case-sensitive BLANK matcher
- comment stripping disabled
- redundant outer parens not stripped
- indented expression swallows properties/annotations
- actionable split uses annotation presence instead of non-empty formula
- bundle scans engine baseline outside `pbip/`
- `--strict` ignored

Negative controls: committed `examples/` scan exits 0, and strict mode also exits 0 because there are no committed stub findings under examples. I could not verify against the private live estate named in the issue; only repo fixtures and committed examples were available.

PR #270, opened 2026-08-21
Add PBIP refresh progress liveness

### LINKED ISSUE #269: Refresh progress IS observable: AMO ProgressReportCurrent gives live row counts cross-connection, so the 300s ceiling can become a liveness check
**MEASURED, end to end.** A refresh can report live row counts headlessly, with zero new dependencies. This is the research log verbatim; the probe is `_scratch/trace-research/trace_refresh_probe.py` (backed up in the session workspace).

---

# FINDINGS — does Power BI Desktop emit `ProgressReportCurrent` during a refresh?

Research log. Written incrementally; the prior attempt was lost because everything was in memory.

Every claim is marked **MEASURED** (I ran it and pasted the output) or **INFERRED**.

## Environment (MEASURED)

- Repo: `C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration`
- Python: `.venv\Scripts\python.exe`
- AMO: 19.84.1.0 from the NuGet global-packages cache (same assemblies `refresh_pbip_model.py` loads)
- Model under test: `_scratch\trace-research\model\AirlineAllianceActivity.pbip` (scratch copy),
  whose `DataFolder` parameter points at `examples\airline-alliance-activity\data\`
  (2 CSVs, 109.28 MB each — read-only use).

## Status

- [x] Reflected the AMO trace surface (re-confirms the settled prior findings)
- [x] Live trace against a running Desktop instance — **MEASURED, works**
- [x] Cross-connection visibility (AMO trace vs ADOMD refresh) — **MEASURED, works**
- [x] Cadence — **MEASURED, ~2.1 s**
- [x] Recommendation — see §5

## ANSWER (all MEASURED, 2026-08-21, Desktop pid 24020, model AirlineAllianceActivity)

`.venv\Scripts\python.exe _scratch\trace-research\trace_refresh_probe.py --pid 24020 --mode server`
(exit 0), then `analyze_trace.py`. Raw event log: `trace-server.jsonl` (563 events).

### 1. Cross-connection visibility: YES ✅ — this is the decisive finding

The AMO **server-level** trace ran on its own connection (`Server.SessionID=7E73D197-EE0A-42FC-B9CB-54B55D8C7028`).
The refresh was issued over a **separate ADOMD connection** (`SessionID=F6A15033-FA87-4139-BA0F-412C89151511`),
exactly as `refresh_pbip_model.py` does it.

> distinct SessionIDs in captured events: `{'F6A15033-FA87-4139-BA0F-412C89151511'}`

All 563 captured events carry the **refresh's** session id, not the trace's. So a listener can watch a
refresh it did not issue. `refresh_pbip_model.py` does **not** need to change how it refreshes.

Use `Server.Traces.Add(...)` (server-level), **not** `Server.SessionTrace` — the latter is scoped to the
AMO client's own session, which is the one session that has nothing to report.

### 2. `ProgressReportCurrent` carries a real, live ROW COUNT ✅

24 class-7 events for the one table with actual volume:

```
Flight Activity: 24 events, IntegerData 10000..240000, ProgressTotal==IntegerData: True
TextData: Reading data for the 'Flight Activity' table. (TableTMID='10')
```

`IntegerData` is the running row count, reported every 10,000 rows. This is the same number Desktop's
own refresh dialog shows.

### 3. It is a COUNTER, not a PERCENTAGE ⚠️

`ProgressTotal == IntegerData` on **every** event. `ProgressTotal` is not a denominator here — for a
file/query source the engine does not know the row total in advance, so there is nothing to divide by.

**You can honestly render `"Flight Activity: 240,000 rows read"`. You cannot render `"60%"`.**
Desktop's own UI has the same limitation, which is why it shows a rising count rather than a bar.

### 4. Cadence: ~2.1 s, and only where there is volume ⚠️

```
gaps between class-7 events = [2.1, 0.0, 0.0, 2.09, 0.0, 0.0, 2.1, ...]
max gap between ANY two captured events = 2.105 s   (p95 = 0.000 s)
refresh elapsed 27.77 s; events span t_rel 4.65 .. 31.94
```

Batches are delivered in bursts (several with identical timestamps), but the wall-clock spacing is a
steady ~2.1 s. The 9 small parameter tables emitted **only** Begin/End with `IntegerData=0` — class-7
appears only for tables big enough to take time, which is exactly where progress is wanted.

### 5. Recommendation — replace the duration ceiling (#253) with a liveness check

`REFRESH_TIMEOUT_SECONDS = 300` (`refresh_pbip_model.py:377` area) is a guess about how long work
*should* take, and it is wrong in both directions: it kills a legitimate 750 s Snowflake refresh, and it
waits 300 s before noticing a refresh that died in second 3.

*"no progress event in N seconds"* is strictly better on both sides. Measured max inter-event gap on a
live refresh is **2.105 s**, so even N=60 is ~28x headroom.

Cost to adopt: **zero new dependencies** — AMO 19.84.1.0 from the NuGet cache is already loaded by
`refresh_pbip_model.py` for `ImageSave`.

### 6. Two traps that cost real time here — both are server-side, invisible to the client

- **`trace.Events` is shadowed.** `Microsoft.AnalysisServices.Trace` derives from
  `System.ComponentModel.Component`, which also declares `Events` (an `EventHandlerList`). Under
  pythonnet the base member wins and `trace.Events.Add(...)` raises
  `AttributeError: 'EventHandlerList' object has no attribute 'Add'`. Resolve the property by its
  declared TYPE (`TraceEventCollection`), not by name — see `_amo_events()`.
- **The client accepts trace columns the server rejects.** `event.Columns.Add(...)` never fails; the
  combination is validated only at `trace.Update()`, which throws
  `"The event Id=5 does not contain the column Id=4"`. `$SYSTEM.DISCOVER_TRACE_DEFINITION_PROVIDERINFO`
  does **not** help — it returns a single XML provider blob (name/version), not per-event columns.
  The probe therefore treats the server as the authority: request everything, parse the complaint, drop
  the named column, retry. Measured result, per event:

  | event | rejected by the server |
  |---|---|
  | `ProgressReportBegin` | EndTime, Duration, CpuTime, ProgressTotal, IntegerData, Severity, Success, Error, ApplicationName |
  | `ProgressReportCurrent` | EndTime, Duration, CpuTime, Severity, Success, Error, ApplicationName |
  | `ProgressReportEnd` | ApplicationName |
  | `ProgressReportError` | CpuTime, Success, ApplicationName |

  Note `ProgressReportBegin` rejects `IntegerData`/`ProgressTotal` — the row count exists **only** on
  `ProgressReportCurrent` (and `ProgressReportEnd`, which reports `0` for the parameter tables).

### 7. Caveat on this measurement

One model, one Desktop build, a 27.8 s CSV-backed refresh. Not yet shown: behaviour against a **remote**
warehouse (Snowflake), where the row batches arrive over the network, and whether cadence holds on a
multi-hundred-second refresh. Issue #262's large fixture is what would close that gap.

## 0. Re-confirmation of the settled facts (MEASURED)

`.venv\Scripts\python.exe _scratch\trace-research\reflect_amo_surface.py` (exit 0):

```
TraceEventClass: 80 values
  --- Progress* ---
    ProgressReportBegin = 5
    ProgressReportCurrent = 7
    ProgressReportEnd = 6
    ProgressReportError = 8

TraceColumn: 44 values
  ActivityID, ApplicationContext, ApplicationName, CalculationExpression, ClientHostName,
  ClientProcessID, ConnectionID, CpuTime, CurrentTime, DatabaseFriendlyName, DatabaseName, Duration,
  EndTime, Error, ErrorType, EventClass, EventSubclass, FunctionName, GraphCorrelationID, Identity,
  IntegerData, JobID, Label, NTCanonicalUserName, NTDomainName, NTUserName, ObjectID, ObjectName,
  ObjectPath, ObjectReference, ObjectType, ProgressTotal, RequestID, RequestParameters,
  RequestProperties, ServerName, SessionID, SessionType, Severity, Spid, StartTime, Success,
  TextData, UserObjectID

Server members relating to trace:
    CancelSession, SessionID, SessionTrace, Traces, get_SessionID, get_SessionTrace, get_Traces

Trace public members:
    .ctor, Annotations, Audit, AutoRestart, Clone, Container, CopyTo, CreatedTimestamp, Description,
    Dispose, Disposed, Drop, Equals, Events, Filter, GetDependents, GetHashCode, GetLifetimeService,
    GetReferences, GetType, GetUpdateOverwrites, ID, InitializeLifetimeService, IsLoaded, IsStarted,
    LastSchemaUpdate, LogFileAppend, LogFileName, LogFileRollover, LogFileSize, Name, OnEvent,
    OwningCollection, Parent, Refresh, Site, SiteID, Start, Stop, StopTime, Stopped, Submit, ...
```

**Key structural fact (MEASURED):** the AMO `Server` exposes **two** trace surfaces —
`Server.SessionTrace` (scoped to the AMO client's own session) and `Server.Traces` (a **server-level**
trace collection). That distinction is the whole cross-connection question, and the probe tests both.

Guust-Franssens:
## Summary
- enable AMO server-level refresh progress by default for CLI and direct `refresh()` callers
- classify progress events by the AMO numeric wire value derived from `TraceEventClass` (`7` for `ProgressReportCurrent`), while tolerating name-form test doubles
- emit live table row-count evidence during refresh
- make liveness non-fatal: silence now prints `[progress] no progress event ... still waiting`; only the generous absolute backstop kills a traced refresh
- supersede the old elapsed-only `still refreshing` heartbeat in traced mode with progress-aware heartbeat output
- keep `--no-progress` as the escape hatch for legacy quiet/fixed-timeout behavior
- degrade safely when tracing cannot start: refresh still runs and the reason is printed on stderr

Fixes #269

## Design choice
Chose **liveness reports, absolute backstop kills**. Evidence supports row-count progress visibility, but not fatal liveness: the original 2.105s gap measurement was local CSV only, and remote first-row latency is unmeasured. Fatal liveness can therefore false-positive the customer Snowflake case (#253). The absolute backstop fixes the original too-low 300s ceiling without introducing that regression.

### Airline fixture
Confirmed rising row-count lines on live Desktop refresh after fixing EventClass classification:

```text
[progress] Flight Activity: 10,000 rows read (4.6s)
...
[progress] Flight Activity: 210,000 rows read (21.4s)
  data   : 246236 row(s) in 'Flight Activity'
```

### 2M-row large-refresh fixture
Generated with `scripts/make_refresh_fixture.py --rows 2_000_000 --seed 262 --print-hash`:

```text
rows=2,000,000 seed=262 bytes=469,932,345 (469.9 MB)
sha256=59db5cf0522ae069f1706da12769820b621ad28e18dd83b6ef08ddad29a67c37
```

Live `refresh_pbip_model.py --pid 40032 --no-save --canaries Orders OrdersTypedAgain CustomerDateRollup CustomerStatusMerge` printed rising counts for all large tables and ended `REFRESH: DATA_OK`. Sample:

```text
[progress] OrdersTypedAgain: 10,000 rows read (5.6s)
[progress] Orders: 10,000 rows read (5.6s)
...
[progress] CustomerDateRollup: 1,860,000 rows read (77.0s)
  data   : 2000000 row(s) in 'Orders'
  data   : 2000000 row(s) in 'OrdersTypedAgain'
  data   : 1993186 row(s) in 'CustomerDateRollup'
  data   : 1597029 row(s) in 'CustomerStatusMerge'
REFRESH: DATA_OK
```

Unthrottled AMO trace against the same fixture captured 968 events: `{'5': 105, '6': 105, '7': 758}`. Measured gaps:
- max gap across all ProgressReport* events: **4.196s**
- max gap across class-7 ProgressReportCurrent events: **4.146s**
- per-object class-7 max gap: **2.107s** (`Orders`, `OrdersTypedAgain`), **2.032s** (`CustomerDateRollup`, `CustomerStatusMerge`)

Desktop PID 40032 was closed after verification.

## Validation
- `C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\.venv\Scripts\python.exe -m pytest -q` — 1784 passed, 7 skipped
- `C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\.venv\Scripts\python.exe -m pylint .github\skills\pbip-model-refresh\scripts` — 10.00/10
- Mutation tests: 8/8 caught, including:
  - exact bad comparison regression `event_class == "ProgressReportCurrent"`
  - removing enum-derived numeric wire value
  - making liveness fatal again
  - removing liveness-warning scheduling
  - reverting traced `CommandTimeout` to legacy `timeout_sec`

### DIFF
--- a/.github/skills/pbip-model-refresh/scripts/refresh_pbip_model.py
+++ b/.github/skills/pbip-model-refresh/scripts/refresh_pbip_model.py
@@ -692,6 +692,7 @@ class RefreshProgressMonitor:  # pylint: disable=too-many-instance-attributes
         throttle_seconds: float = REFRESH_PROGRESS_THROTTLE_SECONDS,
         clock=time.monotonic,
         printer=print,
+        current_event_values: set[str] | None = None,
     ) -> None:
         self.trace = trace
         self.server = server
@@ -707,6 +708,7 @@ class RefreshProgressMonitor:  # pylint: disable=too-many-instance-attributes
         self._last_print_by_object: dict[str, float] = {}
         self._latest_rows_by_object: dict[str, int] = {}
         self._last_row_object: str | None = None
+        self._current_event_values = current_event_values or {"ProgressReportCurrent"}
 
     def mark_refresh_started(self) -> None:
         """Start the liveness clock from the refresh request, not trace construction."""
@@ -725,7 +727,7 @@ class RefreshProgressMonitor:  # pylint: disable=too-many-instance-attributes
         with self._lock:
             self._last_event_at = now
             self._last_event_label = str(label)
-            if event_class == "ProgressReportCurrent" and row_count is not None:
+            if str(event_class) in self._current_event_values and row_count is not None:
                 self._latest_rows_by_object[str(label)] = row_count
                 self._last_row_object = str(label)
                 message = self._format_row_progress_locked(str(label), row_count, now)
@@ -820,6 +822,15 @@ def _load_amo_trace_types():
     return server_type, TraceColumn, TraceEventClass, TraceEvent
 
 
+def _progress_event_values(trace_event_class, event_name: str) -> set[str]:
+    """Accepted EventClass wire values for one AMO trace event."""
+    values = {event_name}
+    enum_value = _enum_int(getattr(trace_event_class, event_name))
+    if enum_value is not None:
+        values.add(str(enum_value))
+    return values
+
+
 def _add_progress_events(
     trace, trace_event_class, trace_column, trace_event_type, wanted: dict[str, list[str]]
 ) -> None:
@@ -875,7 +886,13 @@ def _start_refresh_progress_trace(port: int, liveness_seconds: float) -> Refresh
         name = f"pbip-refresh-progress-{os.getpid()}-{int(time.time())}"
         trace = server.Traces.Add(name, name)
         _negotiate_progress_trace_columns(trace, trace_event_class, trace_column, trace_event_type)
-        monitor = RefreshProgressMonitor(trace, server, trace_column, liveness_seconds=liveness_seconds)
+        monitor = RefreshProgressMonitor(
+            trace,
+            server,
+            trace_column,
+            liveness_seconds=liveness_seconds,
+            current_event_values=_progress_event_values(trace_event_class, "ProgressReportCurrent"),
+        )
         trace.OnEvent += monitor.handle_event
         setattr(trace, "_py_progress_handler", monitor.handle_event)  # noqa: B010
         trace.Start()
--- a/.github/skills/pbip-model-refresh/scripts/refresh_pbip_model.py
+++ b/.github/skills/pbip-model-refresh/scripts/refresh_pbip_model.py
@@ -168,12 +168,12 @@ PERSIST_LOCK_TIMEOUT_SECONDS = 120.0
 REFRESH_TIMEOUT_SECONDS = 300
 
 # A server-level AMO trace observes ProgressReport* events from the separate ADOMD refresh session.
-# Measured max gap during a live CSV refresh was 2.105s, so 120s is ~57x headroom while still surfacing
-# a dead refresh far earlier than the old 300s wall-clock wait.
+# Liveness is deliberately NON-FATAL: local CSV refreshes showed ~2.1s gaps, but we have no evidence
+# for remote first-row latency. Killing on this timer would false-positive a healthy slow source.
 REFRESH_PROGRESS_LIVENESS_SECONDS = 120.0
 REFRESH_PROGRESS_THROTTLE_SECONDS = 2.0
-# Backstop only: liveness is the primary bound when progress is available, but a stuck trace or engine
-# should still be finite. This must be comfortably above measured 700-750s customer Snowflake refreshes.
+# The traced-mode fatal backstop. It must be comfortably above measured 700-750s customer Snowflake
+# refreshes; liveness only reports silence before this backstop fires.
 REFRESH_ABSOLUTE_TIMEOUT_SECONDS = 3600.0
 
 # Grace on top of the XMLA ceiling before the legacy outer wall clock gives up. The XMLA layer honours
@@ -305,29 +305,6 @@ def _catalog_id(conn) -> str:
         reader.Close()
 
 
-def _raise_progress_timeout(
-    monitor: RefreshProgressMonitor,
-    desktop_pid: int | None,
-    source_hint: str | None,
-    latched_unknown: str | None,
-    latched_desktop_unready: str | None,
-) -> None:
-    """Raise the right terminal verdict when progress liveness expires."""
-    if desktop_pid is not None:
-        state = _credential_state(desktop_pid)
-        _raise_if_blocked(desktop_pid, state, source_hint)
-        latched_desktop_unready = latched_desktop_unready or state.desktop_unready
-        if latched_desktop_unready is not None:
-            raise DesktopUnreadyError(desktop_pid, latched_desktop_unready)
-        latched_unknown = latched_unknown or state.unknown_reason
-        if latched_unknown is not None:
-            raise CredentialUnknownError(desktop_pid, latched_unknown)
-    raise TimeoutError(
-        f"no refresh progress event for {monitor.seconds_since_last_event():.1f}s "
-        f"(liveness window {monitor.liveness_seconds:.1f}s; last event: {monitor.last_event_label()})"
-    )
-
-
 # pylint: disable=too-many-arguments,too-many-branches,too-many-locals,too-many-statements
 def _join_refresh_worker(
     worker: threading.Thread,
@@ -361,20 +338,12 @@ def _join_refresh_worker(
     while worker.is_alive():
         elapsed = time.monotonic() - started
         absolute_remaining = max(0.0, total_timeout - elapsed)
-        liveness_remaining = progress_monitor.liveness_seconds - progress_monitor.seconds_since_last_event()
         if absolute_remaining <= 0:
             break
-        if liveness_remaining <= 0:
-            _raise_progress_timeout(
-                progress_monitor,
-                desktop_pid,
-                source_hint,
-                latched_unknown,
-                latched_desktop_unready,
-            )
+        progress_monitor.print_liveness_warning_if_due()
         wait_for = min(
             absolute_remaining,
-            max(0.0, liveness_remaining),
+            progress_monitor.seconds_until_liveness_warning(),
             REFRESH_CREDENTIAL_POLL_SECONDS,
             max(0.0, next_heartbeat - elapsed),
         )
@@ -392,14 +361,6 @@ def _join_refresh_worker(
             progress_monitor.print_evidence_heartbeat(elapsed, total_timeout)
             next_heartbeat += REFRESH_HEARTBEAT_SECONDS
     if worker.is_alive():
-        if progress_monitor.seconds_since_last_event() >= progress_monitor.liveness_seconds:
-            _raise_progress_timeout(
-                progress_monitor,
-                desktop_pid,
-                source_hint,
-                latched_unknown,
-                latched_desktop_unready,
-            )
         return False
     return True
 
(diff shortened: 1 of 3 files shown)


Guust-Franssens:
REQUEST CHANGES: default progress liveness can abort a healthy refresh that would have completed inside the legacy timeout.

This is not the happy path with row-count events; it is the failure path where the trace starts but no ProgressReport event arrives during a legitimate silent phase (for example, a remote query before the first row-count event). The new default treats that as stuck even if XMLA returns before the old wall-clock ceiling.

Reproduction, from this PR worktree:

```powershell
cd C:\Users\gfranssens\vscode-projects\t2p-progress
C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\.venv\Scripts\python.exe _scratch\review-270\silent_healthy_probe.py
```

Probe contents: `_start_refresh_progress_trace` returns a real `RefreshProgressMonitor(liveness_seconds=0.10)`, and the fake ADOMD command returns successfully after 0.25s. That is after the liveness window but well inside `timeout_sec=1`, i.e. a refresh the legacy path would have survived.

Actual output:

```text
[progress] enabled: no progress event for 0.1s stops the refresh (absolute backstop 5s)
TimeoutError no refresh progress event for 0.1s (liveness window 0.1s; last event: refresh start)
elapsed 0.105
worker would have completed within legacy timeout_sec=1
```

That means the failure boundary moved from “refresh exceeds the old timeout” to “trace is silent longer than the progress window,” even when the refresh itself is healthy and completes under the legacy ceiling. Given the issue context that small/parameter tables may emit only Begin/End and a long remote query may be silent before first rows, this default can now kill legitimate refreshes. Consider keeping the legacy wall-clock as the default bound until at least one real progress event is observed, or making liveness apply only after evidence that the trace is emitting useful refresh events for this command.

Guust-Franssens:
Final blind review verdict: REQUEST CHANGES.

Blocking finding: default-on progress liveness can abort a healthy refresh that would have completed inside the legacy timeout. I posted the focused reproduction above: a traced refresh with no progress events returns successfully after 0.25s, but liveness=0.10s aborts at 0.105s even though legacy `timeout_sec=1` would have survived. This is exactly the risk class for legitimate silent phases before a first row-count event.

Commands / outputs I ran:

```powershell
C:\Users\gfranssens\vscode-projects\tableau-to-powerbi-migration\.venv\Scripts\python.exe .github\skills\pbip-model-refresh\tests\test_refresh_wall_clock.py -q
# (typo avoided in actual run; command used was python -m pytest)
C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\.venv\Scripts\python.exe -m pytest .github\skills\pbip-model-refresh\tests\test_refresh_wall_clock.py -q
# 12 passed in 28.15s
```

Real trace ground truth (`C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\_scratch\trace-research\trace-server.jsonl`):

```text
total 563
event_class_counts {'5': 270, '6': 269, '7': 24}
```

So the code/test assumption that `ProgressReportCurrent` arrives as numeric wire value `"7"` is correct; no magic-number hardcode in production (`_progress_event_values` derives it from the enum).

Mutation testing sample:

```text
MUTANT name_only_event_values: caught
MUTANT fallback_uses_absolute_timeout: survived
MUTANT trace_failure_no_degrade: caught
```

The survivor is a test gap: the suite does not catch replacing the traced path's `CommandTimeout=absolute_timeout_sec` with the legacy `timeout_sec`. That is not my blocking finding, but it means the mutation score is not perfect for the timeout semantics.

Safe degradation / cleanup probes:

```text
# trace setup failure test in suite passed and asserts stderr fallback.
# cleanup probe output:
worker_error ValueError boom
worker_error stop 1 drop 1 disconnect 1
liveness_timeout TimeoutError no refresh progress event for 0.1s (liveness window 0.1s; last event: refresh start)
liveness_timeout stop 1 drop 1 disconnect 1
```

Thread-safety review: shared progress state is lock-protected; printing happens after releasing the lock, so I did not find a race or lock-held printing problem.

Absolute backstop: with progress trace available, a no-event modal is bounded by progress liveness (default 120s); if trace setup fails, the legacy 300s + 30s grace path remains. If progress events keep arriving for a blocked refresh, the absolute bound is 3600s.

Could not verify: I did not open Desktop or test a live remote-query silent phase. The false positive is a controlled fake proving the new default can kill a healthy no-event refresh inside the old timeout; the real-world plausibility comes from the stated measurement that some phases/tables may not emit `ProgressReportCurrent` before completion/first rows.

PR #273, opened 2026-08-21
fix(probe): hand the custom-SQL probe to the operator instead of running the full query

### LINKED ISSUE #244: probe_live_source: custom-SQL canary disables query folding, so the ~30s probe runs the FULL query
## Summary

`probe_live_source.py` builds its one-row canary for a **custom-SQL** source using M shapes in which
**query folding is disabled**, so `Table.FirstN(..., 1)` is applied *locally* by the mashup engine
rather than pushed down as `LIMIT 1`. The warehouse therefore executes the **entire** custom SQL.

This matters because the canary's whole justification is that it is cheap — "a ~30-second probe
replaces a 45-minute build you would have to throw away" (`powerbi-semantic-model-gotchas` §5). For a
custom-SQL source that claim does not hold.

## Verified — code

`scripts/probe_live_source.py`, `build_m_query()`:

| connector | emitted M | folding |
|---|---|---|
| Databricks (L246) | `Table.FirstN(Value.NativeQuery(db, "<sql>"), 1)` | ❌ `EnableFolding` omitted |
| Snowflake (L294) | `Table.FirstN(Value.NativeQuery(db, "<sql>"), 1)` | ❌ `EnableFolding` omitted |
| SQL Server (L261) | `Sql.Database(server, db, [Query="<sql>"])` then `Table.FirstN` | ❌ `[Query=]` always blocks |

## Verified — documentation

- [Query folding on native queries](https://learn.microsoft.com/en-us/power-query/native-query-folding) — folding stops at a native query unless `[EnableFolding=true]` is passed as `Value.NativeQuery`'s 4th argument; default is **off**.
- Chris Webb, [Query Folding On SQL Queries Using Value.NativeQuery And EnableFolding=true](https://blog.crossjoin.co.uk/2021/02/21/query-folding-on-sql-queries-in-power-query-using-value-nativequery-and-enablefoldingtrue/) — and `Sql.Database(..., [Query=...])` blocks folding for **all** subsequent steps.

## Verified — the engine does it correctly, and we do not

The deterministic engine emits the **folding-enabled** form for the real model:

- `skills/tableau-migration/scripts/storage_mode.py:114` — `Source = Value.NativeQuery(Database, "<sql>", null, [EnableFolding=true])`
- asserted by its own tests: `test_emit_custom_sql_uses_native_query_with_folding`, `test_snowflake_custom_sql_is_deploy_ready_not_a_scaffold`, `test_emit_databricks_custom_sql_drills_catalog_then_native_query` — all assert `[EnableFolding=true]`.

So **our hand-written probe is less foldable than the artifact it exists to de-risk** — which also
undercuts the design principle `probe_bundle.py` states explicitly: the probe should exercise *the
same M* the real model uses. (Note `probe_bundle.py` is **not** affected: it derives from the emitted
bundle and therefore inherits `[EnableFolding=true]`. That asymmetry is itself an argument for the
derived-probe approach.)

## Inferred — needs a query-history measurement to confirm

That the warehouse executes the full query. Strongly implied by the documented behaviour, but
**not measured**, and `probe_bundle.py`'s own docstring already sets the right standard: *"folding is
documented behaviour, not something this script can verify offline. Confirm in the source system's
query history."*

Nuance worth keeping: Power Query streams, so `Table.FirstN` may stop *reading* early — but Snowflake
still **computes** the result set, and compute is the expensive part. On the airline estate the
benchmarks were `Tail_Query` = 1,020,016 rows / **452 s** and `Flight_Level_Query` ≈ 1.96 M rows /
**~700 s**. A probe costing 452 s is not the "~30 s" the skill promises.

⚠️ I could not confirm from this repo that the airline estate's relations are `custom_sql` — every
tracked `migration-spec.json` here is a file-based public example. But `build_m_query`'s own docstring
names `Flight_Level_Query` as the motivating case (L221-223), so the path is real and is that estate's.

## Existing partial mitigation

`probe_live_source.py:487` sorts custom-SQL tables **last**, so a plain table is preferred when the
source has one. That does not help a datasource whose relations are *all* custom SQL.

## Proposed fix

1. Pass the options record on both `Value.NativeQuery` calls:
   `Value.NativeQuery(db, "<sql>", null, [EnableFolding=true])`
2. Switch the SQL Server branch off `[Query=]` to
   `Value.NativeQuery(Sql.Database(server, db), "<sql>", null, [EnableFolding=true])`
3. Correct the cost claim in `powerbi-semantic-model-gotchas` §5 — a custom-SQL probe is only cheap
   **if** the LIMIT folds, and folding is possible-not-guaranteed (SQL with `ORDER BY`, CTEs or set
   operations may not wrap).
4. Have the probe **report** which shape it used, so a slow probe is diagnosable rather than mysterious.

## Definition of done

- [ ] Both `Value.NativeQuery` call sites pass `[EnableFolding=true]`; SQL Server no longer uses `[Query=]`
- [ ] A test asserts the emitted probe M carries `[EnableFolding=true]` for each connector (mirrors the engine's own tests)
- [ ] §5's cost claim distinguishes a table probe from a custom-SQL probe
- [ ] Measured against a real warehouse's query history: confirm a `LIMIT 1` reaches the source

Guust-Franssens:
Fixes #244.

For a custom-SQL live source the probe no longer runs automatically. It builds the same throwaway probe PBIP and stops with a new terminal verdict `PROBE: OPERATOR_REQUIRED <path>`, **exit 4** (distinct non-zero: unproven, operator action required).

**Two defects on the same two lines, both removed:**

1. **Not cheap.** Folding was off, so `Table.FirstN(Value.NativeQuery(db, sql), 1)` made the warehouse run the ENTIRE custom query. The probe's whole justification is that it is a ~30s alternative to a 45-minute build.
2. **Not decisive.** The module's own docstring rejects `Value.NativeQuery` because Desktop raises a native-query approval modal that *'would hang and land on a false NO_CREDENTIAL'* — and the custom-SQL path used one anyway, unmitigated.

**The trap this avoids:** the operator instruction names **Power BI Desktop** and explicitly rules out DBeaver/Snowsight/SSMS, with the measured reason — a SQL client authenticates as the user, while Power BI uses the per-Windows-user Desktop DPAPI store. Measured 2026-08-01: the `databricks` CLI queried a warehouse happily while Power BI had never authenticated to it. Without that line, *'it runs fine in my SQL client'* becomes a false green laundered through a human.

Deliberately **not** built: a `FROM`-clause parser. Rejected on the issue with measured evidence — the one custom-SQL example in the corpus has two `FROM`s, aliases and a `UNION` in the *easy* case.

The self-contradicting docstring is fixed in the same change, since that contradiction is how this stayed invisible.

Ordinary-table path pinned unchanged by test. Mutation 5/5. CI green.


Guust-Franssens:
REQUEST CHANGES: mixed ordinary+custom-SQL live sources can lift the gate without any operator handoff for the custom query.

Reproduction from this PR worktree:

```powershell
cd C:\Users\gfranssens\vscode-projects\t2p-probe
C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\.venv\Scripts\python.exe _scratch\review-273\mixed_custom_probe.py
```

`mixed_custom_probe.py` builds one live source with two table specs on the same connection:

```python
[
  {"name": "REAL_TABLE"},
  {"name": "Custom_Query", "custom_sql": "SELECT * FROM huge.fact"},
]
```

and monkeypatches `_probe_one_table` so ordinary tables return `DATA_OK` and custom SQL would return `OPERATOR_REQUIRED` if reached.

Actual output:

```text
rc 0 verdict DATA_OK
tables_attempted [{'name': 'REAL_TABLE'}]
audit_exists False
```

So `_probe_one()` sorts ordinary tables before custom SQL, accepts the first `DATA_OK`, and never reaches the custom-SQL table. The gate can then lift in `run_probe()` even though a migrated model from the same source may still contain a `Value.NativeQuery` custom-SQL partition that was never operator-approved and may raise the native-query modal / run the full customer query.

This moves the boundary from “custom-SQL live source requires operator” to “custom-SQL requires operator only when no ordinary table on that source can prove credentials.” If the intended guarantee is merely connection-level reachability, this may be deliberate, but it contradicts the new OPERATOR_REQUIRED rationale: the custom query itself is the cost/modal risk, not just the credential path.

Guust-Franssens:
Final blind review verdict: REQUEST CHANGES.

Blocking finding already posted above: a mixed live source with one ordinary table and one custom-SQL relation can return `DATA_OK` after probing only the ordinary table. The custom-SQL PBIP/operator handoff is never reached, so the gate can lift while the migrated model may still contain an unapproved `Value.NativeQuery`/custom query partition.

Reproduction:

```powershell
cd C:\Users\gfranssens\vscode-projects\t2p-probe
C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\.venv\Scripts\python.exe _scratch\review-273\mixed_custom_probe.py
```

Output:

```text
rc 0 verdict DATA_OK
tables_attempted [{'name': 'REAL_TABLE'}]
audit_exists False
```

Other checks:

```powershell
C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\.venv\Scripts\python.exe -m pytest tests\test_probe_live_custom_sql.py tests\test_credential_gate.py -q
# 84 passed in 48.18s
```

Credential gate semantics are OK:

```text
apply_block 0
marker_exists True
clear_was_earned_after_operator None
status_after_operator 1
marker_exists_after_lift False
clear_was_earned_after_data_ok probe-cleared
status_after_data_ok 0
audit_actions ['block', 'probe-operator_required', 'probe-cleared']
```

Exit-code callers: I found no automated Python caller that invokes `probe_live_source.py` and branches on `returncode`. `run_estate.py` does not call it; `credential_gate.py` only points the operator at it; `tableau-migrator.agent.md` is the real consumer and now treats `OPERATOR_REQUIRED` as a hard stop. Docs/README mention the new verdict.

Mutation sample:

```text
MUTANT custom_sql_opens_desktop caught
MUTANT custom_sql_exit_zero caught
MUTANT custom_sql_wraps_firstn survived
MUTANT real_table_uses_native_query caught
MUTANT sort_custom_first caught
```

The survivor is a coverage gap: the custom-SQL scaffold assertion covers Snowflake but not the Databricks `Value.NativeQuery` branch. Runtime check showed current code is OK for Databricks/Snowflake, but the test would not catch a Databricks-only regression.

Docstring consistency: the old “why not native query” contradiction is gone. One nuance: SQL Server custom SQL is scaffolded as `Sql.Database(..., [Query=...])` plus `Table.FirstN(Source, 1)`, not `Value.NativeQuery`; I do not consider that a blocker because the script still stops before automatic refresh.

Persona size note (non-blocking): in this worktree I measure `tableau-migrator.agent.md` at 29,636 chars (364 headroom), not 29,998. I do not see an obviously safe trim in the changed Workflow text; if the generated persona is near the cap elsewhere, the better lever is trimming the shared AGENTS block so all personas shrink together.

Guust-Franssens:
Implemented the mixed-source semantics deliberately: an ordinary table may prove the shared credential path, but it cannot certify later custom-SQL relations. _probe_one now returns DATA_OK only when no later custom-SQL candidates remain; otherwise it logs that credentials were proven and returns the existing OPERATOR_REQUIRED/exit 4 after writing the custom-SQL probe PBIP for Desktop operator refresh (scripts/probe_live_source.py). This keeps the useful credential fact while preserving the rule that exit 0 means fully proven. Added regression coverage for mixed ordinary+custom-SQL sources and Databricks custom-SQL scaffolding (	ests/test_probe_live_custom_sql.py).

Guust-Franssens:
Re-review verdict: APPROVE / no further changes needed.

I re-reviewed only commit `1c751f2` and reran the original mixed-source reproduction plus look-ahead edge cases.

Validation:

```powershell
C:\Users\gfranssens\vscode-projects\tableau-to-pbi-migration\.venv\Scripts\python.exe -m pytest tests\test_probe_live_custom_sql.py tests\test_credential_gate.py -q
# 86 passed in 50.48s
```

Look-ahead probes:

```text
custom_before_real_unsorted rc 4 verdict OPERATOR_REQUIRED attempted ['REAL', 'Q1']
custom_before_real_unsorted ordered [('REAL', None), ('Q1', 'SELECT 1'), ('Q2', 'SELECT 2')]
multiple_after_real rc 4 verdict OPERATOR_REQUIRED attempted ['REAL', 'Q1']
multiple_after_real ordered [('REAL', None), ('Q1', 'SELECT 1'), ('Q2', 'SELECT 2')]
empty_custom_sql_falsy rc 0 verdict DATA_OK attempted ['REAL']
empty_custom_sql_falsy ordered [('REAL', None), ('Q_EMPTY', '')]
missing_name_filtered rc 0 verdict DATA_OK attempted ['REAL']
missing_name_filtered ordered [('REAL', None)]
```

Interpretation: sorting makes earlier custom-SQL relations move after ordinary tables, so `tables[i+1:]` is complete for truthy `custom_sql`. Falsy `custom_sql` has no query text to hand off; nameless tables are already filtered out and cannot be scaffolded.

Multiple custom-SQL relations: the code probes/hands off the first but logs the count. I consider this acceptable for the gate: exit 4 prevents `DATA_OK` certification, and the operator is explicitly told more custom-SQL relations remain. Scaffolding every custom query could be a useful future enhancement, but I do not think it blocks this fix.

Ordinary-only path still lifts the gate:

```text
probing 1 live source(s): [0]
rc 0
called [('probe', 'runprobe', 0, 77, False), ('lift', 'runprobe', '1 live source(s)')]
PROBE: DATA_OK all 1 live source(s) reachable
```

Databricks scaffold mutation tests (closing the prior coverage gap):

```text
MUTANT databricks_wraps_firstn caught
MUTANT databricks_navigates_schema caught
```

Exit-code callers: repeated exact grep found no automated Python caller branching on `probe_live_source.py` return codes. The real consumer remains the `tableau-migrator` instructions, which treat `OPERATOR_REQUIRED` as a hard stop; `credential_gate.py` only points operators at the probe; `run_estate.py` does not call it.

No Desktop instances were opened.

PR #362, opened 2026-08-28
feat: capture and re-apply the Tableau trial estate, and seed its empty fixtures

Guust-Franssens:
## Why

The Tableau trial site is hand-built and **nothing reproduced it**. The pipeline only ever reads
FROM Tableau and writes TO Fabric (`deploy_estate.py` targets Fabric; there were zero
`tableauserverclient` publish calls anywhere). If the trial lapsed, 35 projects, the live
Snowflake/Databricks published datasources and the RLS entitlement groups were gone -- and every
live-source test lost its subject.

## What

| command | does |
|---|---|
| `capture` | enumerate the site, download every asset, write a re-appliable manifest |
| `apply` | recreate projects (parents first), groups, then **datasources before workbooks** |
| `seed` | give every empty leaf project something migratable |

Ordering is not cosmetic: a workbook published before its datasource silently rebinds to nothing,
which is exactly the empty-report failure the migration pipeline spends effort detecting downstream.

## The seeding is the interesting half

The entire `ZZ Deep` chain (`L1`..`L11`) and **every** deliberately hostile project name were EMPTY.
Nothing ever migrated out of them, so the path handling they were built to stress was never
exercised. 10 leaves seeded with a 1.3 KB self-contained `.twbx`. `Empty On Purpose` is left alone --
its emptiness IS the fixture.

## Three bugs the fixtures caught in this code, while writing it

Worth reading, because each is the fixture doing its job on the author:

1. A project literally named `R/D` collided with `/` as a path separator -- manifest identity was
   ambiguous and the seed came out labelled `Seed - D`. Paths are now **segment lists**, never
   joined strings.
2. That same name became a *zip entry path* inside the `.twbx`, creating a directory in the archive.
   Tableau rejected it with a bare "unexpected error occurred opening the packaged workbook".
3. Filenames were **guessed** rather than read from what TSC returns. It appends its own suffix
   (`X.twbx.twbx`) and returns `.twb` for unpackaged workbooks, so every recorded filename was wrong
   and `apply` reported 45 missing assets.

Only #3 was found by a smoke test I nearly skipped; the other two were found by the hostile fixtures.

## Verification

- `apply --dry-run` against the live site: **exit 0**, every project/group/datasource/workbook
  resolves, including `R/D` and the 12-deep `L11`.
- Seed workbook: Tableau Cloud publishes it and renders **both** views; our own parser reads it with
  **0 limitations**.
- Empty-leaf count went 12 -> 2, and both survivors are correct (`Empty On Purpose`, and Tableau's
  own `External Assets Default Project`).
- ruff clean - pylint **10.00/10** exit 0 - `test_repo_layout` + `test_skills` 66 passed - privacy
  gate exit 0.

## Deliberately NOT done

Permission rules and group **membership**. User LUIDs differ on a fresh site, and a half-correct
permission model is more dangerous than an obviously absent one. `apply` prints this on completion
rather than letting you assume it was handled.

## Not in git

The manifest and the 52 downloaded assets stay gitignored under `_runs/` -- this is a public repo and
they carry customer-shaped names.


.env.example
```diff
@@ -24,6 +24,16 @@ TABLEAU_PAT_SECRET=
 # Optional overrides (defaults shown are what the scripts fall back to if unset).
 # TABLEAU_REST_API_VERSION=3.21
 
+# --- Published-datasource credentials (read by scripts/provision_tableau_estate.py) ---
```
Guust-Franssens: shouldnt this be either in a different file or all the way at the bottom clearly lsited ast optional and not necessary for a "customer scenario"? this is just for our trial runs

Guust-Franssens (commented):
seems okay now

Issue #246, opened 2026-08-20
Nothing compares a bundle's engine-output-receipt against the installed engine, so an estate can silently run 119 versions stale


Guust-Franssens:
Reported by the migration team, 2026-08-20 - found **by accident**, checking the receipt for an unrelated reason.

## Measured by the reporter
`_bundle/engine-output-receipt.json` records `engine.version: "2.141.0"`, while `scripts/engine_source.py` and `preflight.ps1 -CheckUpstream` both confirm **2.260.0** is installed and current. That is ~119 versions - the entire #128-#162 upstream fix batch - never picked up, including #162's Snowflake-stub fix.

They validated the delta is real rather than cosmetic: re-running just the 2 affected source `.twb` files against 2.260.0 turned both previously-stubbed workbooks (WCL Performance Dashboard, Aircraft Attributes Dashboard; 4 tables total) into real `Value.NativeQuery` M partitions.

## Verified in this repo
`grep` for `engine-output-receipt|engine\.version|receipt` across `scripts/` and `.github/` hits `credential_gate.py`, `engine_source.py`, `migration_bundle.py`, `probe_bundle.py`, `run_estate.py` and `scripts/README.md` - and **not `preflight.ps1`**. Nothing anywhere compares a bundle's receipt against the installed engine.

## Why this is worth fixing rather than remembering
`AGENTS.md` already states the receipt exists so an artifact answers *"what built me?"* on its own, months later, without the machine that built it. That promise is currently half-kept: **the receipt records the answer, and nothing ever reads it back.** A stale bundle is byte-identical to a fresh one until someone goes digging.

`preflight.ps1` already does exactly this shape of check for CLI drift - installed-versus-known-good. This is the same check one level out: bundle-versus-installed.

## Definition of Done
1. A check that walks bundles, reads `engine.version` from each `engine-output-receipt.json`, and compares against the canonical engine resolved via `engine_source.engine_root()`.
2. **WARN, do not block.** Consistent with CLI drift, and required by the timing rule - upgrading mid-migration is forbidden, so a blocking check would be unactionable exactly when it fires.
3. Name the bundle and both versions, so the operator can decide.
4. Surfaced from `preflight.ps1`, because that is the thing people actually run.
5. A test with a synthetic bundle whose receipt is deliberately behind.

Guust-Franssens:
Heads-up for whoever picks this up: **#250 needs the same read-side helper.**

[...]

PR #383, opened 2026-08-29
Enforce the destructive-re-run barrier before the engine runs (fixes #250)

### LINKED ISSUE #250: run_estate.py claims to own the destructive-re-run ordering but never checks it: --approved-dax deletes downstream work with no --force and no pre-check
## What

`run_estate.py`'s module docstring documents a destructive-re-run barrier and then claims:

> Hence: all DAX lands in ONE run, and per-workbook work starts only afterwards.
> **This script owns that ordering so no agent has to remember it.**

It does not own it. It *documents* it. Nothing in the code checks whether the target bundle already
contains downstream work before handing the engine a landing re-run that deletes it.

## Verified against the live engine (2.260.0), not repeated from the docstring

The docstring warns its line numbers drift, so I re-derived them by symbol on the installed canonical
plugin (`engine_source.engine_root()` -> `.../tableau-migration/scripts/migrate_estate.py`, 6278 lines).

**1. The stale-output guard exists and works - but exempts exactly the destructive path.**

`migrate_estate.py:6198-6206`:

```python
prior_report = os.path.join(args.output, "report.json")
landing_rerun = bool(args.approved_dax or second_compile)
if not args.force and not landing_rerun and os.path.isfile(prior_report):
    print(f"[STOP] Refusing to build: ... already exists ...")
    return 2
```

So a **plain** re-run into a populated bundle is refused (good). A **`--approved-dax`** re-run is
`landing_rerun`, skips the guard, and needs no `--force`. The comment above it also names `--author`
as exempt. The most destructive path is the one with the fewest safety rails.

**2. The `rmtree` sites are intact** - at 857, 915, 941, 3888, 4181 (drifted from the docstring's
879/3035/3284, as it predicted; behaviour unchanged). `.SemanticModel`, the `.pbip` project dir and
`<name>.Report` are deleted and rewritten.

**3. Our own gates are all post-hoc.** In `run_estate.py:main()`, the engine runs at **phase 1**
(`run_engine_phase`, ~L583). `find_approval_collisions`, `check_definition_of_done`,
`check_pbir_validity`, `check_empty_models` all run at **phase 2** (~L605+), reading the report the
run just produced. Even the collision check - the one gate specifically about `--approved-dax` -
reports a wrong-DAX landing *after* it has landed.

## Why this is urgent rather than theoretical

Reported from a live 38-workbook estate migration:

- their bundle was built with engine **2.141.0**; **2.260.0** has been installed the whole time
- they have ~20 items of hand-authored fix work in that bundle (case-only binding repairs, measures)
- **29.3% of ~1,067 measures are still `BLANK()` placeholders** - bulk measure authoring is their
  single largest remaining task
- the natural way to land bulk approved DAX is `--approved-dax`

That is precisely the input that walks into this: a landing re-run into a bundle full of downstream
work, no `--force`, no prompt, no pre-check, everything gone. The ordering rule the docstring states
("per-workbook work starts only afterwards") was already broken for them by circumstance, not by
error - the engine moved 119 versions under a bundle that was mid-flight.

## Proposed fix

A **pre-engine** guard in `run_estate.py`, before `run_engine_phase`:

when `--approved-dax` (or any landing re-run) targets an `--output` that already contains downstream
modifications, refuse with a distinct exit code unless explicitly acknowledged
(`--accept-bundle-rewrite`), and name the files that would be destroyed.

**The detection mechanism already exists and is already written to disk.** `engine-output-receipt.json`
records path/size/**sha256** for every native engine artifact. Re-hash them: any file whose digest no
longer matches the receipt is downstream work. Anything under `<bundle>/pbip/` not in the receipt is
downstream work.

Note this is the *same neglected artifact* as #246 - the receipt is written by every run and read back
by nothing. Two independent high-severity gaps fall out of that one omission, so the read-side helper
should be shared rather than written twice.

## Definition of done

1. A landing re-run into a bundle with downstream modifications **refuses by default**, non-zero, and
   lists what would be destroyed. Judged by exit code.
2. A landing re-run into a **pristine** bundle (hashes all match the receipt) still proceeds - this
   must not break the documented one-run landing flow.
3. `--accept-bundle-rewrite` (or equivalent) proceeds, and the acknowledgement is recorded in the
   bundle so the artifact says it happened.
4. A bundle with **no receipt** (pre-receipt, or third-party) must not hard-fail - warn and proceed,
   or it breaks every older bundle.
5. Mutation-test the tests: corrupt a hash, add an unexpected `pbip/` file, delete the receipt, and
   confirm a test fails for each.
6. Fix the docstring's ownership claim - either it is enforced, or the sentence says "documents".

Guust-Franssens:
## What

`run_estate.py`'s module docstring claimed **"this script owns that ordering so no agent has to
remember it"**. It owned nothing — it *documented* the ordering and checked nothing.

Every gate in the file runs in **phase 2**, reading output the engine has already written. That
includes `find_approval_collisions`, the one gate specifically about `--approved-dax`: it reported a
wrong-DAX landing *after* it had landed. A landing re-run into a bundle full of hand-authored fix
work was not reported at all — the engine's own stale-output guard **exempts** `--approved-dax`, so
the most destructive path was the one needing no `--force`.

`assess_bundle_rewrite` now runs **before** `run_engine_phase` and makes the claim true.

## The guard

New `EXIT_BUNDLE_REWRITE = 9`, default-on, refusing two conditions with **two separate** opt-outs:

| condition | detection | opt-out |
|---|---|---|
| downstream work would be destroyed | re-hash the bundle against the two baselines the previous run already wrote — `engine-output-receipt.json` and `input_manifest.json`'s `generated_artifacts` | `--accept-bundle-rewrite` |
| a **different engine version** built this bundle | `receipt.engine.version` vs the version resolved for *this* run | `--accept-engine-version-change` |

Both baselines are already written by every run and, until now, **read back by nothing**. The
receipt alone is not enough: it only records `ARTIFACT_SUFFIXES`, so hand-edited PBIR JSON — the
report builder's entire output — is invisible to it. The generated-artifact manifest covers that,
and is mtime-filtered when written, so it is used only for files it already knows about
(modified/missing) and never to decide a file is *new*. `added` is decided from the receipt alone,
via `engine_artifact_records` — the very function that wrote it — so a pristine bundle reconciles
exactly.

**Why two flags and not one:** a single acknowledgement would mean an operator who knows the engine
moved *also* silently waives the destruction guard on work they did not know was there. That is
moving the failure boundary rather than removing it. Two tests pin the separation in both
directions.

Acknowledging appends to `<bundle>/bundle-rewrite-acknowledgement.json` — when, which flag, and
every file destroyed — written before the engine runs, at the bundle root the `rmtree` sites do not
touch, so it survives the rewrite it describes.

**Deliberately not blocked:** a bundle with neither baseline warns and proceeds (a missing receipt
must not brick every older bundle); an empty `artifacts` list is treated as *absence of evidence*
rather than as either a clean bundle or a dirty one; `.pbi` sidecars are excluded on **both** sides
of every comparison so a Desktop refresh is not tampering; and **`--slice-only` never runs the guard
at all**, because it never runs the engine and therefore destroys nothing.

## Verified, by exit code

Real CLI runs against a synthetic bundle, engine 2.339.0 (the installed canonical plugin):

```
run_estate.py --input … --output … --approved-dax … --dry-run          -> exit 9  (names the file)
… --accept-bundle-rewrite --dry-run   (receipt says 2.141.0)           -> exit 9  (version half still binds)
… --accept-bundle-rewrite --accept-engine-version-change --dry-run     -> exit 0  (no acknowledgement written)
run_estate.py --slice-only --output <same modified bundle>             -> exit 0  (ESTATE: READY)
```

Gates: `ruff format`/`ruff check` **0**; `.\.venv\Scripts\python.exe -m pylint scripts` **10.00/10,
exit 0**; `pytest tests/test_run_estate.py` **47 passed, exit 0**; full suite **2360 passed, 3
failed** — the three failures are `tests/test_upstream_repro_pins.py`, reproduced identically on an
unmodified `master` (`git stash`, same three, exit 1) and unrelated to this change.
`scripts/run_estate.py` is **1133 lines** against `max-module-lines = 1200`.

## Mutation-tested

17 mutations applied to `scripts/run_estate.py`, each run against the suite, **all 17 caught**, then
the file restored and the baseline re-run green. Highlights (mutation → test that failed):

- guard never blocks / guard moved after the engine → 9 and 10 tests respectively, including every
  `calls == []` assertion, which is what proves the refusal is *pre*-engine rather than post-crater
- wider generated baseline dropped → `test_a_hand_edited_report_file_is_work_the_receipt_alone_cannot_see`
- `added`/`missing` detection removed → the two tests written for each
- guard blocks whenever a baseline exists (drift-insensitive) → the pristine-re-run, same-version and
  Desktop-sidecar tests, i.e. the **false-positive** guards bite
- the two acknowledgements collapsed into one (both halves) → the two separation tests
- `--slice-only` no longer exempt → `test_slice_only_still_works_against_a_bundle_full_of_downstream_work`
  *and* the pre-existing `test_slice_only_backfill_never_overwrites_an_existing_baseline`
- `.pbi` exclusion removed → the Desktop-sidecar test
- empty attestation compared naively → `test_a_receipt_that_attests_to_nothing_does_not_invent_a_finding`

One mutation was initially scored as caught for the wrong reason (it introduced a `NameError`, so
pytest failed on an error rather than on the assertion). It was rewritten to a semantically valid
form and re-run.

## Engine-version half

The receipt/engine-version guard is a **second, closely related gap in the same file** — the same
neglected artifact and the same pre-engine moment — folded in here rather than split across two
passes over one function. It is **not covered by an existing issue**; raise one if you would rather
it were tracked separately. `scripts/check_engine_receipts.py` (from #246) stays as-is: it is an
advisory, repo-tree-wide preflight sweep against the *installed* engine, whereas this is a binding,
bundle-scoped check against the engine resolved for *this run* (so it also honours `--engine`).

## Docs

`docs/operator-runbook.md` gains the exit-9 row and column (`tests/test_runbook_exit_codes.py` ties
both tables to the `EXIT_*` constants), the acknowledgement file in the bundle-contents table, and
the `--approved-dax` warning block now says the ordering is enforced instead of merely documented.

## What I could not verify

- **No real engine run.** Every `main()` test substitutes `run_engine`; the engine was never invoked
  against a real workbook. What is *not* substituted is the receipt/baseline machinery — those are
  the real `write_engine_receipt` and `write_generated_artifact_manifest` — so "pristine bundle
  reconciles exactly" is measured against real artefact accounting, just not real engine output.
- **Cost on a large estate is unmeasured.** The guard adds roughly one extra hash pass over
  `pbip/`, `semantic_models/` and `data/` (`data/` can hold `.hyper`/`.parquet`). It is the same
  work `write_engine_receipt` already does once per run, so I expect seconds, but I have no
  measurement on a 38-workbook bundle.
- **Whether `powerbi-report-author validate` ever writes inside a report folder.** If it did, it
  would make a subsequent run look dirty. `check_pbir_valid.py` only spawns the CLI and writes its
  own JSON, and the pristine-re-run test passes with the validator patched out — so this is reasoned,
  not observed on a machine with the CLI present.
- **Whether Power BI Desktop ever writes an artifact-suffixed file under `.pbi/`.** The exclusion is
  tested directly with such a file, so the behaviour is pinned either way, but the real-world shape
  is assumed.

### DIFF
--- a/docs/operator-runbook.md
+++ b/docs/operator-runbook.md
@@ -609,8 +609,15 @@ Reference run ✅ verified from `_convert.log`: `bound=23/38 failed=15 warned=20
 > indeterminate state — never a pass. The first cut reported *clean* through five such routes and a
 > reviewer destroyed a sentinel through each at exit 0. **Bundles built before this shipped have no
 > `engine_output_tree`, so their first destructive re-run will need both flags** — that is the
-> intended cost of not being able to prove what is in them. `--slice-only` never runs the engine and
-> is exempt.
+> intended cost of not being able to prove what is in them, and it is **one-time**: that
+> acknowledged run writes the tree, and the next pristine invocation exits 0 again. `--slice-only`
+> never runs the engine and is exempt.
+>
+> ⚠️ **`_build/` inside a `pbip/<project>/` folder DOES trip the barrier — deliberately.** It is
+> this repo's durable replay-script convention (*"every edit re-runnable from `_build/`"*), so it is
+> exactly where an agent's re-runnable work lives, and it sits inside a directory the engine
+> `rmtree`s. A **bundle-root** `<bundle>/_build/` is outside the destructive roots, is never deleted,
+> and is not guarded. The scope is the folder the engine destroys, not the folder name.
 
 What a bundle contains ✅ verified against the reference bundle by listing it, 2026-08-13:
 
--- a/docs/operator-runbook.md
+++ b/docs/operator-runbook.md
@@ -559,6 +559,7 @@ and 2/5 reproduced by running the command:
 | `6` | `EXIT_EMPTY_MODEL` | **live** — a model would open and load **zero rows** (an Import partition over a flat file that never landed) | read `<bundle>/empty-model-check.json`; see the block below |
 | `7` | `EXIT_INVALID_PBIR` | **live** ✅ reproduced — a shipped report FAILS the first-party `powerbi-report-author validate` (measured: `PBIR_ROLE_REQUIRED_MISSING` from a stubbed calc whose projection was dropped) | read `<bundle>/pbir-validity-check.json`; **bind the stub**, do not delete the visual |
 | `8` | `EXIT_BLANK_PLACEHOLDER` | **live** ✅ reproduced — a handover-backed `BLANK()` placeholder is consumed by a report filter or visual field binding | read `<bundle>/blank-placeholder-check.json`; translate the calc or remove the consuming report dependency knowingly |
+| `9` | `EXIT_BUNDLE_REWRITE` | **pre-engine refusal** (#250) — the `--output` bundle already holds work an engine re-run would delete, and/or a **different engine version** built it | land into a FRESH `--output`, or acknowledge with `--accept-bundle-rewrite` / `--accept-engine-version-change`; the acknowledgement is written to `<bundle>/bundle-rewrite-acknowledgement.json` |
 
 ❌ **Correction: exits 5 and 6 are NOT "pending branch only"** — the previous edition said so, and
 §5.1 check 10 was written against the same stale assumption. Both shipped 2026-08-13 (#109, #111).
@@ -593,6 +594,15 @@ Reference run ✅ verified from `_convert.log`: `bound=23/38 failed=15 warned=20
 > and the stale-output guard *exempts* that path — so the most destructive re-run is the one that
 > needs no `--force`. **All DAX approvals land in ONE run; per-workbook agent work starts only
 > afterwards.** ✅ verified from `run_estate.py`'s own docstring.
+>
+> ✅ **This is now ENFORCED, not just documented** (#250). `run_estate.py` re-hashes the target
+> bundle against `engine-output-receipt.json` and `input_manifest.json`'s `generated_artifacts`
+> **before** the engine runs, and **exits 9** naming every file the re-run would destroy. Proceeding
+> anyway needs `--accept-bundle-rewrite`, and the acknowledgement (including the file list) is
+> written to `<bundle>/bundle-rewrite-acknowledgement.json`. A bundle built by a **different engine
+> version** is refused the same way and acknowledged separately with
+> `--accept-engine-version-change` — deliberately two flags, so accepting a known engine bump does
+> not silently waive the destruction guard. `--slice-only` never runs the engine and is exempt.
 
 What a bundle contains ✅ verified against the reference bundle by listing it, 2026-08-13:
 
@@ -607,7 +617,8 @@ What a bundle contains ✅ verified against the reference bundle by listing it, 
 | `data/` | flat-file data landed next to the models that import it (4 folders on the reference bundle) | no |
 | `empty-model-check.json` | the exit-6 verdict: every model, its partitions, and why any of them would load zero rows. **Written on every run, pass or fail** | — |
 | `phase-timings.json` | per-phase elapsed seconds — see the note under §2's table for what it does *not* cover | — |
-| `engine-output-receipt.json` | hashes of the engine's output **and `engine` — what built this bundle** (§1.2) | — |
+| `engine-output-receipt.json` | hashes of the engine's output **and `engine` — what built this bundle** (§1.2). Read back by the pre-engine rewrite barrier (exit 9) to tell engine output from downstream work | — |
+| `bundle-rewrite-acknowledgement.json` | present only when someone knowingly re-ran the engine over downstream work or a different engine version — records when, which flag, and every file destroyed | — |
 | `.credential-gate-audit.log` | append-only audit trail; `credential_gate.py verify` reads it. **This is the non-narrative record** — trust it over any summary, including your own | — |
 | `input_manifest.json`, `source-provenance.json` | what went in, and where upstream it came from | — |
 | `deploy-journal.jsonl` | written by step 6, not step 5: intent-then-outcome per item (§6.1) | — |
@@ -1236,12 +1247,12 @@ Placeholders used in this document — and where the real value lives:
 
 `—` means that script cannot return that exit code.
 
-| script | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
-|---|---|---|---|---|---|---|---|---|---|
-| `preflight.ps1` | ready | critical missing | — | — | — | — | — | — | — |
-| `assess_estate.py` | assessed (may be secondary-degraded) | nothing assessed · sign-in refused (raises) | usage | **a PRIMARY listing is incomplete** | — | — | — | — | — |
-| `run_estate.py` | READY | engine failed | usage | **DoD failed** | approval collision | non-canonical engine | **empty model** | **invalid PBIR** | **BLANK() placeholder** |
-| `deploy_estate.py` | all deployed | item failed / refused | preflight | **incomplete by skip** | — | — | — | — | — |
+| script | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 |
+|---|---|---|---|---|---|---|---|---|---|---|
+| `preflight.ps1` | ready | critical missing | — | — | — | — | — | — | — | — |
+| `assess_estate.py` | assessed (may be secondary-degraded) | nothing assessed · sign-in refused (raises) | usage | **a PRIMARY listing is incomplete** | — | — | — | — | — | — |
+| `run_estate.py` | READY | engine failed | usage | **DoD failed** | approval collision | non-canonical engine | **empty model** | **invalid PBIR** | **BLANK() placeholder** | **bundle rewrite refused** |
+| `deploy_estate.py` | all deployed | item failed / refused | preflight | **incomplete by skip** | — | — | — | — | — | — |
 
 One run returns **one** code, in the order collision → DoD → invalid PBIR → BLANK() placeholder →
 empty model — so a bundle can trip a gate the exit code never mentions. Read the log body,
--- a/docs/operator-runbook.md
+++ b/docs/operator-runbook.md
@@ -559,7 +559,7 @@ and 2/5 reproduced by running the command:
 | `6` | `EXIT_EMPTY_MODEL` | **live** — a model would open and load **zero rows** (an Import partition over a flat file that never landed) | read `<bundle>/empty-model-check.json`; see the block below |
 | `7` | `EXIT_INVALID_PBIR` | **live** ✅ reproduced — a shipped report FAILS the first-party `powerbi-report-author validate` (measured: `PBIR_ROLE_REQUIRED_MISSING` from a stubbed calc whose projection was dropped) | read `<bundle>/pbir-validity-check.json`; **bind the stub**, do not delete the visual |
 | `8` | `EXIT_BLANK_PLACEHOLDER` | **live** ✅ reproduced — a handover-backed `BLANK()` placeholder is consumed by a report filter or visual field binding | read `<bundle>/blank-placeholder-check.json`; translate the calc or remove the consuming report dependency knowingly |
-| `9` | `EXIT_BUNDLE_REWRITE` | **pre-engine refusal** (#250) — the `--output` bundle already holds work an engine re-run would delete, and/or a **different engine version** built it | land into a FRESH `--output`, or acknowledge with `--accept-bundle-rewrite` / `--accept-engine-version-change`; the acknowledgement is written to `<bundle>/bundle-rewrite-acknowledgement.json` |
+| `9` | `EXIT_BUNDLE_REWRITE` | **pre-engine refusal** (#250) — the `--output` bundle holds work an engine re-run would delete, a **different engine version** built it, **or the barrier cannot assess either question** (missing/empty/truncated baseline, `--slice-only`-backfilled baseline, unreadable engine version) | land into a FRESH `--output`, or acknowledge with `--accept-bundle-rewrite` / `--accept-engine-version-change`; the acknowledgement, the destroyed files **and the coverage gaps** are written to `<bundle>/bundle-rewrite-acknowledgement.json` |
 
 ❌ **Correction: exits 5 and 6 are NOT "pending branch only"** — the previous edition said so, and
 §5.1 check 10 was written against the same stale assumption. Both shipped 2026-08-13 (#109, #111).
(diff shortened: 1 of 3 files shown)


Guust-Franssens:
## Review response — the five HIGH findings were one bug, and the diagnosis was right

> **Wherever the guard cannot assess something, it currently reports clean.**

All five reproductions confirmed, all five fixed under one rule: **unassessable is an explicit,
blocking, indeterminate state — never a pass.** No finding was disputed.

| # | root cause | fix |
|---|---|---|
| **1** | addition detection ran off the receipt's `ARTIFACT_SUFFIXES` allowlist, which excludes `.json` | the barrier allowlists **locations, never formats** |
| **2** | `textscan` `.txt` under `<project>.Data` / `data/` was in neither baseline | same fix — every file in every folder a re-run deletes is baselined |
| **3** | missing/empty/malformed baseline disabled the guard, separated from a pass only by warning **text** | coverage gaps are findings; they block without a flag |
| **4** | a truncated receipt made `version_changed` answer "no" | unknown either side ⇒ `version_indeterminate`, blocks |
| **5** | `--slice-only` backfill hashed the working copy and the guard trusted it | marker read, baseline distrusted, `--slice-only` writes no tree |

### What changed

**New authoritative baseline.** `write_engine_output_tree` records **every** file under `pbip/`,
`semantic_models/`, `reports/` and `data/` — no suffix filter — into `input_manifest.json`, written
before the receipt so `input_manifest_sha256` stays correct. Separate key from `generated_artifacts`
on purpose: that one is the tamper-audit baseline `check_migration_progress.py` depends on, and it
answers a different question. `--slice-only` deliberately does **not** write it.

**Coverage is now a first-class value.** `BundleCoverage(known, complete, gaps)`. `complete` (a tree
exists) is the only thing that makes an *added* file decidable; `gaps` is every reason the assessment
is partial. `blocks_on_downstream` fires on `drift.total **or** coverage.gaps`;
`blocks_on_engine_version` on `version_changed **or** version_indeterminate`.

**Not over-corrected.** With a partial baseline the barrier blocks on the gap and reports **no**
file list — naming some unrecognised files would imply the rest had been cleared
(`test_an_emptied_baseline_does_not_invent_a_file_list`). An existing-but-non-bundle `--output`
folder is still not a bundle, so a first run into a scratch directory is untouched.

**Cost, stated plainly:** every bundle built before this ships has no tree, so its first destructive
re-run needs both flags. That is the intended price of not being able to prove what is in it, and it
is documented in the runbook.

### The vacuous test — fixed first, as instructed

`test_slice_only_still_works_against_a_bundle_full_of_downstream_work` asserted only an exit code.
Confirmed vacuous before touching any guard code: `if not args.slice_only:` → `if True:` gave
`1 passed`. It now uses `_relanding(monkeypatch)`, asserts `calls == []` **and** that the sentinel
survives on disk — and the stand-in engine is genuinely destructive (`rmtree`s `pbip/` and `data/`),
so every test in the file that claims work survived is now checking the disk, not a call log.
Re-measured after the fix: the same mutation fails it (among 13).

### Verified, by exit code

Real CLI, `--dry-run`, against a bundle built through `run_estate`'s own writers in
`record_engine_output`'s exact order:

```
PASS exit=0  control: pristine bundle proceeds
PASS exit=9  HIGH 1: newly authored PBIR .json
PASS exit=9  HIGH 2: textscan .txt under <project>.Data
PASS exit=9  HIGH 2b: new flat file under data/
PASS exit=9  HIGH 3: no receipt, no baselines
PASS exit=9  HIGH 3b: emptied baselines (artifacts=[] and files={})
PASS exit=9  HIGH 4: truncated receipt (version guard)
PASS exit=9  HIGH 4b: --accept-bundle-rewrite does not answer it
PASS exit=9  HIGH 5: slice-only backfill laundering
PASS exit=0  control: --slice-only never hits the barrier, sentinel intact
PASS exit=0  control: both flags recover an unassessable bundle
```

Gates: `ruff` **0** · `.\.venv\Scripts\python.exe -m pylint scripts` **10.00/10, exit 0** ·
`tests/test_run_estate.py` **59 passed, exit 0** · full suite **2372 passed, 3 failed** — the same
three `test_upstream_repro_pins.py` reproduced on unmodified `master` earlier in this branch
(`git stash`, identical three), untouched by this change.

### Mutation set: 17 → 26, all caught, zero survivors

First run of the new set left **four survivors**, each a real hole where a *redundant* check was
masked by one that fired first — exactly the rot the reviewer's method is for:

| survived | why | isolation test added |
|---|---|---|
| slice-only backfill trusted | the missing-tree gap always fired first | tree + receipt intact, backfilled `generated_artifacts` beside them |
| absent/unreadable receipt records no gap | the version guard always fired first | `--accept-engine-version-change` given, so only the receipt gap remains |
| missing tree records no gap | the receipt gap always fired first | valid receipt, same engine, **no** tree, one added PBIR file — the realistic legacy-bundle shape |
| any existing folder treated as a bundle | nothing tested the over-reach direction | `--output` holding only `notes.md` must still proceed |

Re-run after adding them: **26/26 caught, 0 survivors**, baseline restored green. Scoring is on a
named `N failed`; a non-zero exit with no `FAILED` line is reported as SUSPECT and counted as a
survivor.

### Line cap — waived, deliberately, with the deferred fix named

`run_estate.py` is **1268/1200**. Shaving 68 lines means deleting the measured prose the repo treats
as the asset. Splitting is the right fix and is **not** blocked by ownership — it is blocked by a
mechanism: a new `scripts/*.py` with a `usage:` line must be classified in
`docs/agent-capability-wiring.md` (whose `agent-facing` status requires naming it in
`scripts/README.md`, contended by three agents) **or** carry `internal: true` + `internal-reason:`,
which **zero scripts in this repo currently use**. Pioneering that mechanism inside a security fix is
the wrong place for it. Waived per the sanctioned pattern (`assess_estate.py`, `deploy_estate.py`,
`check_unit.py`, `parse_tableau.py`), with the extraction named in the comment above the pragma.

**Happy to land that split as an immediate follow-up** if you would rather not carry the waiver — it
is a pure move plus one registry decision, and it needs an owner for `scripts/README.md`.

### Still unverified

- **No real engine run.** `run_engine` is substituted throughout; the stand-in now emulates the
  destructive `rmtree`, but not the engine's conversion. The baseline machinery under test is real.
- **Cost on a large estate.** The tree baseline hashes `data/` extracts (`.hyper`/`.parquet`) once
  more per run, on top of the receipt that already did. Expected seconds; not measured on a
  38-workbook bundle.
- **`--slice-only` on a stub PBIR exits 7 here** — the real `powerbi-report-author validate`
  rejecting a hand-made report, not the barrier. The reproducer asserts the barrier did not fire and
  the sentinel survived rather than asserting exit 0, and says why.
- **Whether Desktop ever writes an artifact-suffixed file under `.pbi/`** — the exclusion is tested
  directly with such a file, so the behaviour is pinned either way; the real-world shape is assumed.

Guust-Franssens:
## Round 2 — confirmed, fixed, and the diagnosis was exactly right

> *"This is not 'the exclusion list is wrong' — it is that **one predicate is doing two jobs**."*

That is the whole bug. `_is_guarded_path` served the rewrite barrier by borrowing
`_is_scratch_path`, which answers a different question — *"is this a stable deliverable worth putting
in the tamper-audit manifest?"* — so it skipped `_build`, `_work`, `scratch`, `temp`, `_probe`,
`tmp` and `_shots` **even inside a directory the engine `rmtree`s**.

And `_build/` is the sharpest possible case, for the reason you gave: `AGENTS.md` mandates *"every
edit re-runnable from `_build/`"*, so `pbip/<project>/_build/replay.py` is **where the replay script
for an agent's declared edits lives**. The barrier was destroying the reproduction instructions for
the very work it exists to protect.

### Fix

The barrier now owns its predicate. `_barrier_covers_path` (`scripts/run_estate.py:607`) excludes
**only** the volatile `.pbi` sidecars; inside `ENGINE_TREE_ROOTS` every other file is hashed.
`_is_scratch_path` keeps its job in `_is_generated_artifact`, where it belongs. The rename is part of
the fix — the old name invited exactly the reuse that caused this.

**Boundary preserved and now pinned in both directions.** A bundle-root `<bundle>/_build/` is
outside the destructive roots, is never deleted, and is still unguarded. Scope is the folder the
engine destroys, not the folder name.

### Verified, by exit code — real CLI, 17 routes

```
PASS exit=0  control: pristine bundle proceeds
PASS exit=9  round-2 HIGH: pbip/<project>/_build/replay.py
PASS exit=9  sweep: pbip/WB/{_build,_probe,_shots,_work,scratch,temp,tmp}/work.py   (7 routes)
PASS exit=0  boundary: bundle-root _build/ is NOT guarded
PASS exit=0  regression: .pbi sidecars still excluded (incl. .pbi/unapplied/Orders.tmdl)
PASS exit=9  regression HIGH 1/2/3/4/5: PBIR .json · textscan .txt · no baselines · truncated
             receipt · slice-only laundering
PASS exit=0  control: both flags recover it
```

### Tests

- `test_a_replay_script_nested_under_a_destructive_root_is_downstream_work` — the reported case;
  asserts exit 9, empty engine call log, **and the sentinel on disk**
- `test_a_bundle_root_build_folder_is_not_guarded` — the over-reach side
- `test_no_scratch_component_survives_inside_a_destructive_root` — parametrised over the **live**
  `run_estate.SCRATCH_DIRS` constant, so growing that set cannot silently re-open the hole for a name
  nobody thought to re-test

### Mutations 26 → 28, all caught, zero survivors

- **M26** barrier predicate re-coupled to `_is_scratch_path` → **8 failed**, including all seven
  parametrised names
- **M27** barrier over-reaches to a bundle-root `_build/` → **1 failed**
- M24 rewritten for the new predicate; the other 25 re-run unchanged

Scored only on a named `N failed`; a non-zero exit with no `FAILED` line counts as a survivor.

### Gates

`ruff` **0** · `.\.venv\Scripts\python.exe -m pylint scripts` **10.00/10, exit 0** ·
`tests/test_run_estate.py` **68 passed, exit 0** · full suite **2381 passed, 3 failed** — the same
pre-existing `test_upstream_repro_pins.py` trio, reproduced on stashed `master` earlier in this
branch. `git add -A` before committing; worktree clean.

**One transient to declare:** the first full run of this pass also failed
`.github/skills/pbip-model-refresh/tests/test_persist_atomicity.py::test_a_keyboard_interrupt_right_after_alignment_rolls_back_and_reraises`.
It **passes in isolation (38 passed, exit 0)** and **did not recur** on a second full run. Separately,
running that file alongside `tests/test_run_estate.py` produces 41 fixture ERRORs — and does so
**identically on stashed master** (59 passed + 41 errors there vs 68 + 41 here; the delta is exactly
my 9 new tests). Pre-existing, in a tree this branch does not touch.

### Line cap — keeping the waiver, and why

**1268 → 1279 (+11).** That is not "meaningfully further", so I did not take the split. Two reasons
beyond the number: landing a ~390-line mechanical move in the round meant to *close* a security fix
makes the security fix harder to review, and the split still needs a capability-registry decision —
a new `scripts/*.py` with a `usage:` line needs either a `docs/agent-capability-wiring.md` row
(whose `agent-facing` status routes through `scripts/README.md`) or `internal: true`, which **zero
scripts in this repo use**. I could not verify from here that `scripts/README.md` is actually free,
and guessing wrong turns a merge-ready fix into a conflict.

**Happy to land `bundle_rewrite_guard.py` as the very next PR** — it is a pure move plus that one
registry decision, and it wants an owner for `scripts/README.md` rather than an assumption.

### Thanks

Three rounds, and each one found something the previous had reasoned past: the ordering, then the
five fail-open routes, then a predicate quietly serving two masters. The `_build/` catch in
particular was not reachable from the code — only from knowing what this repo *uses* `_build/` for.

PR #385, opened 2026-08-29
fix(redact): remove the length floor that let a short warehouse password through

Guust-Franssens:
## The defect

`redact()` skipped any secret shorter than 8 characters, so a 7-character `TABLEAU_SF_PASSWORD` reached `manifest.json` verbatim. Reproduced against the shipped implementation:

```
len= 7  survives_unredacted=True   -> {"password":"Tr0ub4d","code":"401"}
len= 8  survives_unredacted=False  -> {"password":"[REDACTED]","code":"401"}
```

## Why the floor is gone rather than lowered

The floor was protecting against over-redaction, so I measured the over-redaction before removing it, on reflected errors of the shape these callers actually persist (synthetic values throughout):

| secret | chars | redactions | collateral (chars) | diagnostic tokens kept |
|---|---|---|---|---|
| 25-character password | 25 | 1 | 25 | all |
| `Tr0ub4d` | 7 | 1 | 7 | all |
| `ci` (a 2-char PAT name) | 2 | 3 | 6 | 3/3 |
| `e` | 1 | 30 | 30 | HTTP status kept |

The feared damage does not appear until 1–2 characters, and even there the outcome is **noise, not a leak** — the numeric status still survives. There is no length at which skipping is the better trade, so **a lower floor would only move the failure boundary**: at 4, a 3-character password is still published verbatim. Mutation **M2** exists specifically to pin that.

The 8 is not deleted, it is demoted: `NOISY_SECRET_LEN` now drives a one-line warning. Nothing is silently unprotected.

Two alternatives were measured and **rejected**:

- **Whole-word redaction around a short match.** Identical to plain replacement in every realistic case, 4–5× more text destroyed in the noisy ones (24 chars vs 6 for `ci`; 155 vs 31 for `e`), and it does *not* close the known-plaintext oracle it was meant to close (5/7 vs 7/7 secrets uniquely recovered from a fixed error template). Its natural spelling `\w*(?:secret)\w*` backtracks quadratically — 826 ms on 16 KB, still running after 180 s on 500 KB.
- **Withholding the whole text.** Destroys the diagnostic in every case, including the common one where collateral is zero. Mutation **M10** pins that it was not done.

## One pass, longest first

Removing the floor makes sequential replacement unsafe, so replacement is a single `re.sub` over an alternation, longest secret first. Both failures it prevents were measured:

- `redact(text, "abc", "abcdef")` replaced the head of the longer secret and left `def` — a tail of secret material — in the output.
- A 2-character secret matched inside a marker the same call had just inserted: `[R[REDACTED]ACT[REDACTED]]`.

## Call sites — all four verified with a 7-char password and a 2-char PAT name

| call site | secrets passed | verdict |
|---|---|---|
| `assess_estate.Site.scrub_text` | PAT secret, session token | PASS — no leak, `401`/`Unauthorized` kept |
| `capture_tableau_oracle.TableauSession._redact_response` | PAT secret, **PAT name**, token | PASS |
| `harvest_estate_assets.download` | PAT secret, **PAT name** | PASS (redact-before-truncate order preserved) |
| `provision_tableau_estate._describe` | PAT secret, SF password, DBX token | PASS — the five error paths at L327/L437/L449/L510/L641 the #362 review checked |

**Behaviour change worth knowing:** two call sites pass the human-chosen `TABLEAU_PAT_NAME` into `redact()`, so a short PAT name is now redacted where it previously was not. Measured cost on word-heavy text: 3 markers, 6 characters, every diagnostic token kept.

## Mutation evidence — 10/10 caught

Each mutation was written to disk, the bytes re-read and asserted to differ from the original before the tests ran, and scored **only** when pytest named the expected test on a `FAILED` line — a non-zero exit alone is not enough, since a collection error also exits non-zero. An earlier M4 spelling produced `1 error in 0.28s` (a `SyntaxError`) and was correctly scored MISSED, then rewritten.

| # | mutation | expected test | caught | pytest |
|---|---|---|---|---|
| M1 | restore the 8-char floor (the defect) | `test_redact_removes_a_short_human_chosen_password[Tr0ub4d]` | yes | 11 failed, 52 passed |
| M2 | lower the floor to 4 instead of removing it | `test_redact_removes_a_short_human_chosen_password[ab]` | yes | 3 failed, 60 passed |
| M3 | drop the whitespace-only guard | `test_redact_warns_and_skips_a_whitespace_only_secret` | yes | 1 failed, 62 passed |
| M4 | skip whitespace silently | `test_redact_warns_and_skips_a_whitespace_only_secret` | yes | 1 failed, 62 passed |
| M5 | silence the short-secret warning | `test_redact_warns_once_that_a_short_secret_makes_output_noisy` | yes | 1 failed, 62 passed |
| M6 | warn for every secret, not only short ones | `test_redact_does_not_warn_for_a_normal_length_secret` | yes | 1 failed, 62 passed |
| M7 | sort shortest-first | `test_redact_leaves_no_tail_of_a_longer_secret_when_a_shorter_one_shares_its_head` | yes | 1 failed, 62 passed |
| M8 | replace sequentially, not in one pass | `test_redact_does_not_match_inside_a_marker_it_just_inserted` | yes | 1 failed, 62 passed |
| M9 | drop the empty-secret guard | `test_redact_ignores_an_unset_secret_without_complaining` | yes | 1 failed, 62 passed |
| M10 | withhold the whole text (rejected alternative) | `test_redact_keeps_the_diagnostic_when_a_short_password_is_removed` | yes | 8 failed, 55 passed |

## Gates, by exit code

| command | exit | result |
|---|---|---|
| `ruff format --check scripts tests .github/skills` | 0 | 190 files already formatted |
| `ruff check scripts tests .github/skills` | 0 | All checks passed |
| `pylint scripts` | 0 | 10.00/10 |
| `pylint .github/skills/pbip-model-refresh/scripts` | 0 | 10.00/10 |
| `pylint .github/skills/powerbi-ai-readiness/scripts` | 0 | 10.00/10 |
| `pytest tests/test_tableau_env.py` | 0 | 63 passed |
| `pytest` × 4 call-site test files | 0 | 205 passed |
| `pytest` (full suite) | 1 | 2518 passed, 3 failed |

The three full-suite failures are in `tests/test_upstream_repro_pins.py` and are **pre-existing** — the same three fail identically on the unmodified base commit (engine `VERSION` drift, 2.339.0). Verified by stashing this change and re-running.

## Acceptance

1. ✅ A 7-character `TABLEAU_SF_PASSWORD` does not survive in `manifest.json`, stdout or an exception message — pinned end-to-end through `provision_tableau_estate._describe`.
2. ✅ Stated in `.env.example` next to `TABLEAU_SF_PASSWORD` / `TABLEAU_DBX_TOKEN`. (That file is outside the strict file group for this fix, but it is where the issue asks for it — flagging it for the reviewer.)
3. ✅ `test_redact_pins_the_old_floor_from_both_sides` — 7 and 8 characters.
4. ✅ Long machine-generated tokens unchanged; the three other callers verified directly and by their own suites.

## Not verified

- No timing test guards the linear-replacement property in CI; a wall-clock assertion would be flaky. The quadratic hazard belongs to the rejected regex approach, which is not in the diff.
- Whether a warehouse password could ever reach `harvest_estate_assets`' captured child stderr. It is not passed to `redact` there, and the engine's `fetch_tds.py` never handles one, so it is out of scope rather than covered.
- `tests/test_provision_tableau_estate.py:52` carries a comment referring to "`redact`'s 8-character floor", now stale. Left untouched deliberately — outside this fix's file group. The test itself still passes.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -59,6 +59,14 @@ FABRIC_WORKSPACE_ID=
 # refresh. One fixture (`Meridian Sales (No Embedded Credentials)`) is
 # deliberately published without one, and the manifest records that so it is
 # reproduced as-is.
+#
+# Redaction: there is NO minimum length. Every value set here is stripped out
+# of anything the scripts print or persist (`tableau_env.redact`), whatever its
+# length - a 7-character password used to fall through an 8-character floor and
+# reach `manifest.json` verbatim (#381). A value under 8 characters is still
+# redacted but also matches unrelated text, so the scripts warn once that the
+# surrounding diagnostics will be noisy. A whitespace-only value is the one
+# thing that cannot be redacted; unset it instead.
 # TABLEAU_SF_USER=
 # TABLEAU_SF_PASSWORD=
 # TABLEAU_DBX_TOKEN=
--- a/.env.example
+++ b/.env.example
@@ -60,13 +60,16 @@ FABRIC_WORKSPACE_ID=
 # deliberately published without one, and the manifest records that so it is
 # reproduced as-is.
 #
-# Redaction: there is NO minimum length. Every value set here is stripped out
-# of anything the scripts print or persist (`tableau_env.redact`), whatever its
-# length - a 7-character password used to fall through an 8-character floor and
-# reach `manifest.json` verbatim (#381). A value under 8 characters is still
-# redacted but also matches unrelated text, so the scripts warn once that the
-# surrounding diagnostics will be noisy. A whitespace-only value is the one
-# thing that cannot be redacted; unset it instead.
+# Redaction: there is NO minimum length and no exemption. Every value set
+# here is stripped out of anything the scripts print or persist
+# (`tableau_env.redact`) - a 7-character password used to fall through an
+# 8-character floor and reach `manifest.json` verbatim (#381), and a
+# whitespace-only value used to be skipped too. Only an unset (empty) value is
+# skipped. The XML-escaped form the Tableau client actually puts on the wire
+# (`SYNTH-&#228;-PASS` for a non-ASCII value) is redacted alongside the literal;
+# case-changed, NFD, percent-encoded and base64 forms are NOT, and no script
+# here emits one. A value under 8 characters is still redacted but also matches
+# unrelated text, so the scripts warn once that diagnostics will be noisy.
 # TABLEAU_SF_USER=
 # TABLEAU_SF_PASSWORD=
 # TABLEAU_DBX_TOKEN=
--- a/scripts/capture_tableau_oracle.py
+++ b/scripts/capture_tableau_oracle.py
@@ -310,7 +310,9 @@ class TableauSession:
             time.sleep(backoff_delay(attempt))
         raise RuntimeError(f"GET {path} exhausted {self.retry.max_attempts} attempts")
 
-    def export(self, path: str) -> tuple[bytes, float, dict[str, Any]]:
+    # One recovery ladder, and it now holds the raw body and its redacted copy side by side so
+    # classification and reporting cannot be confused for each other.
+    def export(self, path: str) -> tuple[bytes, float, dict[str, Any]]:  # pylint: disable=too-many-locals
         """GET a content-export endpoint, recovering from session loss and transient failures.
 
         Returns ``(body, elapsed_sec, stats)`` where ``stats`` records how much recovery was needed --
@@ -330,9 +332,18 @@ class TableauSession:
             elapsed = time.perf_counter() - started
             if status == 200:
                 return payload, elapsed, {"reauths": reauths, "retries": len(retries), "retry_reasons": retries}
-            text = self._redact_response(payload.decode("utf-8", "replace"))
-            kind, detail = classify_export_error(status, text)
-
+            raw = payload.decode("utf-8", "replace")
+            text = self._redact_response(raw)
+            # CLASSIFY the raw body, REPORT the redacted one. `_redact_response` is handed the PAT
+            # NAME, which is human-chosen: a short one rewrites Tableau's own error codes, and a
+            # `401002` mangled into `4[REDACTED]1[REDACTED][REDACTED]2` is read as a permanent
+            # source-credential failure instead of the recoverable session loss it is, so the view
+            # is abandoned and never re-authenticated. Redaction must never mutate syntax that
+            # control flow depends on. `detail` still comes from the redacted copy because
+            # `classify_export_error` truncates, and slicing an unredacted body can leave a secret's
+            # tail in the retained window.
+            kind = classify_export_error(status, raw)[0]
+            detail = classify_export_error(status, text)[1]
             if kind == "session_lost" and reauths < MAX_REAUTH_PER_VIEW:
                 # Re-auth is a SEPARATE recovery path from transient retry, and is deliberately NOT
                 # gated by the admission deadline. It is bounded instead by MAX_REAUTH_PER_VIEW (and
--- a/scripts/tableau_env.py
+++ b/scripts/tableau_env.py
@@ -41,6 +41,7 @@ import os
 import re
 import warnings
 from pathlib import Path
+from xml.etree import ElementTree
 
 # The secret half of the PAT credential. TABLEAU_PAT_SECRET is canonical; the engine's historical
 # TABLEAU_PAT_VALUE spelling stays accepted so existing environments keep working.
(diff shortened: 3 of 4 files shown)


Guust-Franssens:
## Round 2 — all six blind-review findings closed

Commit `192e726`. The round-1 fix was right about **length** and wrong about **what "the secret" is on the wire** — plus it created one functional regression. Each finding now has a gate test and a mutant proving the gate can fail.

### HIGH 1 — XML-escaped passwords survived

`tableauserverclient` builds its request with `ElementTree`, whose `tostring` defaults to `us-ascii`. A value holding U+00E4 reaches the wire as `SYNTH-&#228;-PASS`; `& < > "` escape too. `_wire_forms()` now redacts the **attribute and text-node forms** alongside the literal — produced by *running the same serializer*, not by guessing an encoding.

Case-changed, NFD, percent-encoded, base64 and newline-split forms are **deliberately not covered**. No call site emits one, and both the docstring and `.env.example` now say so rather than implying a guarantee the code cannot keep.

### HIGH 2 — the fix could disable the session-header rule

Literals were replaced *before* `_TABLEAU_AUTH_HEADER_RE` ran, so a secret inside the header **name** rewrote it and the second pass no longer matched:

```
before: 'HTTP 401 X-[REDACTED]-Auth: SYNTHETIC_SESSION_TOKEN_123'   <- token survives
after : 'HTTP 401 X-[REDACTED]-Auth: [REDACTED]'
```

### MEDIUM 3 — overlapping starts

```
before: 'password[REDACTED]YNTHETIC-LONG-SECRET-12345'
after : 'password[REDACTED]'
```

HIGH 2 and MEDIUM 3 are now **one machinery**: every span — literal *and* header — is measured against the **original** text, overlapping and touching spans are merged, and replacement happens in a single pass. Argument order is now structurally irrelevant rather than incidentally safe.

### MEDIUM 4 — the marker re-emitted the secret

```
before: redact('credential=[REDACTED]', '[REDACTED]')  ->  unchanged
after : 'credential=[HIDDEN]'
```

The marker is chosen per call from a ladder that shrinks its alphabet (`[REDACTED]` → `[HIDDEN]` → `[###]` → `""`), so a marker containing a supplied value is never used and **deletion is the terminal case**.

### MEDIUM 5 — my rationale was factually wrong

You are right. Only an **empty** pattern matches between every character; `" "` matches spaces. And `resolve_env` really does preserve a single-space password that the provisioner treats as truthy — so the excuse published the value it was excusing. **Only `""` is skipped now.** The noise warning stays.

### MEDIUM 6 — the functional regression

```
before: classify(redacted '401[REDACTED]2') -> source_credential, view abandoned
after : classify(raw '401002')              -> session_lost, reauth_count >= 1
```

`export()` classifies the **raw in-memory** body and redacts only the copy that leaves the HTTP layer. `detail` still comes from the redacted copy, because `classify_export_error` truncates and slicing an unredacted body can strand a secret's tail in the retained window.

You were right that I found these PAT-name call sites in round 1 and stopped at *noise*. The control-flow consequence was the part I missed.

### The vacuous test

Renamed to `test_redact_leaves_no_supplied_secret_in_the_output_when_one_is_part_of_the_marker` and re-asserted to the property it claimed: **no supplied secret survives**. Mutation **M7** is aimed squarely at it.

## Mutation evidence — 13/13

Written to disk, re-read, asserted equal to the intended bytes and different from the original; the whole tree asserted identical to the staged copy first. Scored **only** on pytest naming the expected test on a `FAILED` line. An earlier M4 spelling scored **MISSED with `70 passed`** — correctly: it was a no-op mutant, not a weak test, and was rewritten.

| # | mutation | file | caught | pytest |
|---|---|---|---|---|
| M1 | restore the 8-char floor | `tableau_env.py` | yes | 16 failed, 54 passed |
| M2 | lower the floor to 4 | `tableau_env.py` | yes | 7 failed, 63 passed |
| M3 | literal only, no XML wire forms | `tableau_env.py` | yes | 1 failed, 69 passed |
| M4 | literals first, header rule second | `tableau_env.py` | yes | 3 failed, 67 passed |
| M5 | do not merge overlapping spans | `tableau_env.py` | yes | 1 failed, 69 passed |
| M6 | always use the first marker | `tableau_env.py` | yes | 5 failed, 65 passed |
| M7 | marker from the FIRST needle only | `tableau_env.py` | yes | 1 failed, 69 passed |
| M8 | skip whitespace-only secrets again | `tableau_env.py` | yes | 1 failed, 69 passed |
| M9 | silence the short-secret warning | `tableau_env.py` | yes | 1 failed, 69 passed |
| M10 | warn for every secret | `tableau_env.py` | yes | 1 failed, 69 passed |
| M11 | drop the empty-secret guard | `tableau_env.py` | yes | 1 failed, 69 passed |
| M12 | withhold the whole text | `tableau_env.py` | yes | 13 failed, 57 passed |
| M13 | classify the REDACTED body | `capture_tableau_oracle.py` | yes | 1 failed, 69 passed |

Taking your point that a table you cannot re-run is worth less than one you can: the **commit body carries the exact anchor→mutant recipe for every row**, so each is reproducible with a one-line edit plus `pytest tests/test_tableau_env.py -q -rf`.

## Gates, by exit code

| command | exit | result |
|---|---|---|
| `ruff format --check scripts tests .github/skills` | 0 | 190 already formatted |
| `ruff check scripts tests .github/skills` | 0 | All checks passed |
| `pylint scripts` | 0 | 10.00/10 |
| `pylint .github/skills/pbip-model-refresh/scripts` | 0 | 10.00/10 |
| `pylint .github/skills/powerbi-ai-readiness/scripts` | 0 | 10.00/10 |
| `pytest tests/test_tableau_env.py` | 0 | 70 passed |
| `pytest` × 5 call-site test files | 0 | 227 passed |
| `pytest` (full suite) | 1 | 2525 passed, 3 failed |
| reviewer-finding reproduction | 0 | 0 failures across all six |

The three full-suite failures are `tests/test_upstream_repro_pins.py`, **pre-existing** — identical on the unmodified base commit (engine `VERSION` drift, 2.339.0), verified by stashing and re-running.

Worth recording: `pylint scripts` printed **`10.00/10` while exiting 8** on three findings (`R0402`, `R1721`, `R0914`). Exit code, never printed text.

Performance stays linear and matches your baseline: 16 KB **0.78 ms**, 500 KB **18.2 ms** (3-char needle). A 1-character secret matching every byte of a 500 KB body costs ~0.6 s — span volume, not backtracking.

## Reported, not fixed — please route

- **`harvest_estate_assets.py:231`** — same PAT-name shape. Untouched; PR #386 owns the file.
- **`assess_estate.py:305,415`** — `text = self._scrub(payload)` then `classify(status, text)`: the same classify-on-redacted-text shape as MEDIUM 6. **Latent, not live** — that call site is handed only machine-generated values (`_pat[1]`, `self.token`), so no human-chosen short value can collide with `401002`. Left alone deliberately rather than gold-plated, on your own "no call site emits it, say so instead" principle.

## Not verified

- No CI timing assertion guards linearity — a wall-clock gate would be flaky. Measured by hand, above.
- `tests/test_provision_tableau_estate.py:52` still carries a stale "8-character floor" comment. Outside this fix's file group; the test itself passes.

Issue #322, opened 2026-08-24
connections_manifest.py writes real customer server/database names to an operator-chosen, ungitignored path - one 'git add -A' from a public commit


Guust-Franssens:
## The near-miss

`connections_manifest.py` writes a deliverable containing **real customer server and database names** to an operator-chosen directory with **no default, no naming convention, and no `.gitignore` coverage**. In a live migration it landed at `ses-prep/` — an unprefixed, untracked, unignored folder at the repo root — and was **one `git add -A` away from being committed**.

**This repository is public.**

## Evidence

From a live SES migration, reported 2026-08-24 by the engineer running it:

> `ses-prep/` — real content: `connections.json` + `connections.md`, produced by `connections_manifest.py` (the "what needs a credential connection before reports work" doc, **17 sources listed**). This is a legitimate deliverable-shaped artifact that landed with **zero scratch-prefix and zero gitignore coverage** — the one case here that's a genuine near-miss, since it names real server/database targets (though not credentials) and was **one `git add -A` away from being committed**.

The `git add -A` risk is not hypothetical on that machine: the same engineer separately staged **111 untracked scratch paths** into a merge commit and caught it only by diffing the commit's file list against the real merge diff (now documented in `AGENTS.md`).

Verified on master:

```
scripts/connections_manifest.py:4
    usage: python scripts/connections_manifest.py --bundle <dir> --out <dir>
scripts/connections_manifest.py:489
    parser.add_argument("--out", type=Path, help="write connections.md / connections.json here")
```

`--out` has **no default**. Nothing steers the operator toward a protected location, so `ses-prep/` was a perfectly reasonable choice — which is the point. The tool asks an operator to pick a home for customer infrastructure metadata and gives them no guidance.

## Why this is not "just tidiness"

The output is *supposed* to name real systems — that is its whole function ("tell the customer WHICH data sources they must connect"). So unlike scratch, this artifact is **inherently sensitive by design** and will always contain what you would least like to leak. It therefore needs a protected home *by construction*, not by operator discretion.

Note the failure is not the operator's judgement: they chose a descriptive, non-scratch name for what is genuinely a deliverable, not scratch. The tool's contract simply never said where deliverables of this kind live.

## Proposed fix

[...]

PR #386, opened 2026-08-29
fix(harvest): judge every form of --out, and never anchor git through a junction

### LINKED ISSUE #374: harvest_estate_assets.py has the same guard/write asymmetry as #322, plus an NTFS-junction anchor hole the guard never reaches
## The defect

`unignored_output_paths` in `scripts/harvest_estate_assets.py:126` has the **same check/write asymmetry** that #322 fixed in `connections_manifest.py`, plus the **NTFS-junction anchor hole** that #322 could not close.

Found during blind review of the #322 fix. Both were reproduced against `connections_manifest.py`; the identical code shape is present here and has not been re-tested in this file.

## 1. Junction anchor — the guard is never consulted

`anchor = _existing_ancestor(out)`. When `out` is inside an NTFS junction, Windows resolves the junction for the child process's cwd, so `git rev-parse --is-inside-work-tree` answers from **outside** the repo and the function returns `[]` on its early exit — `check-ignore` is never reached.

Measured against `connections_manifest.py` (same logic):

```
mklink /J <lab>\linkdir <outside dir>
  --out=linkdir\manifest2   ->  EXIT=0, guard says safe, landed at junction target
  git add -A --dry-run      ->  add 'linkdir/manifest2/connections.json'   (customer data, stageable)

anchor picked : <lab>\linkdir        <- the junction itself
git rev-parse --is-inside-work-tree  (cwd = <lab>\linkdir)  ->  exit 128, "not a git repository"
```

Asked from the repo root it answers correctly (`check-ignore` exit 1), so **also anchoring at `Path.cwd()`** closes it.

## 2. Check/write asymmetry

The guard normalises with `expanduser().resolve()` while the caller writes the raw argument. On `connections_manifest.py` this let `--out=~/manifest` pass the guard (judged as `%USERPROFILE%\manifest`, outside any work tree) while writing to `<checkout>/~/manifest` — **exit 0, not ignored, stageable.** A literal `~` reaches `argv` from cmd.exe, a quoted PowerShell argument, or any programmatic call.

#322's fix: normalise once, then require **every** candidate path form to pass, and write to the checked resolved form.

## Why this matters here

`harvest_estate_assets.py` downloads **every workbook and published datasource on a site**. Its output is strictly more customer-identifying than the connections manifest. `.gitignore:120-127` states the rule and names the failure mode:

> This is a BACKSTOP, not the control. It cannot protect an output that does not start with `_` — see the `ses-prep/` near-miss in issue #322… Scripts that emit customer-identifying data must refuse to write to an unignored path, not rely on someone choosing a lucky name.

**This repository is public.**

## Acceptance

1. `--out` given as a literal `~/...` must refuse, not write inside the checkout
2. `--out` inside an NTFS junction must be judged from a non-traversed anchor
3. The out-of-checkout remedy must still work (`--out` outside the repo → writes, exit 0)
4. Tests built from the **real** `.gitignore`, not a reduced fixture — a 1-line synthetic fixture provably could not express the trailing-slash trap (`check-ignore -q -- "<path>/"` → exit 0 with an *empty* matched pattern)
5. Mutation-test each: guard disabled, trailing slash on the probe, fail-open on unanswerable git, fail-open on missing git binary, and dropping a path form

## Known residual, do not chase

Naming a junction *target* directly is still staged via the junction. Detecting that needs reparse-point enumeration of the checkout. Note it; do not fix it.

## Also worth carrying over

`connections_manifest.py`'s docstring claimed NTFS-junction coverage it did not have. If this file's prose makes the same claim, correct it — a privacy guard advertising coverage it lacks stops people looking.

Guust-Franssens:
## What

`harvest_estate_assets.py` downloads **every workbook and published datasource on a Tableau site**, and this repository is **public**. Its output guard had the check/write asymmetry #322 fixed in `connections_manifest.py`, plus a junction anchor hole. Both were reproduced by construction before the fix, on Windows **and** Linux.

## The two defects, measured

Fake checkout carrying this repo's **real** `.gitignore`; `git 2.55.0.windows.3`.

**1. Check/write asymmetry** — the guard normalised, the writer did not.

```
--out ~/lab-sweep    guard judged  C:\Users\<me>\lab-sweep     (no work tree -> allowed)
                     CLI wrote     <checkout>\~\lab-sweep\assets
                     check-ignore  exit 1 -- NOT ignored, stageable
```

**2. Junction anchor** — `_existing_ancestor` of the *resolved* path lands at the junction target, and git answers from there about somewhere else.

```
mklink /J <repo>\linkdir <outside>
--out linkdir\sweepout   anchor      <outside>
                         rev-parse   exit 128, "not a git repository"
                         guard       allowed
                         CLI         wrote through the junction
git add -A --dry-run at <repo>:
  add 'linkdir/sweepout/assets/harvested-workbook.twbx'      <- customer content, stageable
```

Asked from `<repo>`, the same git answers `check-ignore` **exit 1**, correctly.

## The fix

`output_path_forms` returns the **lexical** form (`absolute()`, nothing expanded or followed — the string git's own working-tree walk sees) and the **resolved** form (where the bytes land). `refuse_unignored_output` requires *every* form to pass; `main` then writes to the resolved one, so the checked path is the written path. `unignored_output_paths` no longer normalises — the caller chooses the form — and anchors at the first **lexical** ancestor whose work-tree root actually contains that form, so a junction below the root cannot move the question. An unresolvable path **raises** instead of degrading: unanswerable means unsafe.

Anchoring at `Path.cwd()` (the issue's remedy, and what #322 did) closes the measured case; the ancestor walk is a superset — it also holds when the script is invoked from *outside* the checkout, which is a normal way to run it.

| `--out` | before | after |
|---|---|---|
| plainly unignored in-repo | refuse | refuse |
| ignored in-repo (`_sweep*`) | proceed | proceed |
| genuinely outside the checkout | proceed | proceed |
| **junction pointing OUT of the checkout** | **proceed, wrote, stageable** | **refuse, exit 2, nothing written** |
| junction pointing INTO the checkout | refuse | refuse |
| **literal `~`** | **proceed, created `<repo>\~\...`** | **refuse, exit 2, no `~` created** |
| checkout itself reached through a junction | proceed | proceed (no false refusal) |

## Contradicting the issue, with evidence

The **second junction direction** — an `--out` that looks external but resolves *into* the checkout — was **already refused before this change**, precisely because the old guard resolved. That is why the resolved form is *kept alongside* the lexical one rather than replaced by it, and `test_a_junction_leading_INTO_the_checkout_is_refused` asserts the lexical form alone would **not** have caught it.

The containment test is load-bearing in the other direction: a checkout *reached through* a junction reports a toplevel that does not contain the lexical form, and asking it anyway yields `"outside repository"` (exit 128) → a refusal for a plainly ignored `--out`. `test_a_checkout_reached_through_a_junction_is_not_falsely_refused` pins that.

## Linux

Measured in WSL (git 2.43.0) with symlinks: **the same hole, and the same refusals after the fix.** The leak *shape* differs and the control test asserts each platform separately — git walks into an NTFS junction and stages the `.twbx` itself, while a POSIX symlink is staged as a link object (`add 'linkdir'`) and not walked into.

## Mutation testing — 11/11 caught

Each mutation was written to disk and **asserted present** before scoring, and scored on pytest's own count (≥1 failed, 0 errors, full 34-test collection) — not on exit code, which an import error also trips.

| # | mutation | pytest |
|---|---|---|
| M1 | guard disabled (always proceeds) | 16 failed, 18 passed |
| M2 | trailing slash on the check-ignore probe | 12 failed, 22 passed |
| M3 | fail-open when git cannot answer check-ignore | 1 failed, 33 passed |
| M4 | fail-open when the git binary is missing | 1 failed, 33 passed |
| M5 | drop the LEXICAL path form | 6 failed, 28 passed |
| M6 | drop the RESOLVED path form | 2 failed, 32 passed |
| M7 | stop the ancestor walk at the first non-work-tree (old anchor) | 3 failed, 31 passed |
| M8 | drop the lexical-containment applicability test | 1 failed, 33 passed |
| M9 | main writes the RAW argument instead of the checked form | 1 failed, 33 passed |
| M10 | fail-open on an unresolvable path | 1 failed, 33 passed |
| M11 | probe `--out` itself instead of a file inside it | 2 failed, 32 passed |

Every new refusal test asserts **twice** — first that the question the *old* guard asked returns "safe" for that fixture, then that the guard refuses anyway — so a fixture that never exercises the defect fails loudly instead of passing for the wrong reason (the shape found in review of #362). Fixtures use this repo's **real** `.gitignore`, byte for byte, and every link is unlinked in fixture teardown with `os.rmdir`/`unlink` so the target is never touched.

## Scope

`provision_tableau_estate.py` imports `refuse_unignored_output` and is **unchanged**: it already normalised once and wrote where it checked, so it had no asymmetry, and it inherits the junction fix for free. Its 105 tests still pass.

The residual the issue says not to chase — naming a junction **target** directly is still exposed via the junction — is now stated in the module docstring rather than left implied.

## Gates (exit codes)

- `ruff format --check scripts tests .github/skills` → **0**
- `ruff check scripts tests .github/skills` → **0**
- `.venv\Scripts\python.exe -m pylint scripts` → **10.00/10, exit 0**
- `pytest tests/test_harvest_output_guard.py tests/test_harvest_project_scope.py tests/test_provision_tableau_estate.py tests/test_connections_manifest.py` → **155 passed, exit 0**
- Full suite: **2514 passed**, 3 failures in `tests/test_upstream_repro_pins.py` — **pre-existing**, confirmed identical on the base tree with these changes stashed.

## Not verified

- The **pytest suite itself was executed on Windows only**. WSL has no pip and no passwordless sudo, so the Linux evidence comes from a standalone script driving the same guard code, not from pytest. CI will be the first Linux pytest run.
- Reparse points other than junctions and symlinks (mount points, appexec links, WSL DrvFs paths) were not tested.
- No live Tableau site was contacted; every fixture is an empty temp directory and no customer data was used or created.

Fixes #374

### DIFF
--- a/scripts/harvest_estate_assets.py
+++ b/scripts/harvest_estate_assets.py
@@ -15,6 +15,14 @@ their names, and THIS REPO IS PUBLIC, so when the target sits inside a git work 
 refuses to start unless git already ignores it -- see `unignored_output_paths` below. A target
 outside any work tree is fine and runs unguarded.
 
+What the guard does and does not cover (issue #374). It judges BOTH the path git's own working-tree
+walk would see (`--out` made absolute, nothing expanded or followed) AND the path the bytes actually
+land on (`~` expanded, every junction/symlink followed), refuses if EITHER is committable, and the
+run then writes to the second one -- so the path that was checked is the path that is written. It
+does NOT detect an `--out` that names a junction's TARGET directly: the write lands outside the
+checkout, but a junction elsewhere in the checkout still exposes it to `git add -A`. Finding that
+needs reparse-point enumeration of the whole checkout and is deliberately out of scope.
+
 Why this exists
 ---------------
 Both tiers are normally exercised on whatever workbook is in front of us, which selects for the
@@ -115,8 +123,65 @@ def _existing_ancestor(path: Path) -> Path | None:
     return next((p for p in (path, *path.parents) if p.is_dir()), None)
 
 
+def _is_relative_to(path: Path, parent: Path) -> bool:
+    """Lexical containment, the same question git's working-tree walk asks."""
+    try:
+        path.relative_to(parent)
+    except ValueError:
+        return False
+    return True
+
+
+def _resolved(path: Path) -> Path:
+    """`path` with `~` expanded and every junction/symlink followed. Isolated so the guard has one
+    place to fail closed when a path cannot be resolved at all."""
+    return path.expanduser().resolve()
+
+
+def _containing_worktree_root(out: Path, anchor: Path) -> Path | None:
+    """Root of the work tree that LEXICALLY contains `out`, or None when no work tree does.
+
+    Not simply `git rev-parse` at `anchor`, because `anchor` can be reached through a junction and
+    the answer is then about somewhere else entirely (issue #374). Measured on Windows with
+    `mklink /J <repo>\\linkdir <outside>`: `git rev-parse --is-inside-work-tree` run with
+    cwd=`<repo>\\linkdir` exits 128 "not a git repository" -- the OS resolved the junction for the
+    child process's cwd -- so a guard anchored there concludes "outside any work tree" and passes,
+    while `git add -A` at `<repo>` happily stages `linkdir/sweepout/...`. Asked from `<repo>` the
+    same git answers `check-ignore` exit 1, correctly. POSIX has the identical hole via a symlink.
+
+    So walk the LEXICAL ancestors and take the first work tree whose root actually contains `out` as
+    a string. Anchoring at `Path.cwd()` closes the measured case too and is what #322 did; the walk
+    is a superset - it also holds when the script is invoked from outside the checkout entirely.
+
+    The containment test is not decoration: a checkout REACHED through a junction reports a
+    toplevel that does not contain the lexical path, and asking that repo anyway yields
+    "outside repository" (exit 128) -> an unprovable answer -> a refusal for a perfectly ignored
+    output. Such a repo cannot see this path form, so it is not asked; the resolved form covers it.
+    """
+    for candidate in (anchor, *anchor.parents):
+        probe = _git(["rev-parse", "--show-toplevel"], candidate)
+        if probe is None:
+            # git being un-runnable is not evidence that no repository is here, so look for the
+            # checkout directly rather than letting a missing binary quietly disable the guard.
+            if any((parent / ".git").exists() for parent in (anchor, *anchor.parents)):
+                raise OutputPathNotIgnoredError(
+                    f"cannot run git, but {out} is inside a .git checkout, so it cannot be proven ignored"
+                )
+            return None
+        if probe.returncode != 0:
+            continue  # not a work tree here; a junction may have hidden the real one further up
+        root = Path(probe.stdout.strip())
+        if _is_relative_to(out, root):
+            return root
+    return None
+
+
 def unignored_output_paths(out: Path, artifacts: Sequence[str] = OUTPUT_ARTIFACTS) -> list[Path]:
     """Which artifacts under `out` git would offer to commit. Empty list means safe to write.
+
+    `out` is judged EXACTLY as given (made absolute, nothing expanded or followed): the caller
+    decides which form of the path to ask about, because the forms disagree and both matter - see
+    `refuse_unignored_output`, which asks about all of them.
 
     Two measured details decide this implementation, and getting either wrong yields a guard that
     silently always passes - worse than no guard, because it also reassures:
@@ -133,27 +198,19 @@ def unignored_output_paths(out: Path, artifacts: Sequence[str] = OUTPUT_ARTIFACT
     Raises `OutputPathNotIgnoredError` when git is present but cannot answer, or is absent while a
     `.git` checkout is in scope: an unprovable path is treated as unsafe, never as safe.
     """
-    out = out.expanduser().resolve()
+    out = out.absolute()
     anchor = _existing_ancestor(out)
     if anchor is None:  # pragma: no cover - a drive/filesystem root always exists
         return []
 
-    inside = _git(["rev-parse", "--is-inside-work-tree"], anchor)
-    if inside is None:
-        # git being un-runnable is not evidence that no repository is here, so look for the checkout
-        # directly rather than letting a missing binary quietly disable the guard.
-        if any((parent / ".git").exists() for parent in (anchor, *anchor.parents)):
-            raise OutputPathNotIgnoredError(
-                f"cannot run git, but {out} is inside a .git checkout, so it cannot be proven ignored"
-            )
-        return []
-    if inside.returncode != 0 or inside.stdout.strip() != "true":
+    root = _containing_worktree_root(out, anchor)
+    if root is None:
         return []  # outside any work tree: nothing here can be committed by accident
 
     unignored: list[Path] = []
     for artifact in artifacts:
         target = out / artifact
-        probe = _git(["check-ignore", "-q", "--", str(target)], anchor)
+        probe = _git(["check-ignore", "-q", "--", str(target)], root)
         # 0 = ignored, 1 = not ignored, anything else (128, or no git) = no answer, so do not guess.
         if probe is None or probe.returncode not in (0, 1):
             detail = "git could not be run" if probe is None else (probe.stderr.strip() or f"exit {probe.returncode}")
@@ -161,6 +218,33 @@ def unignored_output_paths(out: Path, artifacts: Sequence[str] = OUTPUT_ARTIFACT
         if probe.returncode == 1:
             unignored.append(target)
     return unignored
+
+
+def output_path_forms(out: Path) -> list[Path]:
+    """Every form of `--out` that must pass before customer content is written (issue #374).
+
+    A guard that judges one form while the writer uses another proves nothing. Both of these are
+    real, and each catches what the other cannot:
+
+    * **lexical** - `out.absolute()`, nothing expanded or followed. This is the string git's own
+      working-tree walk sees, so it is the honest answer to "would `git add -A` stage this?" for an
+      `--out` that traverses a junction OUT of the checkout, and for a literal `~` (measured: from
+      cmd.exe, a quoted PowerShell argument, or any programmatic call, `~` reaches argv unexpanded,
+      and `Path("~/x").absolute()` is `<cwd>/~/x` - inside the checkout, unignored).
+    * **resolved** - `~` expanded and every junction/symlink followed. This is where the bytes land,
+      and it is what catches an `--out` that looks external but is a junction pointing INTO the
+      checkout (measured: already refused before this function existed, precisely because the guard
+      resolved - which is why the resolved form is kept rather than replaced).
+
+    An unresolvable path raises rather than degrading to the lexical form alone: a form we cannot
+    compute is a question we cannot answer, and unanswerable means unsafe.
+    """
+    lexical = out.absolute()
+    try:
+        resolved = _resolved(out)
+    except (OSError, RuntimeError, ValueError) as exc:
+        raise OutputPathNotIgnoredError(f"cannot resolve {out}, so it cannot be proven ignored: {exc}") from exc
+    return list(dict.fromkeys((lexical, resolved)))
 
 
 def refuse_unignored_output(
@@ -172,12 +256,18 @@ def refuse_unignored_output(
 ) -> bool:
     """True when the run must STOP before downloading anything. Logs the reason either way.
 
+    Takes `--out` AS THE OPERATOR GAVE IT, not a pre-normalised copy: normalising before the call
+    discards the very form that catches a literal `~` (issue #374). Every form from
+    `output_path_forms` must pass; the caller then writes to the resolved one.
+
     `artifacts` and `hint` exist so a second tool that downloads customer content can reuse this one
     implementation rather than growing a near-copy that drifts. Pass the FILES that tool writes: the
     probe must name a file, never a bare directory (see `unignored_output_paths`).
     """
     try:
-        unignored = unignored_output_paths(out, artifacts)
+        unignored = list(
+            dict.fromkeys(path for form in output_path_forms(out) for path in unignored_output_paths(form, artifacts))
+        )
     except OutputPathNotIgnoredError as exc:
         message = str(exc)
     else:
@@ -559,9 +649,14 @@ def main() -> int:  # pylint: disable=too-many-locals,too-many-statements  # one
 
     logging.basicConfig(level=logging.INFO, format="%(message)s")
     # Before the engine, the .env, the database and above all the download: a customer's workbooks
-    # must never land somewhere this PUBLIC repo would commit them (issue #125).
+    # must never land somewhere this PUBLIC repo would commit them (issue #125). The guard is given
+    # `--out` RAW, so it can judge the literal argv form as well as the resolved one; the write then
+    # uses the resolved form, which the guard has just passed. Checking one form and writing another
+    # is the whole of issue #374 -- measured: `--out ~/sweep` from cmd.exe was judged as
+    # `%USERPROFILE%\sweep` (outside any work tree, so allowed) and written to `<checkout>\~\sweep`.
     if refuse_unignored_output(args.out, args.allow_unignored_out):
         return 2
+    args.out = _resolved(args.out)
     # One resolver, no fallback: the installed plugin is the single canonical engine (issue #107).
     try:
         scripts = engine_scripts_dir()
--- a/scripts/harvest_estate_assets.py
+++ b/scripts/harvest_estate_assets.py
@@ -138,16 +138,26 @@ def _resolved(path: Path) -> Path:
     return path.expanduser().resolve()
 
 
-def _same_dir(a: Path, b: Path) -> bool:
-    """Do these two spellings name the SAME directory on disk? Never a string comparison."""
+def _same_dir(a: Path, b: Path) -> bool | None:
+    """Do these two spellings name the SAME directory on disk? Never a string comparison.
+
+    Three-valued on purpose. `None` means the FILESYSTEM could not answer - a permission failure, a
+    disconnected share - which is not the same as "no", and collapsing it to `False` is how an
+    unignored in-repository output was allowed through: with every comparison erroring, containment
+    looked disproven rather than unproven and the guard proceeded (measured: `main()` exit 0, a
+    customer workbook name written to `parse-sweep.json`, `git status` staging it).
+    """
     try:
         return os.path.samefile(a, b)
     except OSError:
-        return False
+        return None
 
 
-def _identity_relative(node: Path, root: Path) -> list[str] | None:
-    """Segments from `root` down to `node` by FILE IDENTITY, or None when `root` does not hold it.
+def _identity_relative(node: Path, root: Path) -> tuple[list[str] | None, bool]:
+    """`(segments from root down to node, every comparison answered)` - by FILE IDENTITY, not text.
+
+    The second element is what separates "walked the whole ancestry and this really is not inside
+    `root`" from "could not tell". Only the first is safe to act on.
 
     Windows can spell one directory several ways that are not lexically relative to each other, and
     both `absolute()` and `resolve()` preserve every one of them, so comparing `--out` against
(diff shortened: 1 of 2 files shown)


Guust-Franssens:
**Update — the Linux gap in "Not verified" is now closed by CI itself.**

The PR body said the pytest suite had run on Windows only, and that CI would be the first Linux run. It has now run, and the numbers say every new test executed rather than skipping:

| | passed | skipped |
|---|---|---|
| `master` @ `8eb4b6e` ([run 33267212241](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/actions/runs/33267212241)) | 2490 | 22 |
| this PR ([run 33269915604](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/actions/runs/33269915604)) | 2502 | 22 |
| delta | **+12** | **+0** |

Twelve tests were added and twelve more passed, with the skip count unchanged — so the symlink fixtures (`lab_fixture`, `_link_dir`) were exercised on `ubuntu-latest`, not skipped past. Both CI jobs are green (`checks` 3m2s, `windows-bundle` 36s).

Still not verified: reparse points other than junctions and symlinks (mount points, appexec links, WSL DrvFs paths).

Guust-Franssens:
**Second commit ([`8c67f63`](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/commit/8c67f63)) — an independent concern, judge it separately from the #374 fix.**

Adjacent finding reported from #382: `test_the_two_harvesters_do_not_share_one_documented_folder` located each rule's comment block with `text.split("/_sweep*/")[0]`, which splits on the **first occurrence anywhere in the file** — so a comment quoting a rule verbatim earlier on hijacks it, and this test goes red on correct prose in a file no harvest change touches. #382's author reworded their comment rather than edit a file outside their group, and flagged it.

**The hazard is already live**, saved only by ordering: `.gitignore:109-110` quotes `` `/_harvest*/` `` and `` `/_sweep*/` `` verbatim, *below* the rules they quote. Reproduced against the **real** file by moving that one existing paragraph to the top — not one character of content changed:

| | old locator | new locator |
|---|---|---|
| real `.gitignore` as committed | ✅ both found | ✅ both found |
| same file, rule-quoting paragraph moved to top | ❌ **both lost** | ✅ both found |

`_rule_comment_block` now matches a whole **line** equal to the rule and returns the contiguous block above it. A rule that is missing, or listed twice, **raises** — an empty block satisfies `"x" not in block`, which would turn the collision check into a rubber stamp.

Mutation-tested against the committed (ruff-formatted) bytes, **4/4 caught**, each asserted on disk first and scored on pytest's counts (≥1 failed, 0 errors, full 36-test collection):

| # | mutation | pytest |
|---|---|---|
| N1 | revert to the split-on-rule-text locator | 2 failed, 34 passed |
| N2 | accept a duplicated rule (take the first) | 1 failed, 35 passed |
| N3 | no assertion at all — a missing rule yields an empty block | 1 failed, 35 passed |
| N4 | do not walk up — the comment block is always empty | 2 failed, 34 passed |

Deliberately **not** a change to `.gitignore` — that file belongs to #382 and its prose is correct; the test was wrong about how to read it. No closing keyword: this was reported agent-to-agent, so there is nothing to close.

It rides in this PR rather than its own because it edits `tests/test_harvest_output_guard.py`, which commit 1 rewrites — a branch off `master` would conflict on merge. Separate commit, so `git show 8c67f63` is the whole of it.

Guust-Franssens:
**Blind-review round 2 addressed in [`9b05ad5`](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/commit/9b05ad5)** — all three findings, plus one the mutation harness found on top.

All six new tests **fail against the committed implementation** (`git stash push -- scripts/harvest_estate_assets.py` → `6 failed, 36 passed`) while every pre-existing test still passes. That is the reproduction, recorded rather than asserted.

### HIGH 1 — path aliases, and a measurement that redraws the finding

Containment is now decided by `os.path.samefile`, never by path text. But measuring `rev-parse` with each alias as its **cwd** shows the three cases are not closed by one mechanism:

| alias | `rev-parse` as cwd | `samefile` | closed by |
|---|---|---|---|
| `\\?\<repo>` | exit 0, **long-form** toplevel | True | identity comparison |
| `<8.3 repo>\link-out` | exit 0, **long-form** toplevel | True | identity comparison |
| `\\localhost\c$\...` | **exit 128** | True | **HIGH 2's fail-closed rule** |

So HIGH 1's third case is fixed by HIGH 2's fix. Worth stating plainly — the tidy one-fix-per-finding story is wrong here.

### HIGH 2 — unanswerable is unsafe

An ancestor holding a `.git` entry that yields no work-tree root now **raises** instead of being skipped. Reproduced deterministically with a stale gitdir pointer (`rev-parse` 128, `.git` present → old code CLI exit 0 and wrote; now exit 2). The new `rev-parse`-only test uses an **ignored** `--out` (`_sweep`), so it can only pass by failing closed.

### MEDIUM 3 — the remedy's own spelling

`_lexical` collapses `.`/`..` with `os.path.abspath` — textually, never dereferencing, because dereferencing is exactly what would re-open the junction hole. `..\outside\out` now proceeds and writes.

### P14 — the harness found one I introduced

"Probe the caller's spelling instead of the canonical one" **survived** its first run. Rather than delete the mutation or the code, I measured it: `git check-ignore` answers `\\?\` and 8.3 spellings *identically* to the plain path, and a symlinked root the same way on Linux — so the canonical rebuild changes no verdict this suite can observe, and **no end-to-end test can kill it**. It is kept (a permissive answer for a half-understood spelling is the one failure this guard cannot survive) and pinned by a test explicitly labelled a **contract test**, not passed off as coverage.

### Gates (exit codes)

| gate | result |
|---|---|
| `\\?\`, UNC, 8.3+junction into the checkout | **exit 2**, nothing written |
| `rev-parse` 128 with `.git` present | **exit 2**, nothing written |
| `..\outside\...` | proceeds, writes there, no `REFUSING` |
| outbound junction / inbound junction / literal `~` | still refused |
| ignored checkout via junction, ignored `--out` via `\\?\` or 8.3 | still proceeds |
| mutation | **14/14 caught**, 44 collected every run, 0 collection errors |
| `ruff format --check` + `ruff check` | 0 |
| `.venv pylint scripts` | 10.00/10, exit 0 |
| harvest guard suite / four affected files | 44 passed / 163 passed, exit 0 |
| full suite | 2524 passed; same 3 pre-existing `test_upstream_repro_pins.py` failures, confirmed on the base tree |
| Linux (WSL, py 3.12.3, git 2.43.0) | all 11 scenarios, ALL CHECKS PASSED |
| reparse-point scan of the checkout after cleanup | **none** |

### Still not verified

`mountvol`, `mklink /D`, a directory AppExecLink (privileges / availability); no second physical volume; no OS-level unresolvable reparse path beyond the injected `OSError`. One honest conservatism: a `\\?\` path whose parent git cannot use as a cwd fails **closed** (refused) rather than being answered — noted in the docstring, not fixed.

Guust-Franssens:
**CI on the merge with current `master` (`4c32dfc`, which now includes #382 and #385): green.**

| | passed | skipped |
|---|---|---|
| `master` @ `4c32dfc` ([run](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/actions/runs/33276241128)) | 2557 | 22 |
| this PR ([run 33276756897](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/actions/runs/33276756897)) | 2575 | 26 |
| delta | **+18** | **+4** |

This branch adds 22 tests across its three commits, exactly 4 of them `WINDOWS_ONLY` (`\\?\`, UNC, 8.3+junction, ignored-alias-proceeds). +18 ran and passed on `ubuntu-latest`, +4 skipped — so nothing was silently skipped past on Linux, and the branch merges cleanly with #385's `redact()` change.

Guust-Franssens:
**Checked the #385 `redact()` interaction — no change needed here, and the interaction is already proven benign.**

Two corrections to the premise first: **#385 has merged** (`4c32dfc`; `master` is now `b3c7015`), and this PR's last CI run was green *against that merge* — so this is positive evidence, not a code reading. The call site is `harvest_estate_assets.py:363`, not `:231` (line numbers shifted with this branch's commits).

**Verdict: the redacted text is logged and persisted, never parsed, matched or branched on.**

- `redact()` is called exactly once (`:363`), on the fetcher subprocess's stderr/stdout.
- Its result reaches exactly two places: `row["download_error"]` → `parse-sweep.json`, and one `LOG.warning`.
- `git grep download_error` repo-wide: written at `:759`/`:766`, read **only by tests** — `test_harvest_project_scope.py:240` asserts it is `None`, `test_tableau_env.py:463` asserts the secret is absent. Both negative assertions, neither a classification.
- `summarise()` *does* group failures by shape (`:456-464`) — but on `r["ours"]["error"]` / `r["theirs"]["error"]` / `["unsupported"]`, produced by `parse_ours`/`parse_theirs`. **Neither goes through `redact()`.** The one place that pattern-matches error text never sees redacted text.
- Structurally there is no session-recovery decision to corrupt: on failure the loop records, warns and `continue`s (`:757-762`) — no retry, no re-auth. By design (docstring `:43-46`): every asset gets its **own** sign-in precisely because a shared token produced `401002` mid-loop. The oracle's regression was in code that classified `401002` to decide whether to re-authenticate; there is no such branch here.

Worst case from a short `TABLEAU_PAT_NAME` is extra `[REDACTED]` markers in a persisted failure message — cosmetic, recoverable, never control flow. No test pins the name-redaction, so removing it would be an unforced contract change; left alone.

Guust-Franssens:
**Rebased onto `origin/master` (`b3c7015`, includes #382, #385, #390) — clean, no conflicts.** Plus a 4th commit, [`86859c4`](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/commit/86859c4), independent of the #374 work.

### 1. The PAT-name / `redact()` question, now answered against the REAL merged-span implementation

Reproduced the sibling's exact shape at `harvest_estate_assets.py:363`:

| `TABLEAU_PAT_NAME` | result | `401002` intact | warns |
|---|---|---|---|
| `'0'` | `4[REDACTED]1[REDACTED]2: Unauthorized...` | ❌ | yes |
| `'10'` | `40[REDACTED]02: ...` | ❌ | yes |
| `'ci'` | `... user '[REDACTED]'.` | ✅ | yes |
| `'harvest-runner'` | unchanged | ✅ | no |

**Verdict unchanged: cosmetic here, no fix.** The shredding is real, but that text reaches only `row["download_error"]` (JSON) and one `LOG.warning`; the only classifier in the file, `summarise():456-464`, reads `parse_ours`/`parse_theirs` errors, which never pass through `redact()`.

**House convention checked rather than assumed** — post-#385, `redact()` call sites split 2/2 on passing the PAT name: `capture_tableau_oracle.py:206` (the sibling's own site, *after* their fix) and `harvest_estate_assets.py:363` pass it; `assess_estate.py:361` and `provision_tableau_estate.py:252` do not. The sibling fixed the ordering, not the argument list — so removing it here would diverge from their own post-fix state. Left alone.

**One cross-file observation, not actioned** (it is a `redact()` design question, and it affects `capture_tableau_oracle.py` identically): a short PAT *name* trips the new `NOISY_SECRET_LEN` warning — *"A configured secret is shorter than 8 characters"* — about something that is not a secret. Alarm fatigue on a credential-hygiene warning. Routing that is yours, not mine.

### 2. The long-path data point — checked, and the guard does not depend on it

git 2.55.0.windows.3, `core.longpaths` **unset**, OS `LongPathsEnabled=1`, a **310-character** `--out`:

| `--out` | `check-ignore` | guard | CLI |
|---|---|---|---|
| ignored prefix (`_sweep-long/...`) | **exit 0** | proceed | 1, wrote, `git status` clean |
| unignored prefix (`leaky-long/...`) | **exit 1** + `warning: unable to access ...` | **refuse** | **2**, nothing written |

git emits the warning on stderr and still returns the **right exit code**; the guard reads the code, not the warning. And a path git cannot answer for at all already raises under the fail-closed rule from `9b05ad5`. **No dependency, no fix** — pinned by a characterisation test that says exactly that in its docstring, creates nothing on disk (the guard judges non-existent paths by design, so no deep tree and nothing to clean up), and is **discriminating in both directions**: guard-always-proceeds → 1 failed; guard-always-refuses → 1 failed.

### Gates

`ruff format --check` + `ruff check` 0 · `.venv pylint scripts` 10.00/10 exit 0 · guard suite **45 passed** exit 0 · five affected files (incl. `test_tableau_env.py` for the #385 interaction) **236 passed** exit 0. Full suite deferred to CI on the merge ref, since the rebase changed the base.

Guust-Franssens:
**CI green on the rebased branch — and the skip-count jump is master's, not mine.**

| | passed | skipped |
|---|---|---|
| `master` @ `ce9ee46` ([run 33277880933](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/actions/runs/33277880933)) | 2602 | 80 |
| this PR ([run 33278196378](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/actions/runs/33278196378)) | 2621 | 84 |
| delta | **+19** | **+4** |

The 26 → 84 skip jump between my previous run and this one came from **#383** landing on master, not from anything here. This branch adds 23 tests across its four commits, exactly 4 of them `WINDOWS_ONLY` (`\\?\`, UNC, 8.3+junction, ignored-alias-proceeds) — so +19 ran and passed on `ubuntu-latest` and +4 skipped, with nothing silently skipped past. The merge ref already includes `ce9ee46`, so this is verified against the current tip.

Guust-Franssens:
**Round 2 addressed in [`f6d1066`](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/commit/f6d1066)** — the HIGH, plus your P14 correction adopted.

All three filesystem probes reproduced before the fix, all three now refuse:

| finding | before | after |
|---|---|---|
| dangling `.git` junction — `exists()` follows the link (`exists=False`, `lexists=True`, `rev-parse=128`) | CLI **exit 0**, wrote, stageable | **exit 2**, nothing written |
| `samefile()` raising — `OSError` collapsed to `False`, so containment looked **disproven** not **unproven** | proceeds, customer name in `parse-sweep.json` | **refuses** (and an *ignored* path refuses too) |
| nonexistent drive / disconnected UNC share — nothing to examine | proceeds | **refuses** |

`_same_dir` is now three-valued and `_identity_relative` reports whether every comparison answered; a work tree found but containment unprovable raises. A **clean** walk that finds no relationship still returns `None`, so the inbound-junction path is unchanged.

### P14 — you were right, I was wrong, and the honest label is why it got re-examined

I kept the canonical rebuild and called it a contract test on the grounds that no behaviour discriminated. `subst` does:

`
git -C Z:\ rev-parse --show-toplevel   exit 0, reports long C:\...\repo
check-ignore <canonical ignored path>  exit 0    -> ignored
check-ignore Z:\_sweep-p14\...         exit 128  -> UNANSWERABLE
`

Probing the caller's spelling would refuse a plainly **ignored** output. The production code was right; the *claim* about it was not. Contract test **replaced** by a substituted-drive behaviour test, and **P14 is now CAUGHT (1 failed)** instead of surviving.

### Evidence

- **Reproduction:** the three new fail-closed tests **fail against the previous commit** (`3 failed, 45 passed`). The substituted-drive test passes on both — by design; it kills a mutation rather than pinning this fix.
- **Mutation: 17/17 caught**, full 48-test collection, zero collection errors. P9 is gone (its anchor no longer exists); **Q2** covers it more precisely by collapsing the tri-state. New: Q1 `exists()` vs `lexists()`, Q2, Q3 drop the unprovable-identity refusal, Q4 drop the nothing-examined refusal — each **1 failed**.
- **Linux (WSL, py 3.12.3, git 2.43.0):** eleven scenarios incl. both new rules — ALL CHECKS PASSED.
- **Gates:** ruff 0 · `.venv pylint scripts` 10.00/10 exit 0 · guard suite **48 passed** · five related files **239 passed**.
- **Cleanup:** final scan — **0 reparse points**, **no `subst` mappings**, drive list unchanged (`C,G,Temp`).

### Not verified

`subst` has no POSIX equivalent and no discriminating case for the canonical rebuild is known there, so **on Linux that defence remains unpinned** — P14 is caught on Windows only. `mountvol`, `mklink /D`, directory AppExecLink still untested (privileges); no second physical volume.

Thanks for the `subst` case — that is two rounds running where the review found something I had reasoned past rather than measured.

PR #388, opened 2026-08-29
Fixes #254: gate the TMDL expression layout that stops Desktop opening a model

### LINKED ISSUE #254: TMDL DAX must stay single-line: a blank line inside an expression makes Power BI Desktop refuse to open the model
**Source:** customer field report. A **real, user-facing crash** - and it was produced by a background agent, so it is a live authoring hazard for our own personas.

## What happened

While rebuilding a stub table, an agent wrote a DAX measure with blank lines between logical fragments:

```
IF(

  DIVIDE(...) >= 0.98,

  "Healthy",

  IF(...
```

TMDL parser is **line-based** and treats a blank line as *"this property ends here"*. Result:

```
TMDL Format Error: Unexpected line type: Other!
```

Desktop then **refused to open the file at all**. Fix was collapsing to one line - the established convention everywhere else in this codebase, which was never actually written down.

## Why this deserves more than a note

- It is **agent-generated**. Multi-line DAX is normal, readable style in every other context, so an agent has no reason to avoid it unless told.
- It is **not caught by any structural validator we run** - it is a parse failure at Desktop open time.
- It is **total**: the model does not open, so there is no partial-success signal to notice.

## Ask

1. Document it in `.github/skills/powerbi-semantic-model-gotchas/SKILL.md` as an authoring rule.
2. Establish what the documented continuation syntax actually is and state it - "never use newlines" is easy to follow but may be stricter than the format requires, and long single-line DAX is genuinely hard to review.
3. Consider adding it to the offline model-integrity checks: cheap to detect on disk, catastrophic to miss.

Guust-Franssens:
## The measurement that changes the answer

The issue asked what the real TMDL continuation syntax is, and warned that "never use newlines" might
be stricter than the format requires. It is — and the issue's own diagnosis is off by one step.

Measured against `TmdlSerializer.DeserializeDatabaseFromFolder` (AMO 19.84.1, the same parser Power BI
Desktop uses), on 34 synthetic variants injected into a real committed model:

**Blank lines are innocent. The killer is starting the expression on the `=` line and then continuing
onto the next line.**

```tmdl
	measure 'M' = IF(          ← text after '=' commits TMDL to single-line mode
		DIVIDE(1,2) >= 0.98,   ← so this line must be a property. It is not.
		"Healthy", "Bad")      → TMDL Format Error: Unexpected line type: Other!
```

Case E (inline start + continuation, **no blank lines**) and case F (inline start + blank lines, the
shape in the issue) fail **identically** — and F's error points at the `DIVIDE` line, not at the blank
line above it.

### Legal, verified to open

| layout | |
|---|---|
| single line | ✅ |
| starts on the line **after** `=`, every line indented deeper than the properties | ✅ |
| …with **empty lines between fragments** | ✅ preserved as part of the expression |
| …whitespace-only lines / blank right after `=` / blank before the properties | ✅ |
| ` ``` `-enclosed block | ✅ verbatim |

Microsoft documents exactly this: *"Multi-line expressions must be indented one level deeper to the
parent object properties… Vertical whitespaces (blank lines without whitespaces) are allowed and are
considered part of the expression."*
([TMDL overview → Expressions](https://learn.microsoft.com/en-us/analysis-services/tmdl/tmdl-overview))

**So the rule is positional, not about newlines:** an expression is *either* entirely on the `=` line,
*or* it starts on the next line and stays indented past the properties. Mixing the two is the crash.
Single-line stays the house default (the corpus is 100% single-line — 167 documents, zero multi-line
expressions) but that is now a **style choice, not a format constraint**, so a long `VAR`/`RETURN`
measure no longer has to be an unreviewable one-liner.

### One case worse than the crash

A multi-line expression indented to the **same level as the properties** *parses clean* and swallows
them — the DAX silently ends `"...\nformatString: 0.0%"` and the property vanishes from the model.
Nothing else in the toolchain reports it.

## The gate

`scripts/check_datamodel.py`, findings → **exit 1** (matching the file's existing ladder; `2` remains
"no targets found"). Six codes:

| code | |
|---|---|
| `TMDL_EXPRESSION_CONTINUATION` | inline start + continuation — the reported crash |
| `TMDL_EXPRESSION_UNINDENTED` | expression dedents out of itself |
| `TMDL_EXPRESSION_ABSORBS_PROPERTY` | under-indented block eats its own properties |
| `TMDL_MISPLACED_DESCRIPTION` | `///` inside an object body |
| `TMDL_UNTERMINATED_EXPRESSION` | unclosed ` ``` ` — the rest of the file is unassessable, so it is **reported, not skipped** |
| `TMDL_UNREADABLE` | `check_tmdl_model` now decodes **strictly**; the previous `errors="replace"` turned a file the checker cannot read into a *clean* result |

**Error, not warning:** every one is total (the model does not open) except `ABSORBS_PROPERTY`, which
is silent corruption — arguably worse.

Legal-follower detection is an **allowlist on purpose**: `VAR x = 1` has the same
`<word> <word> = <value>` shape as `annotation X = Y` and is fatal
(`UnsupportedObjectType — VAR is not a supported property in the current context`), so a shape-based
rule would have missed it.

## Evidence

- **34/34 agreement** with the real parser across every variant.
- **Zero findings** on the 16 committed example models / 167 TMDL documents.
- `check_datamodel.py --all` → exit **0**; `pylint scripts` → **10.00/10, exit 0**; full suite
  **2571 passed** (3 pre-existing `test_upstream_repro_pins` failures, identical on master).
- **11/11 mutations caught** — table in the commit body. Mutation 3 survived the first run (the
  baseline-indent guard only false-positives on multi-line expressions, and the corpus has none, so it
  was untested); a named test was added for it.

## Known gaps

- The gate reads **layout, not validity** — well-laid-out nonsense still passes.
- Only `measure`/`column`/`calculationItem`/`expression`/`source` headers are scanned; `partition X = m`
  is deliberately excluded (it names a source *type*).
- `check_datamodel.py` crossed `max-module-lines` (1339). **Waived with a stated reason** rather than
  shaved; the real fix — splitting the M half from the TMDL half — is named in the pragma and deferred
  as unrelated to this change.

Fixes #254


Guust-Franssens:
### ⚠️ The red X is inherited from master, not from this PR

`checks` fails on **two tests in a file this branch does not contain**:

`tests/test_read_handover_evidence.py::test_the_terse_citation_is_bounded_in_BYTES_not_characters[\u732b-3]` and `[\U0001f4c8-4]` —
`OSError: [Errno 36] File name too long` on Linux, from writing a 100-character multibyte filename.

Evidence it is pre-existing:

| run | branch | head | result |
|---|---|---|---|
| [33279976752](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/actions/runs/33279976752) | `master` | `d6a1bde` | **fail** — same two tests, `2 failed, 2700 passed` |
| [33280258719](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/actions/runs/33280258719) | this PR | `fa7653e` | **fail** — same two tests, `2 failed, 2749 passed` |

Same two node IDs, same error. The file was added to `master` after this branch's base (`fc5422c`); `git log -- tests/test_read_handover_evidence.py` is empty here. The +49 delta is this PR's new tests, all passing.

**Only the `Tests` step failed.** Every other gate is green: ruff, pylint (all three roots), the data-model gate (`check_datamodel.py --all`), spec contract, privacy, navigation index, agent-capability wiring, and `sync_agent_conventions --check`.

`scripts/read_handover.py` and its tests are owned by another agent in this wave, so I have not touched them. It needs a byte-bounded filename, not a character-bounded one — the same distinction that test's own name is about.

Issue #234, opened 2026-08-19
Adopt the canonical _runs layout in remaining producers; add generated index and safe prune


Guust-Franssens:
## Current executable scope — September 7, 2026

The numbered `_runs/<NNN>-<slug>/` convention, `run.json`, allocator, permanent number reservations, external `--runs-parent`, and verification are shipped. Remaining work is bounded to adoption by still-independent producers, a generated index over authoritative run manifests, and a prune operation restricted to run-owned `scratch/`. Do not reopen the already-settled directory identity or rename existing runs.

---

[...]

PR #389, opened 2026-08-29
Fixes #235: Power BI Desktop is NOT long-path aware - measured, and gated at 259/247

### LINKED ISSUE #235: Bundles may already exceed default Windows MAX_PATH; our build machine masks it (LongPathsEnabled=1)
## Summary

Every bundle this toolkit produces may already exceed default Windows `MAX_PATH`, and we cannot
currently tell, because **this build machine has a non-default opt-in enabled that masks it.**

Surfaced during blind review of #234 (Claude Opus 5), then verified directly.

## Measured

```powershell
Get-ItemProperty 'HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem' -Name LongPathsEnabled
#   LongPathsEnabled = 1        <- NOT the Windows default

$f = Get-ChildItem _bundle-208,_bundle-op -Recurse -File
($f | Where-Object { $_.FullName.Length -gt 260 }).Count   # 93
($f | Where-Object { $_.FullName.Length -gt 282 }).Count   #  0
```

**93 files already exceed 260 characters.** Default Windows is `LongPathsEnabled = 0` with
`MAX_PATH = 260`, so on a stock machine those 93 files are in the danger zone. The reason we have
never noticed is that `LongPathsEnabled=1` only helps processes whose manifest declares
`longPathAware` — **Python 3.6+ does**, which is precisely why our generator can write them.

## Why this matters more than an internal inconvenience

`docs/deterministic-tier-integration.md:451` records the failure mode as *"ships a bundle that dies
on another machine"*. The product of this toolkit **is** a shipped bundle. A path that writes fine
here and fails on a customer's stock Windows box is the worst shape of defect we have.

I previously wrote, in #234, that "282 is an archive budget, not a filesystem limit — we run at 269
on disk today without trouble." That sentence is **wrong in its reasoning**, and this issue exists
because of it. "Without trouble" was a property of this machine, not of the artifact.

## The unknown that decides severity

**Is Power BI Desktop long-path aware?**

- If **yes** — a stock machine can still open a bundle we produced, and this is a packaging concern
  bounded by the 282-char archive break.
- If **no** — the effective write-and-open ceiling is **260**, the tree is **already 9 characters
  over it**, and every consumer on a default machine is at risk today.

Neither reviewer could answer it: both were read-only and it requires launching Desktop. This is the
highest-value unknown currently open on the toolkit.

## Also unbounded, and nobody controls it

The engine tail is **196 characters** in our worst current case, and it is driven by a **48-char
workbook name appearing twice** in the path:

```
pbip\Section_15_-_Tableau_Sales___Customer_Dashboards\Section_15_-_Tableau_Sales___Customer_Dashboards.Report\definition\pages\<page>\visuals\<visual>\visual.json
```

A 70-character customer workbook name adds roughly 44 characters and breaches 282 **regardless of
any directory convention we adopt**. Our own prefix is ~12 characters, so shortening it cannot save
a workbook whose name is long. No version of #234 accounts for this, and it should not pretend to.

## Proposed work

1. **Measure whether Power BI Desktop is long-path aware** — open a bundle whose PBIP path exceeds
   260 chars on a machine with `LongPathsEnabled=0`, or inspect the Desktop executable's manifest for
   `longPathAware`. This decides everything below.
2. **Emit a warning at bundle-write time** when any produced path exceeds 260, naming the file. Cheap,
   and it converts a silent downstream failure into a build-time signal.
3. **Pin the engine tail as a measured constant** (196) with the command that produced it, so the
   budget arithmetic in #234 is reproducible rather than asserted.
4. **Add a total-path check to the archive/ship step**, since 282 is the only empirically observed
   break point — one anecdote, one doc line, no reproduction, so it should be treated as a
   conservative ceiling rather than a measured limit.
5. **Consider shortening the engine's doubled workbook-name segment** — this is upstream
   (`Yarbrdab000/tableau-fabric-skills`) and is the single largest lever available, worth far more
   than any prefix we control. File separately once (1) tells us how urgent it is.

## Not in scope

The `_runs/` convention (#234). That work assumes this constraint rather than resolving it, and its
prefix arithmetic is unaffected by the outcome here — 10 chars beats 12 either way.

Guust-Franssens:
## The answer to #235's deciding question

**Power BI Desktop is NOT long-path aware.** The issue called this "the highest-value unknown
currently open on the toolkit" because it decides severity. It is the **bad** branch: the effective
write-and-open ceiling is **259**, and a bundle over it cannot be opened on *any* Windows machine —
not just a stock one.

Two independent measurements, 2026-08-29. Full write-up: [`docs/windows-path-limits.md`](docs/windows-path-limits.md).

### 1. Static — the embedded application manifest

| binary | role | `longPathAware` |
|---|---|---|
| `PBIDesktop.exe` 2.157.828.0 | the application | **absent** (RT_MANIFEST id 1, 1929 B: `dependency` + `compatibility` only) |
| `msmdsrv.exe` | AS engine that loads the model | **absent** |
| `Microsoft.Mashup.Container.NetFX45.exe` | M engine that reads sources | **absent** |
| `python.exe` | **control (positive)** | present |
| `explorer.exe` | **control (negative)** | absent |

**This is not treated as conclusive on its own, and that nuance matters.** A process can also bypass
`MAX_PATH` by prefixing `\\?\` itself. Measured here, `node.exe` and `pwsh.exe` both report the
manifest **absent** yet read a 262-char path fine (libuv and .NET Core prefix internally). That is
exactly why this issue's earlier probe saw *every* consumer succeed at 460 chars and concluded "the
masking is total" — several of those tools were never depending on the registry at all. So the
manifest was treated as a hypothesis and settled with a live A/B.

### 2. Live A/B in Desktop — conclusive

Byte-identical copies of `examples/shipping-kpis/fabric` (27 files), differing **only** in root
length, on this machine with `LongPathsEnabled = 1`:

| copy | longest path | `.pbip` entry point | result at t+190 s |
|---|---:|---:|---|
| control | 159 | 78 | **opens** — title resolves to `ShippingKPIs` |
| long | 268 | 187 | **fails** — title stuck at `Untitled - Power BI Desktop` |

Both entry points are far below any limit, so the failure is attributable to a deep child and nothing
else. The long copy raised a modal dialog, **"Issues were found"**:

> Cannot read `'<root>\ShippingKPIs.Report\definition\pages\<page>\visuals\<visual>\visual.json'`.
> The specified path, file name, or both are too long. **The fully qualified file name must be less
> than 260 characters, and the directory name must be less than 248 characters.**

## The ceiling, and why that number

Taken from Desktop's own error text, not folklore and not a bisect:

* `FILE_CEILING = 259` — "must be **less than 260**"
* `DIR_CEILING = 247` — "the directory name must be **less than 248**"

**The directory rule costs nothing and covers a real gap.** In a PBIR tree the deepest file is
`visual.json`, whose `\visual.json` tail is exactly 12 chars — and `260 − 248` is also exactly 12. For
that file the two rules bite at the same point, so no false-alarm surface is added. It is only
*stricter* for shorter names (`page.json`, `.platform`) — precisely what a file-only check misses. On
the long fixture, the check flagged 24 paths and the `visuals\<id>` **directories** were among them.

**282 is not encoded.** It remains an unreproduced archive anecdote; pass `--ceiling 282` to test it.

## What was built

`scripts/check_path_ceiling.py` — exit `0` clean / `1` findings / `2` usage / `3` could not evaluate.

* **Cannot inherit this machine.** The verdict is pure string arithmetic; it never asks the OS whether
  a path opens, so neither `LongPathsEnabled=1` nor the host OS can soften it. Linux CI produces the
  same numbers. The registry value *is* read and printed — because #235 exists entirely because nobody
  printed it — but it is context, never an input.
* **Reports the portable number.** "93 files over 260" is a fact about one disk. The longest **tail**
  and the `259 − tail` install-root budget survive relocation. `--min-root-budget N` gates on it
  (opt-in, since the right value depends on where the customer unpacks).
* **Unknown never passes.** Unmeasurable paths are counted and named as `unknown`; an empty target is
  `no_paths`, not `ok`.

### Current worst offenders — `_runs/coldrun-2.339.0-20260829/bundle` (read-only, not modified)

```
measured    : 114 paths (73 files, 41 directories)
longest     : 251 chars  .../pbip/Meridian Revenue by Region/Meridian Revenue by Region.Report/
                          definition/pages/page-ws-Revenuebb7d27f78/visuals/v-RevenuebyRegio864a62f6/visual.json
longest tail: 153 chars  ->  root budget 106
near ceiling (> 240): 3            EXIT=0
```

Clean, with 8 chars of absolute headroom — and the tail is driven by a **26-char workbook name
appearing twice** (59 chars of the 153). A 70-char customer name adds ~88 and breaches the ceiling
regardless of any directory convention we adopt. That doubled segment is upstream
(`Yarbrdab000/tableau-fabric-skills`) and is the largest lever available; our ~12-char prefix cannot
rescue it.

## Mutation table — 20/20 caught

Each scored only on a genuine `N failed` in a **named** test, with the mutation asserted present on
disk first; any pytest exit other than 1 scores `INVALID`, not caught. **The harness earned its
credibility by first reporting 16/20** — the three `EXIT_*` survivors were a real defect: the tests
compared against `cpc.EXIT_*`, so mutating a constant moved the assertion with it. Fixed by pinning
the ladder to literals. Full table in the commit body.

| mutation | named test that failed |
|---|---|
| `FILE_CEILING` 259 → 260 | `test_shipped_ceilings_are_the_measured_desktop_limits` |
| `DIR_CEILING` 247 → 259 | `test_shipped_ceilings_are_the_measured_desktop_limits` |
| over-ceiling `>` → `>=` | `test_path_exactly_at_the_ceiling_is_clean` |
| file/dir ceilings swapped | `test_directory_ceiling_is_applied_to_directories_not_the_file_ceiling` |
| unknown treated as clean | `test_unreadable_directory_is_unknown_never_ok` |
| empty target treated as clean | `test_empty_target_cannot_be_judged_clean` |
| root-budget arithmetic inverted | `test_root_budget_is_the_ceiling_minus_the_longest_tail` |
| min-root-budget gate disabled | `test_min_root_budget_gate_fails_a_tree_that_cannot_be_relocated` |
| `EXIT_OVER_CEILING` 1 → 0 | `test_exit_code_ladder_matches_the_house_contract` |
| `EXIT_SKIPPED` 3 → 0 | `test_exit_code_ladder_matches_the_house_contract` |
| `EXIT_USAGE` 2 → 0 | `test_exit_code_ladder_matches_the_house_contract` |
| **verdict inherits host `LongPathsEnabled`** | `test_verdict_is_identical_whatever_the_host_registry_says` |
| tail becomes absolute length | `test_tail_is_independent_of_where_the_tree_sits` |
| `--warn-only` stops suppressing | `test_warn_only_reports_but_never_fails` |
| near-ceiling band made blocking | `test_near_ceiling_paths_are_advisory_and_do_not_fail` |
| offenders counted, not named | `test_offenders_are_named_not_merely_counted` |
| unknown offenders not named | `test_unreadable_directory_is_unknown_never_ok` |
| JSON report not written | `test_json_report_is_written_and_machine_readable` |
| registry value dropped | `test_registry_value_is_reported_so_a_reader_is_never_fooled` |
| negative thresholds accepted | `test_negative_threshold_is_rejected` |

**No test creates a path over 259 chars.** That is the obvious way to exercise a `MAX_PATH` check and
the reliable way to make CI red on exactly the machines this issue is about — a stock Windows runner
cannot create such a path, so the *fixture* would fail, not the assertion. Boundary tests drive a
**low ceiling over a short tree** instead: identical code path, works on Linux, and it buys a better
property — the fixture files are short and openable *right now*, and the check still calls them over
ceiling, which **is** the proof that the verdict is arithmetic rather than an "can I open this?" probe.

## Also measured, and it would have made this PR red

`pylint` on the Linux CI runner cannot resolve `import winreg`. Confirmed by controlled experiment: a
Unix-only `import posix` linted on Windows emits `E0401`, and only an explicit
`# pylint: disable=import-error` silences it (`useless-suppression` is not enabled here, so it costs
nothing on Windows).

## Proposed follow-up — deliberately NOT done here

`scripts/preflight.ps1` has **zero** long-path checks. It should report the **pair** — the host's
`LongPathsEnabled` *and* the longest generated path — because the interaction is the risk and either
number alone is harmless. `preflight.ps1` was out of scope for this pass; proposing it rather than
doing it.

## Scope note

`scripts/README.md`, `docs/INDEX.md` and `docs/agent-capability-wiring.md` carry **one-line
registration entries only**. They were not otherwise in scope, but `check_navigation_index.py`,
`check_agent_capabilities.py` and `test_every_script_is_documented_in_the_scripts_readme` all fail
without them — adding a script to this repo is not optional in those three files.

## What I could NOT verify

* **The directory rule (247) was not isolated experimentally.** In a real PBIR tree you cannot build a
  case where a directory breaks 247 while every file stays within 259 — the 12-char `\visual.json`
  tail makes them coincide. The value comes from Desktop's stated rule; an artificial fixture that
  separated them would not generalise.
* **The exact boundary was not bisected.** Measured 159 → opens and 268 → fails; 259/247 come from the
  implementation's own message. Arguably stronger than a bisect, but not the same evidence.
* **Only Desktop 2.157.828.0 (Store package) was tested.** An MSI install or a future build could
  differ; the manifest probe re-answers it in seconds.
* **Desktop was not tested on a machine with `LongPathsEnabled = 0`.** It cannot be *better* there
  (the registry opt-in is a prerequisite, not a hindrance), but it was not measured.
* **Whether the archive/ship step has a lower budget of its own.** 282 remains one anecdote.

## Gates (by exit code)

| command | exit |
|---|---|
| `ruff format --check scripts tests` | 0 |
| `ruff check scripts tests` | 0 |
| `.venv\Scripts\python.exe -m pylint scripts` | 0 (10.00/10) |
| `pytest -q tests/test_check_path_ceiling.py` | 0 (29 passed, 1 skipped off-platform) |
| `pytest -q` (full) | 1 — **only** the 3 pre-existing `tests/test_upstream_repro_pins.py` failures, which fail identically on master (verified by stashing this branch's files: baseline green for everything else) |

Desktop instances opened for the experiment (control + long + one stray from a path typo) were all
closed by literal PID; `Get-Process PBIDesktop` returns 0. The >260-char fixture tree was removed with
a `\\?\`-prefixed `rmtree`; longest path remaining in the worktree is 239. `git status --porcelain` is
clean.


Guust-Franssens:
## Update — the boundary is now pinned, and the mechanism corrects the severity *upward*

The crash report from my first experiment identified the actual mechanism, and it changes the
conclusion in a way worth stating plainly.

### 1. `LongPathsEnabled` is a red herring for Desktop — this is worse than the issue assumed

The refusal is **not** an OS error Desktop inherited. It is Desktop's **own managed guard**:

```
Microsoft.PowerBI.Packaging.Project.PBIProjectUtils.EnsureNotLong(String path, Boolean isFolder)
  at Microsoft.PowerBI.Client.Windows.Services.DiskProjectFilesReader.<GetAsync>d__2.MoveNext()
```

surfaced as `FilePathTooLongError` inside `Error Reading StorageSection: ReportDocument`.

`EnsureNotLong` is a length comparison executed **before any filesystem call**. So the registry opt-in
cannot help, and does not — **every measurement on this branch was taken with `LongPathsEnabled = 1`,
and Desktop refused anyway.**

So the framing is **not** *"customers on stock Windows are at risk."* It is **every consumer on every
machine, including ours, regardless of registry configuration.** The registry setting only ever
governed whether our *generator* could **write** these paths — Python declares `longPathAware`, so
here it can. That asymmetry *is* the defect: **we produce artifacts we can never open.**

My first commit said "Desktop is not long-path aware." True, but that describes a symptom. This says
why, and the why changes who is affected.

### 2. Boundary now pinned empirically — one character decides it

Three byte-identical copies of `examples/shipping-kpis/fabric`, differing **only** in root length,
each judged by **two independent instruments** (a window title is a *loading state* before it is a
verdict; the Desktop Bridge is not):

| copy | deepest file | deepest dir | `.pbip` | title @ t+200 s | `powerbi-desktop status --pid` |
|---|---:|---:|---:|---|---|
| control | 200 | 188 | 119 | `ShippingKPIs` | (not probed) |
| **F259** | **259** | **247** | 178 | `ShippingKPIs` | **`ready` / `connected`**, pages enumerated |
| **S260** | **260** | **248** | 179 | `Untitled - Power BI Desktop` | **`error`** — *"Host is not ready to accept operations"* |

`FILE_CEILING = 259` / `DIR_CEILING = 247` are unchanged from my first commit — they are now
**measured** rather than read off an error message.

**Honest limit:** S260 crosses *both* boundaries at once, so that pair cannot say **which** guard
fired. F259 opening proves 259 and 247 are both legal, which is what a gate needs. Separating them
needs a tree whose deepest directory breaks 247 while every file stays within 259 — and the PBIR
layout cannot produce one (`\visual.json` is 12 chars; `260 − 248` is also 12).

### 3. Re-measured on the 52-unit estate run, which landed mid-task

`_runs/estate-2.339.0-20260829` (read-only, not modified):

```
measured    : 4851 paths (2548 files, 2303 directories)
longest     : 287 chars   (28 over the ceiling)
OVER CEILING: 183 paths
near ceiling (> 240): 279 paths                                   EXIT=1
```

**3 of 51 units are already over the ceiling as they sit:**

| unit | paths over | longest tail | root budget |
|---|---:|---:|---:|
| `Section 15 - Tableau Sales _ Customer Dashboards` | 48 | 197 | **62** |
| `Section 12 - Row Level Calculations (Functions)` | 32 | 195 | **64** |
| `Section 12 - Aggregate Calculations` | 2 | 171 | 88 |

A root budget of **62** means the *entire* path to the bundle — drive, user folder, everything — must
fit in 62 characters or the unit cannot be opened. That is not hypothetical; that estate exists now.

The failing shape is exactly as predicted — the deepest routine PBIR structure, with the unit name
appearing **twice**:

```
<unit>\<unit>.Report\definition\pages\<page-id>\visuals\<visual-id>\visual.json
```

A 48-char workbook name spends **97 characters** on those two segments alone.

### 4. Contradiction: I could not reproduce "it crashes rather than failing gracefully"

Reported honestly rather than repeated. Across **two independent runs** of the A/B, Desktop showed a
modal **"Issues were found"** dialog that **names the offending `visual.json`** and states the rule —
an *attributable* error, not a silent crash. Afterwards: **no Frown window on any process**, and **no
crash artifact** under the Store app's local cache.

Most likely origin of the Frown: a `Stop-Process -Force` on an instance holding that modal, or
telemetry raised for the *handled* `FilePathTooLongError`.

This **lowers one part** of the severity — a customer hitting the limit does get told which file is
at fault — and lowers **no other part**: the bundle still cannot be opened. I'd rather flag the
disagreement than let an overstated severity claim into the issue.

### 5. Both ceilings now printed in both framings

The issue is written in "fails at 260/248"; the constants are "longest legal 259/247". Printing both
removes any off-by-one argument:

```
ceilings              : file <= 259 (Desktop refuses 260), directory <= 247 (refuses 248)
host LongPathsEnabled : 1 (NON-DEFAULT) - does NOT affect Power BI Desktop, which enforces its
                        own limit in managed code
```

That second line is the single sentence whose absence produced the wrong conclusion in #234.

### Mutation table: 22/22

All 20 from the first commit still caught, plus the two new reporting behaviours:

| mutation | named test that failed |
|---|---|
| registry line stops saying Desktop is unaffected | `test_registry_value_is_reported_so_a_reader_is_never_fooled` |
| first-refused framing dropped from ceilings line | `test_both_ceilings_are_reported_separately_and_unambiguously` |

`FILE_CEILING`/`DIR_CEILING` mutations now each fail **two** named tests, since the reporting
assertion is independent of the constants assertion.

### Still unverified

* **Which of the two guards fires** — not separable in a real PBIR tree.
* **The classic per-machine installer.** Everything was measured against **2.157.828.0, the
  MSIX / Microsoft Store package** (`...\Power BI Desktop Store App\...`). `EnsureNotLong` is managed
  code shared by both, so a difference would be surprising — but surprising is not measured, and MSIX
  is already flagged elsewhere in our docs as unresolved.
* **Desktop with `LongPathsEnabled = 0`** — cannot be *better* (the guard never consults the
  registry), but not measured.
* **The exact `EnsureNotLong` comparison operator** — boundary established behaviourally, source not
  read.
* **Whether the archive step has a lower budget.** 282 remains one anecdote; not encoded.

### Gates (exit codes)

`ruff format --check` 0 · `ruff check` 0 · `pylint scripts` 0 (10.00/10) · `check_navigation_index` 0
· `check_agent_capabilities` 0 · targeted pytest 0 (31 passed, 1 skipped) · `pytest -q` 1 with **only**
the 3 pre-existing `test_upstream_repro_pins.py` failures.

### Cleanup

All three Desktop instances closed by **literal PID** (`0` PBIDesktop, `0` msmdsrv remaining; no
orphan was left by the earlier crash — checked before assuming). Both >259-char fixture trees removed
with `\\?\`-prefixed `rmtree`; longest path remaining in the worktree is **239**. `git status
--porcelain` clean.

Guust-Franssens:
## Update 2 — the estate run is now the headline, and git is a **second** failing consumer with a worse failure mode than reported

### 1. Your numbers reconcile exactly — and mine were never in conflict

| metric | your figure | mine | note |
|---|---:|---:|---|
| longest full path | 287 | **287** | ✅ |
| files over 260 | 80 | **80** | ✅ |
| files over 282 | 66 | **66** | ✅ |
| longest relative | 196 | **196** | (my "tail 197" = 196 + separator) |
| checkout root | 90 | **90** | ✅ |

My earlier **183** is the same tree measured against *Desktop's* ceilings: **82 files > 259 + 101
directories > 247 = 183**. The extra 103 are the **directory rule**, which a file-only measurement
cannot see — and §3 below is why that matters more than it looked.

**3 of 51 units** are already over as they sit; worst root budgets **62 / 64 / 88**.

Agreed on the driver, and I've named it structurally rather than as a long-name accident: the unit
name is **duplicated** in `pbip\<NAME>\<NAME>.Report\`, so it scales with *every* unit. A 48-char unit
name spends **97 characters** there before any content; our own prefix is ~12, so nothing we control
can rescue it.

### 2. git failure reproduced — with a control

`core.longpaths` is unset in **every** scope here (git's default is `false`). Faithful copy of the 3
offending units at the same 90-char root — 179 files, exactly 80 over 260:

| `core.longpaths` | result |
|---|---|
| **unset (default)** | 74 × `warning: could not open directory ...: Filename too long`, then `error: unable to index file`, `fatal: adding files failed` → **exit 128, 0 of 179 staged** |
| **`true`** (control) | 0 warnings → **exit 0, 179 of 179 staged** |

Length is the sole variable; the config is the sole difference.

### 3. ⚠️ The part that wasn't in the report: git's failure is **not always loud**

The two modes differ by **which** path is overlong:

| overlong thing | git |
|---|---|
| a **file**, parent directory readable | `error: unable to index file` → **FATAL**, exit 128 |
| a **directory** | `warning: could not open directory` → **SKIPPED**, contents never seen |

In the second case git never sees the files inside, so nothing reports them missing. Synthetic tree
with one 265-char directory:

```
git add -A             -> EXIT 0     (1 of 2 files staged)
git commit             -> EXIT 0
git ls-tree -r HEAD    -> README.txt          <- the file in the long dir is GONE
git status --porcelain -> EXIT 0, prints NOTHING
```

**A green, silent, content-missing commit.** Its only trace is a `warning:` on stderr — discarded by
any script that redirects stderr.

This is more than a curiosity: until now the **247 directory ceiling rested on Desktop's stated rule
alone**, because an isolating fixture is impossible in a PBIR tree (I flagged that as unverified in my
last update). This is a **second consumer, from a different codebase, with a reproduction**,
confirming that directories are their own failure surface. The directory rule is no longer inferred.

### 4. Both opt-ins now reported

```
host LongPathsEnabled  : 1 (NON-DEFAULT) - does NOT affect Power BI Desktop, which enforces
                         its own limit in managed code
host git core.longpaths: false (git default) - `git add` on an over-ceiling tree either aborts
                         (exit 128) or, when the OVERLONG path is a DIRECTORY, silently drops
                         its contents and exits 0
```

Neither reaches the verdict. `git config --get` exiting 1 is treated as git's **documented default
(false)**; git being *absent* is `None` — *"unknown, do not read as enabled"*. Those two are
deliberately distinct, and both are mutation-covered.

**Three consumers, three answers, one machine** — which is exactly why "it works here" was never evidence:

| consumer | tolerates 287 chars? | fixable by config? |
|---|---|---|
| Python 3.6+ (our generator) | **yes** | — |
| Power BI Desktop | **NO** | **no — nothing helps** |
| git (`core.longpaths` unset) | **NO** | yes |

### 5. Ceiling reconsidered — justified, per your point 4

**The absolute ceiling stays blocking.** It is not hypothetical: where this bundle sits, git breaks
*today* and Desktop refuses it, and this repo's own build root is 90 characters. "Where it sits" is a
real place where real tools run.

**But you're right that it's the weaker half**, so the portable number is promoted from opt-in to
always-printed: any bundle whose `root_budget = 259 − tail` falls below **40** now prints
`TIGHT ROOT BUDGET`. `--min-root-budget N` still turns it into a hard gate.

40 is **derived, not invented**: `C:\Users\<name>\Documents\` is ~28 characters before the customer
creates a single folder.

One piece of arithmetic worth recording, because it defines the blind spot precisely:
`root_budget = ceiling − tail`, and a tree is clean only when `ceiling ≥ root_len + tail` — therefore
**a clean tree always has `root_budget ≥ root_len`.** The advisory can only fire on a **short-rooted**
build: a CI runner, a container, `C:\b\`. That is exactly the automated case where nobody is watching,
and it is why a purely absolute gate was insufficient.

### Mutation: 30/30 — and one survivor was a real defect

New this pass: 8 mutations covering the git reader, the git reporting, and the root-budget advisory.

**`verdict inherits git core.longpaths` SURVIVED the first run.** Making the `unknown` branch depend
on `read_git_long_paths` is **inert on this machine** (`core.longpaths` is unset, so the added
condition is always true) but is a **live defect on any machine that has set it to `true`** — the "it
works here" trap this entire issue is about. I had proved the *registry* value cannot reach the
verdict and never proved the same for git on that branch. Closed by
`test_git_long_paths_cannot_rescue_an_unknown_path_either`.

### Cleanup note worth passing on

`shutil.rmtree` with the `\\?\` prefix is **not** enough to delete a long-path tree containing a git
repo — git object files are read-only, so it also needs an `onexc` chmod handler. Documented in the
docs page. Longest path remaining in the worktree: **239**. No Desktop launched this pass.

### Gates (exit codes)

`ruff format --check` 0 · `ruff check` 0 · `pylint scripts` 0 (10.00/10) · `check_navigation_index` 0
· `check_agent_capabilities` 0 · targeted pytest 0 (45 passed, 1 skipped) · `pytest -q` 1 with **only**
the 3 pre-existing `test_upstream_repro_pins.py` failures.

### Still unverified

Which of Desktop's two guards fires · the classic per-machine installer (all Desktop measurement was
**MSIX/Store 2.157.828.0**) · Desktop at `LongPathsEnabled = 0` · the exact `EnsureNotLong` operator ·
whether the archive step has its own lower budget (**282 is still not encoded**, though the estate now
has **66 files past it**).

Guust-Franssens:
## All three findings fixed. The Medium is a **retraction** — I reproduced it against the assembly before accepting it.

### MEDIUM — I was wrong about the mechanism, and I retract it

I loaded `Microsoft.PowerBI.Packaging.dll` 2.157.828.0 and invoked the method directly, independently
reproducing your result:

```
PBIProjectUtils.EnsureNotLong(string path, bool isFolder)

FILE 258 ALLOWED  259 ALLOWED  260 ALLOWED  261 THREW PathTooLongException
DIR  246 ALLOWED  247 ALLOWED  248 ALLOWED  249 THREW PathTooLongException
```

**It compares with `>`.** 260 and 248 are allowed, so its message states *intent*, not the comparison
— and the end-to-end refusal at file 260 / dir 248 **did not come from that guard**.

Most consistent explanation, now marked **⚠️ inferred, not measured**: the .NET/Win32 `MAX_PATH` limit
applying because `PBIDesktop.exe` carries no `longPathAware` manifest entry — which is this branch's
**original** framing, before `EnsureNotLong` was over-read into it. My second commit over-corrected on
the strength of a stack frame. `EnsureNotLong` *is* real and does fire: the crash report that named it
came from a **268**-char path, and 268 > 260.

`FILE_CEILING = 259` / `DIR_CEILING = 247` unchanged — they follow the **observed** boundary and are
deliberately one tighter than the assembly's inclusive limits. Corrected everywhere: module docstring,
constant comments, docs page, and the rendered line now reads `observed refusal at 260`. "Still
unverified" now states plainly that **neither ceiling is independently established** and that 247 is
conservative (`EnsureNotLong` tolerates 248, git tolerates 260).

### HIGH 1 — UTF-16 code units, and surrogates → unknown

Both halves fixed. All absolute lengths **and tails** now use
`len(s.encode("utf-16-le", "strict")) // 2`. Strict encoding **raises** on a lone surrogate, which is
caught and recorded as `unknown` — never clean; `_safe_repr` keeps `--json` serialisable.

You're right that this is the same defect class again — *cannot represent it → treat it as fine*.
Covered by a monkeypatched surrogate test that runs everywhere **plus** a real
undecodable-POSIX-filename test (Linux only, skipped on Windows with a reason).

### HIGH 2 — root budget is now the minimum across both ceilings

```python
root_budget = min(_ceiling_for(record, limits) - record["tail"] for record in measured)
```

Plus the report now **names the binding path**, so the constraint is attributable rather than a bare
number. Regression test uses exactly the blank-PBIR-page shape and asserts the 2-character
overstatement, and `--min-root-budget <overstated>` now exits **1**.

### Re-measured after the UTF-16 change

Estate figures are **unchanged** — 183 over ceiling, longest 287, root budget 62 — because that tree
has no astral characters. But the binding path is now correctly attributed to a `visuals\<id>`
**directory** at tail 185 (`247 − 185 = 62`), which equals `259 − 197` *only* because `\visual.json` is
12 characters and the two ceilings also differ by 12. `examples` still exits **0**.

### Gates

| gate | exit |
|---|---|
| astral path over UTF-16 ceiling but not code-point ceiling | **1** ✅ |
| `surrogateescape`d POSIX filename | **unknown**, never clean ✅ |
| blank-page shape with `--min-root-budget <overstated>` | **1** ✅ |
| `root_budget` = min across both ceilings, UTF-16 units | ✅ |
| docs/comments corrected: observed boundary ≠ assembly's inclusive limits | ✅ |
| `examples` exit 0 | ✅ |
| estate re-reported | ✅ (unchanged; binding path now attributed correctly) |

`ruff format --check` 0 · `ruff check` 0 · `pylint scripts` 0 (10.00/10) · `check_navigation_index` 0 ·
`check_agent_capabilities` 0 · targeted pytest 0 (**52 passed, 2 skipped**) · local full suite
**2,600 passed** with only the 3 pre-existing `test_upstream_repro_pins` failures · **CI green on both
jobs**.

### Mutation: 33/33

New: `UTF-16 → code points`, `surrogates encodable → clean`, `root budget back to file-only`,
`binding record dropped`, `ceilings line stops saying "observed"` — each with a named failing test.
Scoring unchanged: mutation asserted on disk first, credited only on a genuine `N failed`; any other
pytest exit scores `INVALID`.

### Two things I did not smooth over

1. **Warning count.** I counted **74** `Filename too long` warnings; you counted 73 on the same shape.
   I could not reconcile it and have listed it under "still unverified" rather than quietly adopting
   either number.
2. **A 20-minute local suite stall** during this pass was **not** my change. I traced it rather than
   waiting: nested `pytest` processes whose parent chain resolved to **another agent's** shell (PID
   25656) running the full suite concurrently. The nesting itself is `test_skills.py`'s normal
   subprocess pattern. A clean re-run finished in 13.2 min.

Net effect, as you said: the boundary correction **strengthens** the PR — an inferred mechanism is
replaced by a measured one, and the gate now rests on the observed end-to-end boundary rather than on
any single implementation's comparison.

PR #395, opened 2026-08-30
fix(check_path_ceiling): write --json before rendering, and degrade an unencodable console

Guust-Franssens:
Follow-up to #389 (review finding, non-blocking). No issue was filed, so this carries a plain
reference — `Refs #235` — and no closing keyword.

## The defect

`_safe_repr` protected *unknown* paths, but measured paths were rendered raw. Reproduced on master
before fixing, with a real combining-character filename (`e` + U+0301):

```
PYTHONIOENCODING=cp1252
python scripts/check_path_ceiling.py <dir> --json out.json
UnicodeEncodeError: 'charmap' codec can't encode character '\u0301'
CHECK_EXIT=1        JSON_EXISTS=False
```

The path itself measured fine (70 UTF-16 units) and passed under `--quiet`, so this is purely an
**output-path** defect. Two things make it worth fixing rather than noting:

- **`--json` is the machine-readable contract and produced no artifact** — exactly the case an
  automated consumer depends on. A CI job asking for JSON gets an exit code and nothing to read.
- The crash exits **1**, which is *also* the "findings" code — so a consumer cannot distinguish a
  crash from a real over-ceiling verdict.

## The fixes

**1. Structural — write the JSON before any console output.** No rendering failure of *any* kind can
now destroy the requested artifact, not just encoding ones.

**2. Boundary — `_make_stdout_lossy()` reconfigures stdout with `errors="backslashreplace"`.**
Escaping paths inside `render` would also work, but would mangle every non-ASCII path on the UTF-8
terminals where they display perfectly well. Degrading only at the point of output keeps readable
consoles readable and unreadable ones merely ugly. Best-effort: a substituted stdout (`StringIO`,
`redirect_stdout`) has no `reconfigure`, and that is a no-op, not an error.

After: `CHECK_EXIT=0`, `JSON_EXISTS=True`, and the console prints the name with `\u0301` escaped.

## Tests — two angles, deliberately

- **`test_json_is_written_before_the_console_is_touched`** makes `render` raise, proving the
  **ordering** deterministically and identically on every OS — no hostile console needed. This is
  the *only* test that catches Fix 1 in isolation, because with Fix 2 in place the console no longer
  crashes and the artifact survives either ordering. Fix 1 is defence in depth, and is tested as
  such rather than assumed.
- **Two subprocess tests with `PYTHONIOENCODING=cp1252`** (the pattern already used by
  `tests/test_scripts_help.py`) exercise the real end-to-end shape, asserting **both** that the JSON
  exists **and** that the exit code is unchanged — `EXIT_OK` for a clean tree and `EXIT_OVER_CEILING`
  for a finding, so a degraded console cannot quietly alter a verdict.

⚠️ **The tests do not become the hazard they test.** The combining character appears only in a
fixture's **leaf filename**, never in a directory the test itself must print, and the `--json` output
path is ASCII.

⚠️ **Platform asymmetry handled explicitly.** They run on **both** Windows and Linux — the codec ships
everywhere and the name is valid on both filesystems. `_make_combining_fixture` returns `None` and
the test **skips with a reason** if a filesystem will not store the name unchanged, rather than
passing vacuously. Verified running (not skipping) on Windows; CI covers Linux.

## Mutation table — 5/5

Each scored only on a genuine `N failed` in a **named** test, with the mutation asserted on disk first.

| mutation | named test that failed |
|---|---|
| FIX 1 reverted: JSON written **after** rendering | `test_json_is_written_before_the_console_is_touched` |
| FIX 2 removed: stdout left strict | `test_cp1252_console_still_produces_json_and_the_expected_exit_code` |
| FIX 2 neutered: `strict` instead of `backslashreplace` | `test_make_stdout_lossy_sets_backslashreplace_on_a_real_stream` |
| FIX 2 lies: reports reconfigured when it did not | `test_stdout_without_reconfigure_is_a_no_op_not_an_error` |
| FIX 2 raises instead of degrading | `test_stdout_without_reconfigure_is_a_no_op_not_an_error` |

**One of my assumptions was wrong and the tests caught it:** I expected pytest's captured stdout to
lack `reconfigure`. It does not — pytest's default fd capture supplies a real `TextIOWrapper` — so the
fallback is now exercised with an explicit `io.StringIO` stand-in instead of relying on the harness to
provide one.

## Gates (exit codes)

`ruff format --check` 0 · `ruff check` 0 · `pylint scripts` 0 (10.00/10) · `check_navigation_index` 0 ·
`check_agent_capabilities` 0 · `pytest -q tests/test_check_path_ceiling.py` 0 (**57 passed, 2 skipped**).

## Scope

`scripts/check_path_ceiling.py`, its tests, and one paragraph in `docs/windows-path-limits.md`
recording the `--json` ordering contract. Nothing else.

### DIFF
--- a/docs/windows-path-limits.md
+++ b/docs/windows-path-limits.md
@@ -16,6 +16,12 @@ python scripts/check_path_ceiling.py <bundle> [--json report.json] [--min-root-b
 
 Exit `0` clean · `1` findings · `2` usage · `3` could not evaluate.
 
+`--json` is a **machine-readable contract**: the artifact is written **before** anything is printed,
+so a console that cannot encode a path — a Windows cp1252 terminal meeting a filename with a
+combining character — degrades the *display* (`errors="backslashreplace"`) instead of destroying the
+*output*. Before that ordering was fixed, such a run exited 1 with **no file written at all**, and
+exit 1 is also the "findings" code, so a consumer could not tell a crash from a real finding.
+
 ---
 
 ## 1. Three consumers, three different answers, one machine
--- a/docs/windows-path-limits.md
+++ b/docs/windows-path-limits.md
@@ -18,9 +18,18 @@ Exit `0` clean · `1` findings · `2` usage · `3` could not evaluate.
 
 `--json` is a **machine-readable contract**: the artifact is written **before** anything is printed,
 so a console that cannot encode a path — a Windows cp1252 terminal meeting a filename with a
-combining character — degrades the *display* (`errors="backslashreplace"`) instead of destroying the
-*output*. Before that ordering was fixed, such a run exited 1 with **no file written at all**, and
-exit 1 is also the "findings" code, so a consumer could not tell a crash from a real finding.
+combining character — degrades the *display* instead of destroying the *output*. Before that ordering
+was fixed, such a run exited 1 with **no file written at all**, and exit 1 is also the "findings"
+code, so a consumer could not tell a crash from a real finding.
+
+Degrading is **local to each write** (`text.encode(..., "backslashreplace")` at the point of output).
+The check never calls `sys.stdout.reconfigure()`: that fixed the crash but permanently mutated the
+**caller's shared stream** — measured, a clean `--quiet` run returned 0 and left an unrelated
+caller's later output escaped — and it also could not help a restricted stream that has no
+`reconfigure` at all, which still raised and exited 1. Note that
+`codecs.getwriter("cp1252")(...)` exposes **no `.encoding` attribute**, so a rewrite keyed off
+`stream.encoding` silently does nothing for exactly the stream that needs it; escaping falls back to
+pure ASCII, which every codec accepts.
 
 ---
 
--- a/scripts/check_path_ceiling.py
+++ b/scripts/check_path_ceiling.py
@@ -526,6 +526,28 @@ def _nonneg(raw: str) -> int:
     return value
 
 
+def _make_stdout_lossy() -> bool:
+    """Let `print` degrade an unencodable path instead of raising. Returns True if reconfigured.
+
+    Every path in the report comes from the filesystem, so it can contain anything the filesystem
+    allows - including characters this console's codec cannot represent. On a Windows cp1252 console
+    a legal filename carrying a combining character (`e` + U+0301) made `print(render(report))` raise
+    `UnicodeEncodeError` and take the whole run down.
+
+    Escaping the paths inside `render` would fix that too, but it would mangle every non-ASCII path
+    on the UTF-8 terminals where they display perfectly well. Degrading only at the point of output
+    keeps readable consoles readable and unreadable ones merely ugly.
+
+    Not every stdout is a real stream - pytest's capture and `contextlib.redirect_stdout` replace it
+    with objects that have no `reconfigure` - so the failure to reconfigure is not an error.
+    """
+    try:
+        sys.stdout.reconfigure(errors="backslashreplace")  # type: ignore[union-attr]
+    except (AttributeError, ValueError, OSError):
+        return False
+    return True
+
+
 def main(argv: list[str] | None = None) -> int:
     """CLI entry point."""
     parser = argparse.ArgumentParser(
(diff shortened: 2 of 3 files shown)


Guust-Franssens:
## Fixed — and you were right that it landed on the thing this PR exists to remove

Both halves reproduced before fixing.

**A — the caller's shared stream was mutated, on every invocation:**
```
before=surrogateescape  after=backslashreplace  MUTATED=True
main_rc=0  isatty=False        (--quiet and non-TTY too)
```

**B — and the converse was worse.** A restricted stream with no `reconfigure` fell through to
`return False`, and `main()` still raised `UnicodeEncodeError` and exited **1** — indistinguishable
from a finding.

### The vacuity is mine, and it's the same family one layer down

`io.StringIO` accepts every `str`, so the fallback test **could only ever prove the happy path**.
I caught the ordering interaction proactively this round and then walked straight into a fixture that
cannot exhibit the condition it is named for. `_strict_cp1252_stream()` now carries a docstring
saying exactly why `StringIO` was worthless there, so it does not come back.

### The fix — lossiness local to each write, `sys.stdout` never touched

```python
def _emit(text, stream):
    try:
        print(text, file=stream)
    except UnicodeEncodeError:
        print(_console_safe(text, stream), file=stream)
```

Closes both halves at once: no side effect on the caller, and no dependence on `reconfigure`
existing. The happy path is untouched, so a UTF-8 terminal still prints paths verbatim.

**One measurement changed the design.** `codecs.getwriter("cp1252")(...)` exposes **no `.encoding`
attribute** (only `.errors`):

```
codecs.StreamWriter .encoding -> <ABSENT>    .errors -> 'strict'    .reconfigure -> <ABSENT>
io.StringIO         .encoding -> None        .errors -> None
```

So the obvious `stream.encoding`-keyed rewrite would silently no-op on exactly the stream that needs
it. `_console_safe` falls back to **pure ASCII**, which every codec accepts — and a mutation covers
that specific trap.

### Also fixed, same defect, not in your review

`ERROR: not a directory: {...}` interpolates **caller-supplied paths** and used a raw `print` to
stderr. Routed through `_emit`, with a test.

### After

```
A: rc(--quiet)=0  rc(full)=0   before=surrogateescape  after=surrogateescape   MUTATED=False
B: clean   -> rc=0  json=True  printed 'OK: ...'
   finding -> rc=1  json=True  printed 'OVER_CEILING: ...'
```

### Tests added

| test | covers |
|---|---|
| `test_stdout_errors_is_unchanged_after_main` | A, including `--quiet` and non-TTY |
| `test_strict_unreconfigurable_stream_preserves_the_exit_code` | B, parametrized **clean** and **finding** |
| `test_stream_without_an_encoding_attribute_still_prints` | `StringIO` keeps its honest role: the no-op path |
| `test_console_safe_escapes_to_ascii_when_the_stream_hides_its_encoding` | the missing-`.encoding` trap |
| `test_console_safe_uses_the_stream_encoding_when_it_is_declared` | doesn't over-escape when it *can* encode |
| `test_usage_error_on_an_unencodable_path_does_not_crash` | the stderr path |

The ordering test is **kept unchanged** — as you said, it remains the only thing isolating the
JSON-before-render fix, because with local degrading in place the console no longer crashes and the
artifact survives either ordering.

### Mutation: 7/7

Each scored on a genuine `N failed` in a **named** test, mutation asserted on disk first. Includes an
explicit **regression mutation** that reinstates the global `stdout` mutation — caught by
`test_stdout_errors_is_unchanged_after_main`.

⚠️ **Stated plainly: this 7/7 is my own, unverified by a second party.** You did not perform the
on-disk ordering mutation on the previous round, so **no second-party mutation score is claimed for
this PR** — for either the earlier 5/5 or this 7/7.

### Gates (exit codes)

`ruff format --check` 0 · `ruff check` 0 · `pylint scripts` 0 (10.00/10) · `check_navigation_index` 0 ·
`check_agent_capabilities` 0 · `pytest -q tests/test_check_path_ceiling.py` 0 (**62 passed, 2 skipped**).

Pushed as `895c5fb`. Worktree clean.

PR #398, opened 2026-08-30
Wire check_path_ceiling into check_unit, so one command still answers 'is it done' (refs #235)

Guust-Franssens:
## What

`scripts/check_path_ceiling.py` shipped yesterday (#389, closing #235) as a real, **blocking** gate — and **nothing called it**. It was a nineteenth `check_*.py` an agent had to remember by name.

It now runs behind the existing `scripts/check_unit.py` facade, alongside stubs, empty models, blank placeholders, PBIR validity and the rest. **No new command** — fewer things to remember is the whole point.

### Scope: `all` only

The scan walks the **whole target tree**, so it cannot be attributed to a layer: a `--scope model` run would be judging report paths, and `--scope report` would be judging model paths. That is exactly the cross-layer confusion *"own your layer"* exists to prevent, and it is why `integration` is also wrong — that scope means *cross-artifact coherence* (report reference vs model field), not "a property of the whole tree".

It joins `engine-receipt` and `desktop-orphans` in `ALL_ONLY_CHECK_IDS`, whose existing members are all whole-unit/orchestrator properties. Bonus: `_omitted_checks()` now names `path-ceiling` in every scoped run, so a scoped PASS still cannot look complete.

### Gating, not advisory

The native gate already exits 1. Demoting it inside the facade would make `check_unit` **the one place a bundle Power BI Desktop cannot open reads as done** — a fail-open, which is the shape `_run_cli_gate`'s tests exist to prevent.

The brief's counter-argument — a breach is not fixable by the persona that hits it (install-root length + engine-side name duplication) — argues for **routing, not silence**. The row carries an `orchestrator` owner hint. The same "not my layer" argument would silence `desktop-orphans` and `engine-receipt`, both of which are gating here.

The genuinely strong advisory argument is **host-relativity**: the verdict depends on where the repo is checked out. That is not hidden, it is stated — see below.

### Target: the unit/bundle dir

As every other gate takes. Verified on both shapes:

| target | result |
|---|---|
| `examples/shipping-kpis` (unit) | 58 paths, root budget 153, `PASS` |
| `_runs/estate-2.339.0-20260829` (bundle) | 183 over ceiling, longest 287, root budget 62, exit 1 |
| `examples` / `migrations` (whole trees) | `ok`, exit 0 |

Real shipping units pass with room to spare, so gating does not paint the estate red.

## Making the verdict judgeable

The facade runs every gate with `--quiet`, which suppresses exactly the numbers that let a reader tell a fragile bundle from a deep checkout. `_annotate_path_ceiling` carries them into the row:

```
path-ceiling: FINDINGS (native over_ceiling exit 1)
    suspected owner: orchestrator (install root length + engine-side name duplication; not a layer defect)
    detail: 2 of 9 paths over ceiling; longest 313; root budget 10 at root length 66
            (measured against this checkout root; a shorter install root may pass)
    rerun: ... check_path_ceiling.py ...
    finding: kind=file; evidence=...\pages\pppp...\page.json
```

- `root_budget` is the **portable** number — the longest install root the tree tolerates.
- A payload with **no census** is left un-annotated: synthesizing *"0 of 0 paths over ceiling"* for a scan that never ran would read as reassurance.
- `worst_offenders` added to `_payload_findings`' walk keys so the breach **names its evidence** (only `check_path_ceiling` emits that key).

That output is from a **real end-to-end run** against a genuinely over-ceiling tree (313-char path), not a unit test — `check_unit` exit **1**.

## Mutation table — 12/12 caught, judged by exit code

| # | mutation | named test that failed |
|---|---|---|
| M1 | unregister from `ALL_ONLY_CHECK_IDS` | `test_path_ceiling_is_registered_all_scope_only` |
| M2 | move gate to `REPORT` scope | `test_path_ceiling_is_registered_all_scope_only` |
| M3 | **demote `over_ceiling` to advisory** | `test_path_ceiling_over_ceiling_is_a_finding_not_a_pass` |
| M4 | uppercase the native pass status | `test_path_ceiling_clean_scan_is_a_pass` |
| M5 | drop the native skip statuses | `test_path_ceiling_gate_names_the_native_skip_statuses` |
| M6 | remove the annotation hook | `test_path_ceiling_runs_end_to_end_on_a_real_unit` |
| M7 | drop `root_budget` from the annotation | `test_path_ceiling_detail_carries_root_budget_and_says_it_is_host_relative` |
| M8 | annotation overwrites an existing diagnostic | `test_path_ceiling_detail_preserves_an_existing_diagnostic` |
| M9 | stop rendering `worst_offenders` | `test_path_ceiling_finding_names_the_offending_paths` |
| M10 | misroute the owner hint to a builder | `test_path_ceiling_owner_is_the_orchestrator_not_a_builder` |
| M11 | hide `path-ceiling` from omitted checks | `test_external_resolves_scope_model_matches_golden` (golden) |
| M12 | synthesize a census when none ran | `test_path_ceiling_annotation_invents_no_census_when_the_scan_never_ran` |

**12/12 is suspicious by default**, so the harness was itself controlled. Only pytest **exit 1** scored as caught, and each row shows `1 failed, 62 deselected` — a collected test that *asserted* false. Two negative controls prove the harness can report NOT CAUGHT:

- syntax error in `check_unit.py` → pytest exit **2** (collection error) → scored `NOT CAUGHT (2)`
- `-k` matching nothing → pytest exit **5** → scored `NOT CAUGHT (5)`

That is precisely the false-positive shape (`non-zero == caught`) this repo has been bitten by before.

## Gates (exit codes)

| gate | exit |
|---|---|
| `pytest -q -n auto --dist loadfile` | 1 — **3 failures, all `test_upstream_repro_pins.py` engine pins** (pre-existing tripwires). 2999 passed, 9 skipped |
| `pylint scripts` | **0** (10.00/10) |
| `pylint .github/skills/pbip-model-refresh/scripts` | **0** (10.00/10) |
| `pylint .github/skills/powerbi-ai-readiness/scripts` | **0** (10.00/10) |
| `ruff format --check` / `ruff check` | **0** / **0** |
| `check_agent_capabilities.py` | **0** |

`tests/test_check_unit.py` alone: **63 passed, exit 0**.

Baseline before this branch was **4** failures — the 3 pins plus a flake in `.github/skills/pbip-model-refresh/tests/test_persist_atomicity.py`. That file is untouched here, failed a **different test name** in each of two runs, and passes in isolation (38 passed, exit 0). It is a parallel-execution flake in a sibling's file, not a regression.

## What I could NOT verify

- **The 259/247 ceilings themselves.** Inherited from #389; no Power BI Desktop re-measurement was done here. This PR is wiring.
- **Behaviour on a checkout deep enough to make a shipping unit breach.** Everything was measured at root length 44 (`C:\Users\gfranssens\vscode-projects\t2p-occam`). The gating decision means a sufficiently deep checkout would turn healthy units red; the `root_budget` annotation is the mitigation, but **that scenario was not exercised**.
- **Linux CI path behaviour.** The over-ceiling proof used a 313-char path created on Windows with `LongPathsEnabled=1`. The committed tests deliberately avoid long paths (they tighten `--ceiling` instead), so they are portable — but no long-path tree was built on Linux.
- **Whether `worst_offenders` in the walk keys affects a future gate.** Verified today that only `check_path_ceiling.py` emits that key; a future gate adopting the name would start rendering findings from it.
- **`no_paths` / `unknown_paths` mapping against the real scanner.** Those are pinned by a *contract* assertion (M5) plus monkeypatched runs, not by making the real script emit them — exit 3 alone already yields NOT_CHECKED, so the status strings are belt-and-braces. **Unobserved in this corpus**, not unobservable.
- **Runtime cost at estate scale.** `check_path_ceiling` took 1.1–3.5 s on the trees measured, well inside the facade's 300 s per-gate timeout, but a very large bundle was not timed through `check_unit`.

refs #235 (already closed by #389 — this is the wiring follow-up, so no closing keyword).


Guust-Franssens:
## Review finding accepted — and it is worse than reported

Reproduced independently before fixing, on a **byte-identical** tree:

| root length | status | `root_budget` | `root_budget_is_tight` |
|---|---|---|---|
| 65 | `ok`, exit 0 | 79 | **False** |
| 94 | `over_ceiling`, exit 1 | 79 | False |

Console output at root 65, before the fix:

```
path-ceiling: PASS (native ok exit 0)
```

That is the whole row. `_render_actionable_detail` renders `detail` for non-clean rows only, so the host-relative annotation was invisible on **exactly the row nobody reads twice**. Confirmed at `check_unit.py:1566`.

## Pushing back on the suggested trigger — with that measurement

The review proposed showing the warning *"at minimum when `root_budget_is_tight` is true"*. **On the very tree that motivated the finding, that flag is `False`** (79 ≥ the advisory 40). A tight-only trigger would have printed nothing — and nothing for the 70-budget hypothetical either.

So the budget is rendered **unconditionally**:

1. **For this gate the budget *is* the result.** PASS means "it fits *here*"; the budget is the only portable number in the row. Suppressing it makes the row say less than the scan found. That is unlike `stub-measures`, where PASS genuinely means "nothing to say".
2. **A threshold tuned for "alarming" cannot also mean "safe to relocate"** — measured above.
3. **A conditional number is unreadable in its absence.** Nothing printed cannot be told from: a comfortable budget / a dropped annotation (mutation M6 proves that can happen silently) / a gate that never ran. A number always present is self-verifying.

Cost is one clause, one row, one gate, `all` scope only — the other 18 gates are untouched. `TIGHT` is appended when the advisory *does* fire, so it keeps its teeth.

## What it prints now

```
path-ceiling: PASS (native ok exit 0; root budget 153 - breaches above a 153-char installation root)
path-ceiling: FINDINGS (native over_ceiling exit 1; root budget 10 TIGHT - breaches above a 10-char installation root)
    suspected owner: orchestrator (install root length + engine-side name duplication; not a layer defect)
    detail: 2 of 9 paths over ceiling; longest 313; root budget 10 at root length 66 - needs an
            installation root of at most 10 characters and this one is 66, so a shorter installation
            root may pass
```

Rendered at **every** status, through the same per-gate headline mechanism `stub-measures`, `connection-fidelity` and `scaffold-partitions` already use — not a new mechanism.

**Wording is now status-aware and states the breach direction** (second half of the finding):

| status | says |
|---|---|
| `ok` | *"fits here with N characters of headroom: any installation root **LONGER** than N characters **WILL** breach"* |
| `over_ceiling` | *"needs an installation root of at most N characters and this one is X, so a shorter installation root may pass"* |
| budget < 0 | *"**NO** installation root can hold this tree"* — relocation cannot help, so it does not advise it |

The old text said only that a *shorter* root may pass: vacuous on a pass, and it left the actual risk unstated.

## Incidental: `render()` hit pylint R0912

Adding the row took `render()` to **13/12 branches**. Note this printed **10.00/10 while exiting 8** — the "perfect score, non-zero exit" trap; caught by judging on exit code. Fixed by **extraction**, not a waiver: the per-check headline chain moved to `_render_check_headline`. It is a refactor, not a rewrite — **`tests/golden/` is byte-identical** (`git diff --stat tests/golden/` → empty). The extracted function carries a `too-many-return-statements` waiver *with a stated reason*, matching `_run_cli_gate` in the same file.

## Mutation table — 22/22 caught

All 12 original wiring mutations were **re-run against the restructured `render()`**, plus 10 new ones.

| # | mutation | named test that failed |
|---|---|---|
| M1–M12 | (original wiring: unregister, wrong scope, demote to advisory, uppercase status, drop skip statuses, remove hook, drop root_budget, overwrite diagnostic, drop worst_offenders, misroute owner, hide from omitted, synthesize census) | all re-verified, all caught |
| **R1** | **restore the PASS suppression** | `test_path_ceiling_render_shows_the_budget_on_a_passing_row` |
| **R2** | **show budget only when `root_budget_is_tight`** (the suggested trigger) | `test_path_ceiling_render_shows_the_budget_even_when_not_tight` |
| R3 | restore suppression, judged by the **real CLI** | `test_path_ceiling_runs_end_to_end_on_a_real_unit` |
| R4 | drop the `TIGHT` escalation marker | `test_path_ceiling_tight_budget_is_escalated_in_the_headline` |
| R5 | drop the breach direction from the headline | `test_path_ceiling_pass_row_headline_carries_the_root_budget` |
| R6 | make wording status-blind | `test_path_ceiling_wording_is_status_aware` |
| R7 | negative budget falls through (detail branch) | `test_path_ceiling_impossible_budget_says_no_root_can_hold_it` |
| R8 | negative budget falls through (headline branch) | `test_path_ceiling_impossible_budget_says_no_root_can_hold_it` |
| R9 | invent a note when there is no budget | `test_path_ceiling_headline_is_silent_when_there_is_no_budget` |
| R10 | relocation boundary no longer bounded by budget | `test_path_ceiling_pass_then_relocate_breaches_and_the_pass_said_so` |

**Only pytest exit 1 scored as caught**; every row shows `1 failed, 71 deselected` — a collected test that *asserted* false. Negative controls (re-confirmed): syntax break → exit **2**, unmatched `-k` → exit **5**, neither scored.

The reviewer's reproduction is now **committed** as `test_path_ceiling_pass_then_relocate_breaches_and_the_pass_said_so`. Both the tree depth *and* the relocation distance are derived from a calibration scan rather than hard-coded, so it holds at any temp-directory depth (measured: 107 chars on this Windows runner, ~50 on Linux) without encoding the scanner's ceiling arithmetic. Overshooting the budget by exactly one keeps the binding path one unit over its *own* ceiling, so nothing it creates needs Windows long-path support.

## Regression net (exit codes)

| gate | exit |
|---|---|
| `check_path_ceiling examples` | **0** |
| `check_path_ceiling migrations` | **0** |
| `check_path_ceiling _runs/estate-2.339.0-20260829` | **1** (over ceiling, as expected) |
| `check_unit examples/shipping-kpis --scope all` | `path-ceiling: PASS (… root budget 153 …)` |
| full suite `-n auto --dist loadfile` | 1 — **3 failures, all instructed engine pins**; 3008 passed, 9 skipped |
| `tests/test_check_unit.py` | **0** (72 passed) |
| `tests/test_repo_layout.py` | **0** (41 passed) |
| `pylint scripts` | **0** (10.00/10) |
| `pylint .github/skills/pbip-model-refresh/scripts` | **0** |
| `pylint .github/skills/powerbi-ai-readiness/scripts` | **0** |
| `ruff format --check` / `ruff check` | **0** / **0** |

**`_runs/` was not copied because it was not mutated** — only `check_path_ceiling` (a read-only `os.walk` + string arithmetic) was pointed at it, never `check_unit`. Proven rather than asserted, with a before/after census: **2548 files, 14 234 506 bytes, newest mtime 2026-08-29 20:37:03 — identical on both sides of the scan.**

## What I could NOT verify

- **A shipping unit at a checkout deep enough to breach.** Still measured at root length 44. The new PASS line is precisely the mitigation for this, but the red-estate scenario itself remains unexercised.
- **Whether 40 (`SHIPPING_ROOT_BUDGET_ADVISORY`) is the right escalation threshold.** I argued it is the wrong *gate* for showing the number and left the constant alone — it is `check_path_ceiling`'s, and re-deriving it is out of scope here.
- **Linux CI long-path behaviour.** The committed relocation test deliberately stays under 260 characters; no long-path tree was built on Linux.
- **The 259/247 ceilings**, still inherited from #389.
- **Whether the extra headline clause is unwanted noise in day-to-day use.** That is a judgement only repeated real runs settle; the argument for always-on is stated above and is the falsifiable part of this change.

Issue #367, opened 2026-08-29
Test-BlockingDialog reports ANY window >=100x100 as a credential block - a Refresh progress dialog trips it, and a false credential wall is a hard stop


Guust-Franssens:
## User story

> **As an** engineer running several Power BI Desktop refreshes concurrently,
> **I want** the credential probe to distinguish a live progress dialog from a genuine blocking modal,
> **so that** a slow warehouse cold-start is not reported as a credential wall — a verdict this
> toolkit treats as a hard stop.

## The mechanism

The probe has **two** detectors, and only one of them looks at what the window says.

**`Test-CredentialModal` — signature-based, precise.** It matches descendant window text against
`credential_modal_signature.regex`:

```
You aren.t signed in|Personal Access Token|Databricks Client Credentials|
specify how to connect|Account Key|Enter your credentials|Please specify how to connect
```

**`Test-BlockingDialog` — size only** (`.github/skills/pbip-model-refresh/scripts/probe_desktop_credential.ps1`):

```powershell
function Test-BlockingDialog {
  foreach ($w in Get-PidWindows) {
    if ($w.ClassName.StartsWith('WindowsForms10.Window.8')) { continue }
    if ($w.Width -lt 100 -or $w.Height -lt 100) { continue }
    return $w        # <- ANY window >= 100x100 of any other class
  }
  return $null
}
```

No signature. No title. No modality test. **Any** window at least 100x100 whose class is not
`WindowsForms10.Window.8` is returned as a blocking dialog. Power BI's own Refresh progress dialog
satisfies that trivially.

(The same 100x100 floor exists in the Python arbiter — `_credential_modal.py:153-154`,
`MIN_DIALOG_WIDTH/HEIGHT = 100`, applied at `:189` — but there it is only a pre-filter ahead of
`match_credential_modal`, which *does* require a signature hit. The PowerShell blocking path has no
such second stage.)

## Field report, 2026-08-28

Running three concurrent refreshes, the probe returned `BLOCKED_BY_DIALOG` for one unit. Rather than
accept the verdict, they screenshotted the window:

> *"we screenshotted the actual window directly and found it was Power BI's own ordinary Refresh
> progress dialog, stalled behind a slow Snowflake warehouse cold-start while 3 sibling refreshes ran
> concurrently — not a credential or native-query approval wall at all. Canceled the stale duplicate
> refresh and re-ran cleanly."*

Their own summary of the defect:

[...]

PR #400, opened 2026-08-30
Classify the dialog instead of measuring it, in the Python detector too

### LINKED ISSUE #376: inspect_credential_modal is size-only too - and unlike #367's detector it feeds probe_desktop_query, the gate of record
## The defect

`inspect_credential_modal` decides a blocking dialog **by size alone**, exactly as `Test-BlockingDialog` did before #367:

```python
if candidates:
    return CredentialDetection(blocking_dialog=candidates[0])
```

There is no second signature stage. A window large enough to be a candidate becomes a `blocking_dialog` finding regardless of what it says — so a **Refresh progress dialog** is reported the same as a credential prompt.

## Why this is worse than #367

#367 fixed the PowerShell detector. This one **feeds `probe_desktop_query.py` — the gate of record.**

The repo's documented position is that the modal-watch probe can false-positive (serverless cold start), which is precisely why the one-row DATA probe is authoritative. If the arbiter that probe consults is itself size-only, that safeguard is weaker than the docs claim.

A false credential verdict is a **hard stop**: it halts a migration and demands a human. There is a recorded incident of an agent sitting on an unresponsive external system for **129 minutes / 298 tool calls** before anyone realised what was happening, and #341 is an open issue about this exact misdiagnosis class.

## Found by

Blind-review scope discipline during #367. The agent fixing #367 identified this while checking its callers, declared it **out of its declared scope**, left it unchanged, and documented that decision in `SKILL.md` rather than silently widening its diff. That was the right call — and it is why this needs its own issue rather than being absorbed.

⚠️ Note the #367 issue text asserts *"the Python arbiter is fine because it has a second signature stage."* **That is wrong** — verified by reading the code. Do not rely on that sentence.

## What good looks like

#367 landed a five-verdict model in PowerShell that this should mirror:

| verdict | meaning |
|---|---|
| `CREDENTIAL_MISSING` | credential signature matched — the only hard stop |
| `CREDENTIAL_PRESENT` | ran to deadline, nothing unclassifiable up |
| `REFRESH_IN_PROGRESS` | progress dialog already up at t=0 |
| `DIALOG_UNRECOGNIZED` | we **read** it; matched neither signature |
| `DIALOG_UNREADABLE` | no readable text at all — **absent ≠ empty** |

The last two matter: this repo has three separate open issues about gates that collapsed *"we looked and found nothing"* into *"we never looked"*.

**Failure direction:** err away from declaring a hard stop. A false positive exits non-zero and stops everything; a false negative lands on `CREDENTIAL_PRESENT`, which the repo already treats as untrustworthy on its own. One direction has a backstop, the other does not.

## Acceptance

1. A progress dialog is not reported as a blocking credential dialog
2. A genuine credential prompt still is — do not fix the false positive by breaking the true positive
3. An unreadable window is a distinct third verdict, never silently one of the other two
4. Mutation-test each: restore the size-only verdict, remove the credential check, collapse unreadable into unrecognised. Assert each mutation is present on disk before trusting a "caught" result
5. Check every caller before changing a verdict name — #367 verified `_verdict_lines.py` and `refresh_pbip_model.py` before retiring `BLOCKED_BY_DIALOG`

## Related

#367 (PowerShell half, landing) · #341 (stale `localSettings.json` misread as a credential wall — same misdiagnosis class)

Guust-Franssens:
Fixes #376 — `inspect_credential_modal` decided "blocking dialog" by **size alone**, exactly as `Test-BlockingDialog` did before #367, and on the path that feeds `probe_desktop_query.py` — **the gate of record**.

## What it keyed on, before and after

Measured by driving the real detector against synthesised windows. Every one is **702x355 and non-main**, so size cannot separate them — only text can.

| window | before | after |
|---|---|---|
| `Refresh` + `Evaluating...` + `Cancel` | `BLOCKED_BY_DIALOG`, **exit 1** | `REFRESH_IN_PROGRESS`, exit 3 (and **ignored** while our own operation is in flight) |
| `Please specify how to connect` | `CREDENTIAL_MISSING`, exit 1 | unchanged |
| native-database-query approval | `BLOCKED_BY_DIALOG`, **exit 1** | `DIALOG_NEEDS_HUMAN`, exit 3 |
| progress text **beside** that prompt | `BLOCKED_BY_DIALOG`, exit 1 | `DIALOG_NEEDS_HUMAN`, exit 3 |
| progress text beside unexplained prose | `BLOCKED_BY_DIALOG`, exit 1 | `DIALOG_UNRECOGNIZED`, exit 3 |
| caption `Refresh`, no content (the WPF shape) | `BLOCKED_BY_DIALOG`, exit 1 | `DIALOG_UNREADABLE`, exit 3 |
| no text at all | `BLOCKED_BY_DIALOG`, exit 1 | `DIALOG_UNREADABLE`, exit 3 |
| `Save changes?` / `Discard` | `BLOCKED_BY_DIALOG`, exit 1 | `DIALOG_UNRECOGNIZED`, exit 3 |
| credential text in an **80x60** window | **NO FINDING AT ALL** | `CREDENTIAL_MISSING`, exit 1 |

## The half that was not in the issue

That last row is a **second instance of the same defect, in the opposite direction**. `match_credential_modal` was fed only `blocking_dialog_candidates(...)`, so the 100x100 filter gated the **hard stop** as well as the classification: a credential prompt in a smaller window produced **no finding at all** — a silent false negative on the one verdict that matters most. `Test-CredentialModal` in the arbiter has always scanned every window, any class, any size; the two now agree.

## Consumer trace (established, not assumed)

`probe_desktop_query._credential_verdict` → `refresh_pbip_model.main` / `_refresh_and_save` → the printed verdict line → `probe_live_source._classify_failure`.

`BLOCKED_BY_DIALOG` is in `_verdict_lines.CREDENTIAL_STOP_VERDICT_RE`, so **before**, any dialog ≥100x100 produced `NO_CREDENTIAL` — whose directive is *"You may NOT build. You CANNOT fix this yourself… TERMINATE THE RUN NOW."* **After**, each new token falls through to **`ERROR`** ("the probe itself could not run"): `rc != 0`, `_lift_gate` is never reached, the gate stays armed, and nothing asserts a wall we did not observe. Verified end-to-end by driving the real emitters through the real classifier.

**I deliberately did not change `scripts/_verdict_lines.py`.** Adding the new tokens to the credential-stop family would map them back to `NO_CREDENTIAL` — the exact false claim being removed; `ERROR` is the honest verdict and already fails safe. Divergence stated rather than hidden.

## Is a third state needed? Yes — and a fourth

`unreadable` (we could not read it) is kept distinct from `unrecognized` (we read it, it matched nothing), from `benign` (positively read as progress), and from `credential`. `benign-title-only` — a reassuring **caption** with no content behind it — joins the **unreadable** band, not the benign one: a caption is not content, and Win32 child-HWND enumeration sees nothing inside a WPF dialog, so a caption is frequently all there is.

## Three deliberate divergences from #390 (not cargo-culted)

Win32 child-HWND text is strictly **less** evidence than a UIA harvest, so three arbiter mechanisms are **not** ported, each for a stated reason:

| arbiter mechanism | here | why |
|---|---|---|
| prose **join** before the credential match | not ported | the join needs a control-type signal to skip an interposed `Cancel`; Win32 has none, so the only join available is the naive whole-window one #390 discarded. A join can *manufacture* a phrase (`Account` + `Key` → `Account Key`) — and that error lands on the one verdict the issue says to err **away** from. Recall lost this way routes to `unrecognized`/`unreadable` (exit 3, loud), and the arbiter is the escalation path. |
| enabled-owner **exoneration** | not ported | a *suppression* path needing owner/enabled state this module does not harvest, unverifiable without a live Desktop. Omitting it costs one more exit 3. |
| `benign-unverified` (truncated harvest) | not needed | a text read that throws already fails the **whole** enumeration → `unknown_reason`. A partial read never reaches the classifier. |

Everything the two detectors *do* share — signature files, kind names, verdict tokens, precedence — is now literally the same vocabulary.

## One asymmetry, in both directions

A proven-benign progress dialog is **reported at t=0** (`REFRESH_IN_PROGRESS` — somebody else owns this instance; stacking a second refresh is what the 2026-08-28 field report had to unpick) and **ignored once our own operation is in flight**. Nothing else is ever ignored.

Correspondingly, in the **bounded** refresh wait a dialog finding is now **latched**, not raised, so it cannot abort a refresh that may still finish — while `probe_desktop_query`'s poll still **acts** on it, because that loop has no deadline of its own and a latched dialog behind a wedged query would wait for ever.

## Mutation table — 16/16 caught

Scored on a **named `N failed`** plus non-zero exit, never a bare exit code. That guard earned itself: M14's first form produced a `SyntaxError` (pytest exit 2) and was correctly refused as a catch rather than counted — the #390 harness bug, reproduced and avoided.

| # | mutation | result | named test that failed |
|---|---|---|---|
| 1 | restore the SIZE-ONLY verdict | CAUGHT | `test_poll_loop_ignores_our_own_refresh_progress_dialog` (+13) |
| 2 | remove the credential signature check | CAUGHT | `test_classify_dialog_reads_the_window_instead_of_measuring_it[credential]` |
| 3 | collapse UNREADABLE into UNRECOGNIZED | CAUGHT | `test_unreadable_owned_dialog_reports_unreadable_not_no_modal` (+7) |
| 4 | let a benign CAPTION dismiss the window | CAUGHT | `test_classify_dialog_...[benign-title-only]` |
| 5 | re-gate the credential scan behind 100x100 | CAUGHT | `test_the_credential_scan_is_not_gated_by_the_size_filter` |
| 6 | first benign element classifies the window | CAUGHT | `test_classify_dialog_...[mixed-content]` |
| 7 | always dismiss a benign dialog | CAUGHT | `test_probe_query_does_not_stop_for_power_bi_s_own_progress_dialog` |
| 8 | never dismiss a benign dialog | CAUGHT | `test_poll_loop_ignores_our_own_refresh_progress_dialog` |
| 9 | restore raise-on-dialog mid-wait | CAUGHT | `test_poll_loop_latches_unreadable_dialog_and_raises_at_the_deadline` |
| 10 | benign outranks the unreadable band | CAUGHT | `test_a_window_we_could_not_account_for_outranks_a_progress_dialog` |
| 11 | smuggle a CREDENTIAL_MARKER into guidance | CAUGHT | `test_a_dialog_finding_keeps_the_gate_shut_without_claiming_a_credential_wall` |
| 12 | in-flight detector exists but is not the default | CAUGHT | `test_poll_loop_default_detector_is_the_in_flight_one` |
| 13 | `probe_desktop_query` returns exit 1 again | CAUGHT | `test_probe_query_stops_at_exit_3_for_a_dialog_it_could_not_read` |
| 14 | poll stops using the in-flight detector | CAUGHT | `test_probe_query_poll_ignores_a_progress_dialog_while_the_query_runs` |
| 15 | `refresh_pbip_model` returns exit 1 again | CAUGHT | `test_refresh_and_save_wires_a_dialog_finding_to_exit_3` |
| 16 | `main()` stops reporting a t=0 dialog | CAUGHT | `test_refresh_main_reports_a_t0_dialog_at_exit_3_before_any_mutation` |

## Callers checked before renaming (issue acceptance 5)

`BLOCKED_BY_DIALOG` / `BlockingDialog` / `DialogBlockedError` retired from the Python path. Every consumer verified: both entry points, `join_with_credential_poll`, `_verdict_lines.py` (unchanged — the token stays in the regex, now unemitted), `tests/test_probe_live_source_verdict.py`, `SKILL.md`, `docs/data-source-credentials.md`, `powerbi-semantic-model-gotchas/SKILL.md`. The stale hand-written `BLOCKED_BY_DIALOG` fixture in `test_probe_live_source_verdict.py` is replaced by one **harvested from the real emitter** — the #152/#153 discipline.

## Gates

`pytest -q -n auto --dist loadfile` → **3009 passed**, 9 skipped, only the 3 known `test_upstream_repro_pins.py` failures · `-m timing` 7 passed · pylint **10.00/10, exit 0** on all three roots · `ruff` clean · `check_navigation_index.py` 0 · `tests/test_skills.py` 27 passed · `sync_installed_skills.py --check` 0.

## What I could NOT verify

- **No Power BI Desktop was used.** Every window in the corpus is synthesised. The `MIN_PROMPT_WORDS = 5` backstop and `benign_dialog_signature.regex` remain **inferred** from the refresh UI, not captured from a live dialog.
- **In practice a real WPF refresh dialog will most often land on `DIALOG_UNREADABLE`, not `REFRESH_IN_PROGRESS`** — Win32 child enumeration returns nothing inside WPF, so only the caption survives, and a caption cannot dismiss. That is exit 3 either way (the false *hard stop* is gone), but it is a recall limit, not a clean win, and I am not claiming otherwise. Capturing a real dialog's Win32 text is the follow-up that would settle it.
- **The `Authentication required` alternative** in `blocking_prompt_signature.regex` carries a `CREDENTIAL_MARKER`, so its evidence excerpt re-classifies as `NO_CREDENTIAL` downstream. It fails toward "a human is needed", never toward a clear — same known limit #390 recorded.
- **`benign` is reachable only when a dialog exposes real child-HWND content** (WinForms dialogs do). Unobserved in this corpus whether Power BI's own dialog does.
- **Shadowing did not apply here:** `pbip-model-refresh` is deliberately **held back** from the published plugin (`scripts/build_plugin.py:37-47`), so no plugin copy of these scripts exists to go stale. The sync still mattered for `powerbi-semantic-model-gotchas/SKILL.md`, which does ship; that file was copied and `--check` now exits 0.
- **Pre-existing, untouched:** `scripts/README.md:60` carries a leftover merge-conflict marker `>>>>>>> d8ef6c6`. Not mine, not in scope, flagged here rather than silently fixed.

### DIFF
--- a/.github/skills/pbip-model-refresh/SKILL.md
+++ b/.github/skills/pbip-model-refresh/SKILL.md
@@ -365,9 +365,27 @@ stay byte-identical.
 >
 > | exclusion | the claim it makes |
 > |---|---|
-> | `main_frame(...)` | it is the application, and the thing dialogs block. Identified by **direct ownership** first (an owned window names its owner), falling back to the `MainWindowHandle` convention — first visible unowned window — only when nothing is owned, i.e. when there are no dialogs to hide. |
+> | `main_frame(...)` | it is the application, and the thing dialogs block. Identified by **transitive ownership** — every owned window is walked to its **unowned root** — falling back to the `MainWindowHandle` convention only when *nothing* is owned. **Ambiguity fails closed**: two or more roots, an unresolvable chain, or several candidate unowned windows all return `None`, which excludes nothing. |
 > | `is_proven_non_blocking(...)` | an **enabled owner** proves this window blocks nothing. One-way: a *disabled* owner never convicts (Power BI's own refresh dialog disables it too) and `None` — no owner — means the test **did not apply**, which is not the same as passing it. |
 > | `renders_nothing(...)` | **unowned AND zero-area**: no owner to disable *and* no pixels to display. Both conjuncts are required; either alone is one of the dead proxies. |
+>
+> ⚠️ **Round 4: "first owner" was itself a proxy for "the root", and it hid a credential modal.** The
+> traversal originally stopped at the first ownership edge. An owned window can own another popup, so a
+> Z-order of **`tooltip → credential dialog → frame`** made the *credential dialog* the frame — and the
+> frame is excluded from the prepass **and** from classification, so the modal vanished. Three native
+> reproductions, all missing it:
+>
+> | construction | before |
+> |---|---|
+> | nested chain `tooltip → credential dialog → frame` | frame = the `#32770` credential dialog; **no modal** |
+> | unowned `Internet Explorer_Hidden` ahead of the real frame, reading `Enter your credentials` | selected as the frame; prepass skipped it; **no modal** |
+> | that misidentified dialog owning an *enabled* tooltip, frame busy with benign refresh content, `operation_in_flight=True` | **no modal, no dialog finding, no unknown state** |
+>
+> **Two rules came out of it, and both are load-bearing.** Follow ownership *transitively*; and where
+> identity is **ambiguous, fail closed** — return `None` and exclude nothing. The cost of failing
+> closed is a loud exit 3; the cost of guessing is a real modal disappearing. That is why the unowned
+> fallback requires **exactly one** rendering unowned window, and why it is not reached at all once any
+> window is owned (its premise is that there are no dialogs).
 >
 > The arbiter's one-way enabled-owner exoneration is therefore **ported**, not skipped — that
 > divergence note in the module docstring is gone.
--- a/.github/skills/pbip-model-refresh/SKILL.md
+++ b/.github/skills/pbip-model-refresh/SKILL.md
@@ -318,10 +318,52 @@ stay byte-identical.
 > in `probe_desktop_query`'s poll it is **acted on**, because that loop has no deadline of its own and
 > a latched dialog behind a wedged query would wait for ever.
 >
-> **Downstream, honestly:** `scripts/_verdict_lines.py` does not recognise the new tokens, so
-> `probe_live_source` classifies them **`ERROR`** ("the probe itself could not run") — `rc != 0`, the
-> gate stays armed, and nothing claims a credential wall we never observed. That is the intended
-> outcome and it was verified end-to-end, not assumed.
+> **Downstream, honestly:** `probe_live_source` now recognises the dialog tokens **structurally**
+> (`scripts/_verdict_lines.py`'s `DIALOG_VERDICT_RE` → `classify_child_verdict`) and maps them to
+> **`ERROR`** — *"the probe itself could not run"*. `rc != 0`, the gate stays armed, and nothing claims
+> a credential wall we never observed.
+>
+> ⚠️ **That structural step is load-bearing, not tidiness.** Blind review of PR #400 measured what
+> happened without it: the parent fell through to `CREDENTIAL_MARKERS`, an **unanchored scan of the
+> whole transcript**, so `DIALOG_NEEDS_HUMAN` quoting its own evidence excerpt `Authentication
+> required` — an alternative that genuinely lives in `blocking_prompt_signature.regex` — was relabelled
+> `NO_CREDENTIAL`. The parent contradicted the child on a single word and fired *"a human must sign in;
+> terminate the run"*. Adding the tokens to the **credential-stop** family instead would have been the
+> other wrong answer: it asserts the very wall the child says it did not see.
+
+> ⚠️ **Blind review of PR #400 found the same defect class INSIDE the fix, four more times. Read this
+> before touching the classifier — three of the four were "we could not establish it" quietly becoming
+> "clean" again.**
+>
+> | # | what collapsed into the clean bucket | fix |
+> |---|---|---|
+> | 5 | **A length heuristic stood in for evidence.** `MIN_PROMPT_WORDS = 5` excused every unmatched element under five words, so `Refresh` + `Evaluating...` + **`Please enter your password`** classified `benign` and was **suppressed entirely** in flight. `Password:` (two words) too. Neither matches `credential_modal_signature.regex`, so the prepass did not rescue them. **In that shape the fix was worse than the bug** — the repo's rule is that a credential modal is never worked around. | The amnesty is **gone**. Any content element that is not recognised progress status vetoes dismissal, unless it is in the **enumerated** `benign_chrome_signature.regex` (`Cancel`/`OK`/`Close`) — a positive claim about specific strings that carry no prompt, not a claim about their size. |
+> | 3 | **A geometry threshold stood in for harmlessness — the original defect, surviving inside its own fix.** `dialog_candidates` still rejected everything under 100x100, and the all-window credential prepass only rescued *known-signature* text. Measured: a visible **80x60** unreadable owned window beside a normal main window returned `modal=None, dialog=None, unknown_reason=None` — byte-identical to a healthy Desktop. | **No size test at all.** Every visible non-main window is classified. Two exclusions remain and both are positive claims: **zero-area** windows (they rasterise nothing, so they show a human nothing) and the named `HELPER_WINDOW_CLASSES`. |
+> | 4 | **A report title fabricated a hard stop.** The all-window prepass read the Desktop **main** window, caption and children, so a report legitimately named `Account Key` / `Personal Access Token` / `Databricks Client Credentials` produced `CREDENTIAL_MISSING` at exit 1. | The **identified main window** is excluded from the prepass — and only it. Every real dialog is still scanned at every size and in every class, so an unusual modal class stays detectable. ⚠️ Helper windows are excluded from *classification* but **not** from the prepass: Power BI's AAD sign-in renders in the `Internet Explorer_Hidden` web view, so credential text really can appear there. |
+> | 1 | **Production bypassed the semantics the tests were checking.** `_join_refresh_worker` overrode `join_with_credential_poll`'s in-flight default with the t=0 `_credential_state`, the progress-monitor branch called the t=0 **raise** helper, and the final timeout check repeated it — so a proven-benign progress dialog **stopped the refresh it belonged to**. Measured in both branches: `DialogFoundError(REFRESH_IN_PROGRESS)` on the first poll, with the detector receiving **no `in_flight` argument at all**. The existing test asserted the *helper's* default identity, which production overrode: a test passing for the wrong reason. | Both wait branches and the final check use `_in_flight_credential_state`, and the progress branch latches through the shared `raise_latched_verdict` instead of raising at t=0 semantics. Its latches used to be computed and **discarded**, so it always degraded to a bare `TimeoutError`. The test now asserts on **production's call site** — the argument the detector actually receives. |
+>
+> **The through-line, and the rule that comes out of it:** a fix that adds a *suppression* path has to
+> make its positive claim actually positive. "Short", "big enough", "we read something", "the caption
+> looked fine" are all the same mistake wearing different clothes.
+
+> ⚠️ **The PowerShell arbiter still has finding 5's hole — measured, filed as #406, deliberately NOT
+> fixed here.** `probe_desktop_credential.ps1` keeps `$MinPromptWords = 5`, and driving its shipped
+> classifiers through the test harness shows the identical result: `Refresh` + `Evaluating` +
+> `Please enter your password` → `benign` → `REFRESH_IN_PROGRESS`, and **suppressed to `$null`** under
+> `-RefreshInFlight`; `Password:` likewise. It was left alone on purpose, following the precedent that
+> created issue #376 itself: #367's author found this defect in the Python half, declared it out of
+> scope, and filed it rather than silently widening the diff. Removing the amnesty there also reverses
+> a reviewed decision — `test_short_data_labels_beside_progress_text_do_not_block_suppression` exists
+> to keep `CREDENTIAL_PRESENT` reachable while Desktop shows its own refresh dialog — so it needs its
+> own issue and its own review, not a drive-by.
+>
+> **Reachability, stated rather than hidden.** Removing the amnesty costs the Python detector its
+> `benign` path whenever a dialog exposes a table name as child text (`Orders` is not progress status
+> and not chrome, so it vetoes). That is the sanctioned trade: `benign` is used only to avoid aborting
+> **our own** in-flight operation, so losing it costs extra **exit 3**s, never a silent clear.
+> ⚠️ Unobserved in this corpus whether Power BI's real refresh dialog exposes table names as child
+> HWNDs — no Desktop was available. If it does, expect `DIALOG_UNRECOGNIZED` where you hoped for
+> `REFRESH_IN_PROGRESS`; that is loud and recoverable, and it is the direction this bundle errs in.
 
 > ⚠️ **Do not wait blindly for `NO_BRIDGE` / `not_connected` — bound the bridge wait, then prove the
 > executable by PID.** Field report, 2026-08-18, two machines: a box with two Desktop versions
--- a/.github/skills/pbip-model-refresh/SKILL.md
+++ b/.github/skills/pbip-model-refresh/SKILL.md
@@ -346,8 +346,43 @@ stay byte-identical.
 > make its positive claim actually positive. "Short", "big enough", "we read something", "the caption
 > looked fine" are all the same mistake wearing different clothes.
 
-> ⚠️ **The PowerShell arbiter still has finding 5's hole — measured, filed as #406, deliberately NOT
-> fixed here.** `probe_desktop_credential.ps1` keeps `$MinPromptWords = 5`, and driving its shipped
+> ⚠️ **Round 3 killed three PROXIES at once — blocking is now decided from MODALITY. Read this before
+> adding any exclusion.** The round-2 fix replaced one unsound proxy (size) with three more — a class
+> prefix, zero area, and a helper-class name allowlist — and the reviewer defeated **all three** with
+> native Win32 experiments. They are the same mistake: answering *"is this window blocking a human?"*
+> by looking at something else.
+>
+> | proxy | native repro that defeated it | now |
+> |---|---|---|
+> | `Internet Explorer_Hidden` **name** allowlist | a visible **900×700** host reading `Password:` / `Sign in` / `Continue?` — or nothing at all — returned `modal=None, dialog=None, unknown=None`. Keeping it in the *prepass* only ever rescued **exact signature matches**. | no name is consulted anywhere |
+> | `WindowsForms10.Window.8` **class prefix** | that prefix names a WinForms **family**, not one HWND: an owner **and its owned `FixedDialog`** both reported the exact class `WindowsForms10.Window.8.app.0.2b2196a_r3_ad1`. A real credential dialog was removed from the prepass *and* from classification. | the frame is identified by **ownership** |
+> | **zero area** | a real `WS_VISIBLE` **owned 0×0** window built with `CreateWindowEx`, owner disabled: `owned-visible=True owner-win32-enabled=False rect=0x0`. Genuinely blocking, and on the unbounded query-poll path suppressing it is a false clear **that is also a hang**. | zero area suppresses only **conjoined with unowned** |
+>
+> **Win32 answers the question directly, so it is asked directly: a modal disables its owner.**
+> `DesktopWindow` now carries `owner_hwnd` (`GetWindow(GW_OWNER)`) and a **three-valued**
+> `owner_enabled` (`IsWindowEnabled(owner)`), and exactly three things are excluded from
+> classification, each a positive claim:
+>
+> | exclusion | the claim it makes |
+> |---|---|
+> | `main_frame(...)` | it is the application, and the thing dialogs block. Identified by **direct ownership** first (an owned window names its owner), falling back to the `MainWindowHandle` convention — first visible unowned window — only when nothing is owned, i.e. when there are no dialogs to hide. |
+> | `is_proven_non_blocking(...)` | an **enabled owner** proves this window blocks nothing. One-way: a *disabled* owner never convicts (Power BI's own refresh dialog disables it too) and `None` — no owner — means the test **did not apply**, which is not the same as passing it. |
+> | `renders_nothing(...)` | **unowned AND zero-area**: no owner to disable *and* no pixels to display. Both conjuncts are required; either alone is one of the dead proxies. |
+>
+> The arbiter's one-way enabled-owner exoneration is therefore **ported**, not skipped — that
+> divergence note in the module docstring is gone.
+>
+> ✅ **Proven against real windows, not just dataclasses.**
+> `test_the_win32_harvest_reads_real_ownership_and_owner_enabled_state` builds the reviewer's own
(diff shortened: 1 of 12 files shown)


Guust-Franssens:
## Verdict on the review: all five findings confirmed by experiment, none disputed

I reproduced every one against the PR-#400 build before changing anything. No push-back — the review
was right on all five, and finding 5 was worse than the summary suggested (`Password:`, two words,
does it too).

| # | reproduced on the PR-#400 build | now |
|---|---|---|
| 5 | `Refresh`+`Evaluating...`+`Please enter your password` → `benign`/`REFRESH_IN_PROGRESS`; `dialog_verdict(in_flight=True)` → **`None`** | `mixed-content`/`DIALOG_UNRECOGNIZED`, **never suppressed** |
| 3 | 80x60 unreadable owned window beside a normal main window → `modal=None, dialog=None, unknown_reason=None` — **identical to healthy** | `DIALOG_UNREADABLE` |
| 4 | main window titled `Account Key - Power BI Desktop` → `PREFLIGHT: CREDENTIAL_MISSING`, **exit 1** | no finding; the probe continues |
| 2 | `DIALOG_NEEDS_HUMAN` evidence `Authentication required` → parent `NO_CREDENTIAL` | parent `ERROR` |
| 1 | **both** wait branches → `DialogFoundError(REFRESH_IN_PROGRESS)` on the first poll; detector saw `['<not passed>', '<not passed>', '<not passed>']` | no exception; detector sees `[True, True, True]` |

## The through-line, taken literally

The class is *"a claim that sounds positive but is not"*. Each fix replaces the proxy with the real
thing rather than tightening it:

- **length → enumeration.** `MIN_PROMPT_WORDS` is deleted, not lowered. Any unrecognised content
  element vetoes dismissal unless it is in the new **`benign_chrome_signature.regex`**
  (`^&?Cancel$|^&?OK$|^&?Close$`) — a positive claim about specific strings that carry no prompt.
- **geometry → "it rasterises nothing".** No size test at all. The two remaining exclusions are
  zero-area windows and named `HELPER_WINDOW_CLASSES`, and the docstring on that list demands
  evidence for anything added to it, explicitly forbidding a restored threshold.
- **"scan everything" → "scan every DIALOG".** Only the identified main window leaves the credential
  prepass. Helper windows are excluded from *classification* but deliberately **kept in the prepass** —
  Power BI's AAD sign-in renders in the `Internet Explorer_Hidden` web view, so credential text really
  can appear there. That is now a named test.

## Finding 2 — you were right that my reasoning was sound and my conclusion was not

`_verdict_lines.py` gained `DIALOG_VERDICT_RE` + `classify_child_verdict`, checked alongside
`DESKTOP_GONE`/`DESKTOP_UNREADY` and **before every free-text scan**, mapping to `ERROR`. The
credential-stop family is untouched, and a test now asserts the dialog tokens are **not** in it — the
wrong answer I had correctly rejected is pinned as wrong.

⚠️ **Outside the bundle, for merge sequencing:** `scripts/_verdict_lines.py` and
`scripts/probe_live_source.py`. The three authoritative-verdict branches and the success gate moved
into `_verdict_lines.py` (`classify_child_verdict`, `_is_earned_success`) — a real seam (verdict-line
decisions there, free-text decisions in `_classify_failure`) that also bought back the
`max-module-lines` headroom the addition consumed: **1215 → 1180**, with `R0916` gone too. Three
re-exports (`_has_data_ok_verdict`, `_has_desktop_gone_verdict`, `_has_desktop_unready_verdict`) are
now unused *in* `probe_live_source` but load-bearing for the seam tests, so they carry an explicit
`noqa: F401` — `ruff --fix` deleted them once already.

## Mutation table — 17/17 caught

Scored on a **named `N failed`** plus non-zero exit. **Four escaped the first run, and every one was a
real gap, not an equivalent mutant** — the two that would have been easiest to wave through are M14
and M15:

| # | mutation | caught by |
|---|---|---|
| 1 | restore the `MIN_PROMPT_WORDS` amnesty | `test_short_prompt_beside_progress_text_is_never_dismissed[Please…]` (5 failed) |
| 2 | widen the chrome allowlist to `.*` | `test_classify_dialog_reads_the_window_instead_of_measuring_it[mixed-content]` |
| 3 | drop the chrome allowlist (benign unreachable) | `test_poll_loop_ignores_our_own_refresh_progress_dialog` |
| 4 | restore the 100x100 candidate filter | `test_a_small_dialog_is_still_classified[80-60-unreadable]` |
| 5 | exclude empty-title windows (a new suppression path) | `test_unreadable_owned_dialog_reports_unreadable_not_no_modal` |
| 6 | let the prepass read the MAIN window again | `test_a_report_title_cannot_fabricate_a_credential_prompt[Account Key]` |
| 7 | narrow the prepass to `dialog_candidates` | `test_the_credential_prepass_reads_windows_that_classification_skips` ⟵ *escaped first* |
| 8 | parent stops recognising dialog verdicts | `test_a_dialog_verdict_is_recognised_structurally_and_is_not_a_credential_stop` |
| 9 | dialog verdicts join the CREDENTIAL-STOP family | same test ⟵ *escaped first; needed the explicit "not in that family" assertion* |
| 10 | drop a token from the parent's regex (drift) | same test |
| 11 | a dialog verdict stops vetoing the success gate | `test_a_dialog_verdict_cannot_clear_the_live_source_gate` |
| 12 | no-trace branch reverts to the t=0 detector | `test_the_production_refresh_path_polls_with_the_in_flight_detector[False]` |
| 13 | progress branch reverts to the t=0 detector | same, `[True]` |
| 14 | progress branch reverts to the t=0 **raise** helper | `test_a_dialog_we_could_not_read_does_not_abort_a_refresh_that_completes[True]` ⟵ *escaped first* |
| 15 | progress branch discards its latches again | `test_a_latched_dialog_surfaces_at_the_deadline_in_both_wait_branches[True]` ⟵ *escaped twice* |
| 16 | final timeout check reverts to the t=0 helper | `…polls_with_the_in_flight_detector[False]` |
| 17 | `_in_flight_credential_state` stops passing `in_flight` | same |

**M15 is worth reading.** It escaped its *first replacement test* as well: `refresh()`'s own final
check re-observed the still-present dialog and raised anyway, so a **defence-in-depth path was masking
a missing latch**. The test only became load-bearing once the dialog *vanishes* before the deadline —
the #154 shape, where only a real latch can report it. That is the same "passing for the wrong reason"
failure you named in finding 1, one level down.

For finding 1 specifically, both required shapes exist and are distinct: M12–M17 break **production's
call site**, while the older `test_poll_loop_default_detector_is_the_in_flight_one` breaks the helper's
signature. Neither catches the other.

## Not fixed — measured, and filed as #406

**`probe_desktop_credential.ps1` has finding 5's hole too.** Driving its shipped classifiers through
the bundle's own harness:

| window content | t=0 | `-RefreshInFlight` |
|---|---|---|
| `Refresh`, `Evaluating`, `Please enter your password` | `benign` / `REFRESH_IN_PROGRESS` | **`$null` — suppressed** |
| `Refresh`, `Evaluating`, `Password:` | `benign` / `REFRESH_IN_PROGRESS` | **`$null` — suppressed** |
| `Refresh`, `Evaluating`, `Please enter your password now` | `mixed-content` | `DIALOG_UNRECOGNIZED` |

I left it, following the precedent that created #376: #367's author found this defect in the Python
half, declared it out of scope, and filed it. There is also a substantive reason — removing the
amnesty there reverses a reviewed decision
(`test_short_data_labels_beside_progress_text_do_not_block_suppression` keeps `CREDENTIAL_PRESENT`
reachable while Desktop shows its own refresh dialog) and needs its own review, with the live-WPF
`serial` tests re-run. #406 states the options rather than assuming mine.

## Gates

`pytest -q -n auto --dist loadfile` → **3033 passed**, 9 skipped, only the 3 known
`test_upstream_repro_pins.py` pins · pylint **10.00/10, exit 0** ×3 roots · ruff clean ·
`check_navigation_index.py` 0 · `sync_installed_skills.py --check` 0.

**On the sync, since you asked me to confirm rather than assume:** it was a **no-op this round**, and
that is verifiable rather than argued — `scripts/build_plugin.py`'s `SHIPPED_SKILLS` holds
`pbip-model-refresh` back from the published plugin, and no *shipped* bundle changed in this push. Last
round it did copy one file, because I had edited `powerbi-semantic-model-gotchas/SKILL.md`, which does
ship. `--check` exits 0 either way, so preflight will not block on `STALE`.

## What I could NOT verify

- **Still no Power BI Desktop.** Every window is synthesised.
- **Removing the length amnesty costs the Python `benign` path whenever a dialog exposes a table name
  as child text** — `Orders` is neither progress status nor chrome, so it vetoes. That is the trade
  you sanctioned ("fail to the exit-3 indeterminate band"), and it is cheap here because `benign` only
  avoids aborting *our own* in-flight operation. Whether Power BI's real refresh dialog exposes table
  names as child HWNDs is **unobserved in this corpus**.
- **Removing the size filter may classify Desktop-owned windows I have never seen.** If a healthy
  Desktop owns other visible, non-zero-area, non-main windows they will now report `DIALOG_UNREADABLE`
  (exit 3, loud). Unobserved in this corpus. The remedy is to identify them and add them to
  `HELPER_WINDOW_CLASSES` **with evidence** — the docstring says so, and a test asserts the size
  constants stay deleted.
- **Pre-existing, untouched:** `scripts/README.md:60` still carries a leftover `>>>>>>> d8ef6c6`
  conflict marker.

Guust-Franssens:
## Round 3: all three findings confirmed, none disputed — and the mechanism direction was right

Reproduced against the round-2 build before touching anything. Every one collapsed into the healthy state:

| # | round-2 build | now |
|---|---|---|
| 1 | `Internet Explorer_Hidden` 900×700 reading `Password:` / `Sign in` / `Continue?` / nothing → `modal=None, dialog=None, unknown=None` | `DIALOG_UNRECOGNIZED` / `DIALOG_UNRECOGNIZED` / `DIALOG_UNRECOGNIZED` / `DIALOG_UNREADABLE` |
| 2 | owned dialog sharing its owner's exact class, reading `Enter your credentials` → **nothing** | `CREDENTIAL_MISSING`; unreadable variant → `DIALOG_UNREADABLE` |
| 3 | visible **owned 0×0**, owner disabled → **nothing** | `DIALOG_UNREADABLE` |

**You were right that three patches was the wrong ask.** I did not write three; I deleted all three proxies and asked Win32 the question directly.

## One mechanism: a modal disables its owner

`DesktopWindow` now carries `owner_hwnd` (`GetWindow(GW_OWNER)`) and a three-valued `owner_enabled` (`IsWindowEnabled(owner)`). Exactly three exclusions remain, each a positive claim rather than a correlate:

| exclusion | the claim |
|---|---|
| `main_frame(...)` | **Direct ownership evidence first** — an owned window names its owner, so the frame identifies itself whenever a dialog is up. The `MainWindowHandle` convention (first visible unowned window) is a fallback reached only when *nothing* is owned, i.e. when there are no dialogs to hide. |
| `is_proven_non_blocking(...)` | An **enabled owner** proves this window blocks nothing. One-way: a *disabled* owner never convicts (Power BI's own refresh dialog disables it too), and `None` — no owner — means the test **did not apply**, which is not passing it. |
| `renders_nothing(...)` | **unowned AND zero-area.** Both conjuncts required. Zero area alone is finding 3; unowned alone would silently drop every unowned window. |

`MIN_DIALOG_WIDTH/HEIGHT`, `HELPER_WINDOW_CLASSES` and `is_desktop_main_window` are **deleted**, and a test asserts they stay deleted *and* that `dialog_candidates`' source consults neither `class_name` nor `width`. The arbiter's one-way enabled-owner exoneration is now **ported**, not skipped — that divergence note is gone from the module docstring.

## Proven against real windows, not dataclasses

You wrote native experiments; so does the suite now. `test_the_win32_harvest_reads_real_ownership_and_owner_enabled_state` builds your reproduction with `CreateWindowEx` **in the test process** and asserts the production harvest reads it:

```
frame hwnd=55248332: harvested=True  owner_hwnd=0
owned 0x0 hwnd=4064010: harvested=True  owner_hwnd=55248332  owner_enabled=False  rect=0x0
  -> renders_nothing=False   is_proven_non_blocking=False   classified=True
```

This was not optional cover: **every synthesised-window test survives a mutation that hard-codes `owner_hwnd = 0` or `owner_enabled = True`** (N8/N10 escaped until the native test existed). Only a native test can see the harvest.

⚠️ **A note for anyone extending it**, because it bit me and it bit silently: an untyped `GetModuleHandleW` truncates the 64-bit `HINSTANCE`, and `RegisterClassW` then faults — a `faulthandler` **access-violation dump on a test that still reported a pass**. Set ctypes `argtypes`/`restype` explicitly, which is the same rule `_configure_user32` already states in production. (I also cast `DefWindowProcW` directly as the wndproc rather than routing Windows' synchronous callbacks through Python.)

## Mutations 10/10 — four escaped first, all real gaps

| # | mutation | caught by |
|---|---|---|
| 1 | restore the class-prefix exclusion in the **prepass** | `test_an_owned_dialog_sharing_its_owners_class_is_still_a_credential_wall` |
| 2 | restore the class-prefix exclusion in the **candidate rule** | 5 failed |
| 3 | restore the helper-class **name** allowlist | `test_the_aad_host_is_classified_by_modality_not_by_its_name` (6 failed) |
| 4 | suppress on **zero area alone** | `test_a_small_dialog_is_still_classified[0-0-unreadable]` |
| 5 | restore the **100×100** filter | 7 failed |
| 6 | let `owner_enabled is None` exonerate | `test_an_enabled_owner_is_the_only_thing_that_exonerates_a_dialog` |
| 7 | `main_frame` ignores ownership, convention only | `test_the_frame_is_identified_by_ownership_before_any_convention` ⟵ *escaped first* |
| 8 | harvest stops reading `owner_enabled` | `test_the_win32_harvest_reads_real_ownership_and_owner_enabled_state` ⟵ *escaped first* |
| 9 | `renders_nothing` drops the **area** conjunct | `test_an_unowned_window_that_renders_pixels_is_still_classified` ⟵ *escaped first* |
| 10 | harvest stops reading ownership at all | the native test ⟵ *escaped first* |

**N7 is the instructive one.** My first version of that test passed under the mutation, because with ownership evidence disabled the convention loop *skips owned windows* and still returned the frame. It only becomes load-bearing with a **decoy unowned window enumerated ahead of the frame** — the case where the convention alone picks the decoy and the real frame is then classified as a dialog on every healthy poll.

## On the sync — reporting rather than syncing blind, as asked

`--check` came back **exit 1** in my worktree, and the cause is not the one you hit. Diffing the three copies:

- **installed vs my worktree: one hunk, at line 442 — my own round-1 edit**, i.e. exactly the change this PR makes to a shipped bundle.
- Your #388 sync is intact; nothing else differed.

So the drift was expected mid-PR: this branch edits a *published* bundle and had not republished it. I ran `sync_installed_skills.py` (1 file copied), and `--check` is now **exit 0**. Preflight will not block on `STALE`.

## What I could NOT verify

- **Still no Power BI Desktop.** The native test proves the *harvest*; it cannot prove what Power BI's own windows look like.
- **Whether Power BI's AAD host is owned.** It is classified either way now — owned via modality, unowned via `renders_nothing`'s area conjunct — so finding 1 is closed on both branches, but which branch reality takes is **unobserved in this corpus**.
- **Whether a healthy Desktop owns visible windows that will now classify as `DIALOG_UNREADABLE`.** The exoneration should catch them (a healthy Desktop's owner is enabled), which is strictly better than the name allowlist it replaced, but it is **unobserved in this corpus**. If it happens it is exit 3 — loud, and the direction this bundle errs in.
- **The `#406` PowerShell hole is still open** by your agreement; unchanged this round.
- **Nothing outside the bundle changed this round** — `scripts/_verdict_lines.py` and `scripts/probe_live_source.py` are untouched since round 2.
- **Pre-existing, untouched:** `scripts/README.md:60` still carries a `>>>>>>> d8ef6c6` conflict marker.

## Gates

`pytest -q -n auto --dist loadfile` → **3045 passed**, 9 skipped, only the 3 known engine pins · pylint **10.00/10, exit 0** ×3 roots · ruff clean · `check_navigation_index.py` 0 · `sync_installed_skills.py --check` **0**.

PR #401, opened 2026-08-30
fix: resolve a merge conflict that shipped to master, and gate the class

Guust-Franssens:
## What happened

`scripts/README.md` has carried a full merge-conflict block on **master** since `d8ef6c6` (PR #392):

\\\
<<<<<<< HEAD
...
=======
...
>>>>>>> d8ef6c6 (Fix leg-aware credential gate clears)
\\\

**Every gate stayed green.** `tests/test_repo_layout.py` asserts each tracked `scripts/` file is *listed* in the README — and a conflict block lists **both** sides, so the malformed file satisfied it. This is the same defect class the repo keeps surfacing (**malformed input collapsing into the clean bucket**), this time applied to CI itself.

Found by the `fix-376` agent as an incidental observation while working on something unrelated.

## Resolution

Both sides were newer at something, so it is a genuine both-sides merge, not a pick-one:

| entry | taken from | why |
|---|---|---|
| `connections_manifest.py` | **HEAD** | the four-refusals privacy text incl. git-ignored output (#377/#382) |
| `connection_target.py` | **d8ef6c6** | adds the `ogr`/`ogrdirect` spatial-connector aliases |

## The guard

`tests/test_no_conflict_markers.py` scans git-tracked files for the **opening and closing** markers.

Two deliberate design choices, both stated in the module docstring:
- **The bare `=======` separator is ignored** — it is a valid Markdown setext H1 underline, so flagging it would fire on ordinary prose. A conflict block always carries the unambiguous pair anyway.
- **Markers are built at runtime** from repeated characters, so the test file cannot flag itself.

Binary/unreadable files are skipped explicitly rather than silently dropped.

## Mutation evidence

| step | result |
|---|---|
| baseline | `1 passed`, exit 0 |
| marker injected into `docs/INDEX.md` | `1 failed`, exit 1 |
| reverted | `1 passed`, exit 0 |

Verification: ruff exit 0 · `test_repo_layout` + guard 42 passed · `check_navigation_index.py` exit 0.


Guust-Franssens:
Post-approval hardening, pushed as `c15e538` — strictly stricter than the approved version, flagged here for transparency.

After the approve I ran my own edge-case probe on the byte scanner. Five of six shapes behaved correctly; one did not:

| shape | detected |
|---|---|
| CRLF line endings | YES |
| marker on the very first line | YES |
| no trailing newline | YES |
| 200 KB binary + marker | YES |
| **lone CR (old-Mac) line endings** | **no** |
| marker mid-line (must NOT fire) | correctly no |

Splitting on `b"\n"` alone put the marker mid-"line" in a lone-CR file, where the line-start check could never see it. Git does not write lone-CR endings, so the practical risk is negligible — but *"we scan every byte, except…"* is exactly the unstated exception this guard exists to eliminate, and it would have sat there unreported.

`LINE_BREAK` now matches `\r\n | \r | \n`, CRLF first so it counts as **one** separator and reported line numbers stay correct for all three conventions.

Two tests added, both mutation-proven:

- reverting to `b"\n"`-only splitting fails `test_every_line_ending_convention_is_scanned`
- `test_a_marker_not_at_line_start_is_not_a_finding` pins the other direction, so prose quoting a marker cannot trip the guard

Full state: **7 passed**, ruff exit 0, `test_repo_layout` and `check_navigation_index` unchanged.

Issue #403, opened 2026-08-30
No reliable legible full-dashboard reference: the offline path is always 192x192 and often absent, the server path dies when data sources are not connected


Guust-Franssens:
A migrated Tableau **dashboard** becomes a PBIR page. To judge that page's fidelity an agent needs a picture of the source dashboard. **Neither of our two capture paths reliably provides a legible one**, and nothing in the output states which grade of evidence was actually obtained.

## Path A - offline \.twb\ thumbnails (\xtract_twb_thumbnails.py\)

Measured across all 45 workbook assets in the trial estate:

| | |
|---|---|
| dashboards in the estate | **18**, across 13 workbooks |
| dashboards **with** an embedded thumbnail | **8** |
| dashboards with **none** | **10** |
| thumbnail dimensions | **192x192, every single one** |

The 10 missing are all \Seed - *.twbx\. Thumbnails are written by Tableau **Desktop on save**, so programmatically generated workbooks carry none. That is a plausible shape for a real customer estate too, wherever workbooks are produced by tooling rather than authored by hand.

**Where a thumbnail does exist it IS a genuine full-page composite** - extracted \Dashboard 1\ from \global_superstores_db.twb\ and the title, a horizontal bar chart, a scatter plot, a histogram and the segment legend are all identifiable. Layout, zone count, chart types and relative arrangement are legible.

**What is not legible at 192x192:** any value, any axis label, any tick, fonts, and exact colour. For a dashboard composing 6 worksheets each sub-viz occupies roughly 60px.

⚠️ \scripts/README.md\ says of this script: *""At ~192px it is decisive for **mark count**, mark type, layering and spatial distribution.""* That claim was made for **worksheets**. It does not carry over to a dashboard, where 192px is shared across every zone - and the README does not distinguish the two cases.

## Path B - server REST \/views/{id}/image\ (\capture_tableau_oracle.py --images\)

High resolution, and it **does** cover dashboards (they are returned by \/views\ like any sheet). But it fails wholesale when the workbook's data sources are not connected on the site:

\\\
global_superstores_db -> 0/26 captured, 26 failed, exit 3
  400074  ExportViewException:            Error: data sources not connected
  400081  DataSourcesNotConnectedException: Error: data sources not connected
\\\

Worksheets and dashboards alike. See #402 for the classification half of that.

It is also **default-view-state only** (no \?vf_\ filter pinning), which is the existing #194 ceiling.

## The gap

[...]

Issue #194, opened 2026-08-17
capture_tableau_reference.py sends operators to a dead end: server_rest exits 3 asking them to implement an endpoint capture_tableau_oracle.py already ships


Guust-Franssens:
## Summary

Two scripts solve overlapping problems. The one **named** for reference images is a stub that
**exits 3 telling you to implement an endpoint that a sibling script already implements**, and neither
`docs/reference-capture.md` nor `scripts/README.md` cross-links them. Reported from the field
2026-08-17 by someone who followed the naming and the docs correctly and hit a dead end.

### 1. The doc's status cell is factually wrong

`docs/reference-capture.md:76`:

```
| **Server/Cloud REST** (`/views/{id}/image?resolution=high`, ...) | Canonical when ... | ❌ specified-only (no Server to test against) |
```

`scripts/capture_tableau_oracle.py:404`:

```python
png, elapsed, stats = session.export(f"/sites/{session.site_id}/views/{view_luid}/image?resolution=high")
```

Same endpoint, same query string. Not only implemented — **exercised against a live Tableau Cloud site
today**, with retry accounting. `--images` is a documented flag (`capture_tableau_oracle.py:426`).
`docs/reference-capture.md` mentions `capture_tableau_oracle.py` **zero times**.

### 2. `scripts/README.md` hides the capability behind the wrong blurb

Line 52's `capture_tableau_oracle.py` description is framed entirely around *"the numeric oracle"* —
no mention of `--images`, no mention of the image endpoint. Line 53, immediately below, describes
`capture_tableau_reference.py` as *"Acquires a **provenance-stamped** reference image of the source
dashboard"* — the phrase a reader searching for "get me a screenshot" will match — and links to the
doc in §1 that says Server REST is a stub.

So the reader is routed, correctly by every signal we publish, into the dead end.

### 3. 🔴 The part the report didn't reach: it is an active failure, not a stale doc

`scripts/capture_tableau_reference.py:212-218`:

```python
def capture_server_rest(_slug_dir: Path) -> list[dict] | None:
    """STUB: Tableau Server/Cloud REST image export. Not implemented - no Server to validate against."""
    raise NotImplementedError(
        "server_rest provider is specified but not implemented (no Tableau Server available to test "
        "against). See docs/reference-capture.md. Set credentials via env (TABLEAU_SERVER_URL / _SITE / "
[...]

PR #405, opened 2026-08-30
Tableau dashboard reference: vector SVG, a PROBED render ladder for on-prem reach, and a stated grade (refs #403)

Guust-Franssens:
## The question

**What is the best available way to obtain a legible, full-page reference image of a Tableau
DASHBOARD?** Answered by measurement against site `fabric-migration-lab`, REST **3.29**, over **60
dashboards** (52 capturable, 8 blocked). Every number below is measured, not quoted.

**Answer: `/views/{id}/image?format=svg` — new in REST 3.29, same endpoint and credential as the PNG
we already call, one extra request. It is vector, self-contained, and its `<text>` elements carry the
dashboard's literal labels.** `/pdf` is a genuine but second-best alternative. Nothing rescues a
workbook whose data sources are disconnected.

## Prerequisite nobody had

**REST `/views` does not carry `sheetType` at 3.29.** All 360 views returned the same seven keys
(`contentUrl, createdAt, id, name, tags, updatedAt, viewUrlName`) and no type field. The **Metadata
API** (`POST /api/metadata/graphql`, `workbooks { dashboards { name luid } }`) is the only route that
labels a view as a dashboard. You cannot prefer a dashboard-specific route without it. This also
found **60** dashboards on the site vs the 18 in the harvested `.twb` estate.

## Measured comparison

| route | resolution | vector | credential | live conn. | dashboards | survives disconnected sources |
|---|---|---|---|---|---|---|
| `.twb` embedded thumbnail | **192×192 always** | no | **none** | **none** | yes | **yes** |
| `/image` | declared size (650×800 … 1500×850) | no | PAT | yes | yes | no |
| `/image?resolution=high` | **exactly 2× declared, 52/52** | no | PAT | yes | yes | no |
| **`/image?format=svg`** (3.29+) | **unbounded** | **yes** | PAT | yes | yes | no |
| `/pdf?type=Unspecified` | unbounded | yes | PAT | yes | yes | no |
| `/crosstab/excel`, `/data` | n/a (numbers) | n/a | PAT | yes | yes | no |

### 1. Is there a size/DPI control beyond `resolution=high`? **No, for raster.**

`resolution` is a strict enum of one: `high` → 200; `standard` → **400**; `veryhigh` → **400**;
`HIGH` → **400** (case-**sensitive**, unlike `format`). `vizWidth=3000&vizHeight=2000` is accepted and
**silently ignored** — byte-identical responses (728,498 B with and without). `maxAge=1` likewise.

`resolution=high` = **exactly 2× the dashboard's declared size, 52/52, no exception**. The dashboard's
author sets the ceiling, not the caller. Worst case measured: `Section 11 - Actions` is 650×800 →
**1300×1600 maximum, forever**.

That is the mechanism behind #403's hypothesis. `Superstore | Order Details` is an 800×800 dashboard
carrying **410 text labels**; at its 1600×1600 ceiling it is structurally legible and
content-illegible simultaneously — a structurally-correct page with wrong content passes.

### 2. Real dashboard resolution vs the 192×192 thumbnail

Declared sizes 650×800 … 1500×850. `resolution=high` therefore spans **1300×1600 … 3000×1700**.
`HR | Summary` = 1400×800 → **2800×1600**. Against a 192×192 thumbnail that is a **~121× area**
increase, and it is legible: I read `7,984`, `8,950`, `966` and the scatter's `35/40/45/50` axis ticks
off it directly.

**So the premise needs a correction:** the illegible reference is the *thumbnail* path. `--images` on
the oracle already produces a legible dashboard reference where the site is reachable. The real
defects are (a) the fixed 2× ceiling on dense pages, and (b) nothing in the manifest states which
grade was obtained.

### 3. `?format=svg` — the finding

Documented for `customviews` at 3.29; **it works on `/views/{id}/image` too**. Server's own rejection
text names the valid values: `'PNG', 'SVG'`.

- **Real text.** `<text>` holds literal strings — `Active Employees`, `7,984`, `Hired`, `8,950`.
  Content is readable **without rendering**. 21…410 per dashboard, median 69.
- **Self-contained.** Raster children are `data:` URIs; **external refs = 0** on every file. Durable
  offline evidence.
- **Exact geometry.** `round(mm × 96/25.4)` = declared size **+1 px per axis, 52/52**. `round` not
  `int`: the true value lands at 1400.99, so truncation is off by one erratically (three different
  offsets across the same 52). My own test caught me getting this wrong first time.
- **Rasterises free.** Chromium/Playwright (already a repo dependency, already shelled out to by
  `capture_tableau_reference.py`) at `deviceScaleFactor: 3` → **4203×2403**, layout-identical to the
  REST PNG, **no page margin**. **No new Python dependency.**
- **Fails loudly below 3.29.** 3.21/3.24/3.28 → HTTP 400 *"SVG export requires API version 3.29 or
  later"*. Never a silent PNG. ⚠️ the script still **defaults to 3.21**.
- **Not free.** 39 KB…5.0 MB per dashboard (PNG 48…897 KB). A crosstab *worksheet* gave a **21 MB**
  SVG / 37,439 `<text>` vs a 4.5 MB PNG. Prefer it for dashboards; think before sweeping an estate.

### 4. `/pdf` — works, vector, second choice

Vector confirmed by content-stream inspection, not by squinting: embedded `/FontFile` programs, **792
hex `Tj` glyph runs** on `HR | Summary`, and **1.1 MB of inflated vector content stream in a 37 KB
file** (`Bar Charts`, zero image XObjects). Two demotions:

- **Default page is US Letter portrait** (`MediaBox [0 0 612 792]`) — a 1.75:1 dashboard squeezed onto
  0.77:1. **`?type=Unspecified` is the fix**: page = `0.75 × declared_px + 72 pt` in *both* axes,
  **52/52** — the dashboard at 1:1 CSS scale plus a 0.5-inch margin per side. `A3/A4/Tabloid/Ledger/
  Letter` + `orientation` all work and all distort. `vizWidth/vizHeight` ignored here too.
- **Needs a new dependency.** No PDF rasteriser present (`pypdfium2`, `pymupdf`, Ghostscript, poppler,
  ImageMagick — all absent). Its one real edge: it **embeds fonts**; SVG only names them
  (`Trebuchet MS`, `Arial`), so glyphs substitute on a machine without them (positions do not move).

### 5. Does `/pdf` survive "data sources not connected"? **No.**

`image`, `image?format=svg`, `pdf` **and** `data` all return the identical HTTP 400:
`ExportViewException: Error: data sources not connected`. Tested on `global_superstores_db`
Dashboards 1 & 5, plus `RESTAPISample` and `filtering` (which fail the same way — they are not, as the
issue implies, uniquely a `global_superstores_db` problem). The failure is in **VizQL, upstream of the
output format**, so no format choice rescues it. Final answer; names a human action. Only the offline
192×192 thumbnail survives.

### 6. Other routes

| route | verdict |
|---|---|
| `/workbooks/{id}/powerpoint` | exists (3.8+); workbook-level, one slide per view — **untested**, and a PPTX still needs unpacking to an image |
| `?:size=W,H` embed param | needs a **browser + session cookie**, not REST. Not tested; SVG makes it unnecessary |
| `/crosstab/excel` | works, tiny (2.5 KB) — numbers, not a picture |
| Local `.twb` render | still the only credential-free, connection-free route above 192×192; still unbuilt |

## Recommendation, and its cost

**Run `--images --svg` together for any dashboard capture.** PNG = immediately viewable; SVG =
resolution-independent *and* content-as-text. Cost: **one extra REST call and some bytes per view**;
same endpoint, same credential, no new dependency. Requires `TABLEAU_REST_API_VERSION=3.29` in `.env`.

**This does not upgrade the evidence grade.** Still default-state, still outside `reference/`, still no
`capabilities` manifest. It raises the ceiling *within* text-grade. #194's cap stands.

## Shipped — this PR

`chore/tableau-dashboard-reference-probe`, commit `39a1c21`, `refs #403`.

- `--svg` on `capture_tableau_oracle.py`; manifest now records `dimensions_px` (PNG) and
  `width_px`/`height_px`/`text_elements`/`external_refs`/`vector` (SVG) + `image_ok`/`svg_ok` — #403's
  "nothing states which grade was obtained".
- **Two bugs found by having two legs:** `capture_view` returned early after a failed image (would have
  swallowed the SVG entirely), and `write_manifest` counted only `image` in its complete/blocked/failed
  sets. `_render_statuses()` now feeds all three the same list.
- Pre-3.29 → `unsupported_api_version` + the `.env` remedy, reported **separately** from
  `credential_blocked`: one is fixed by an `.env` line, the other by a human in Tableau.
- Docs: route survey in `docs/reference-capture.md`; `scripts/README.md`; one paragraph in `AGENTS.md`.

**Mutation table — 8/8 caught, each by a test named for the behaviour:**

| mutation | verdict | named failing test |
|---|---|---|
| `round` → `int` in the mm conversion | CAUGHT | `test_svg_root_millimetres_are_converted_to_the_viewport_pixel_size` |
| drop the version-gate reclassification | CAUGHT | `test_a_pre_329_site_is_reported_as_a_configuration_fault_not_a_broken_view` |
| count only `image` in the render legs | CAUGHT | `test_a_failed_svg_leg_makes_the_view_incomplete_and_the_run_partial` |
| treat a `data:` URI as external | CAUGHT | `test_a_data_uri_raster_child_is_self_contained_but_a_server_href_is_not` |
| return early after a failed image leg | CAUGHT | `test_a_failed_png_no_longer_swallows_the_svg_leg` |
| stop recording PNG dimensions | CAUGHT | `test_png_and_svg_are_captured_side_by_side_from_the_same_endpoint` |
| `svg_ok` counts every view | CAUGHT | `test_a_data_only_capture_is_still_complete_when_no_render_was_requested` |
| mark the SVG leg as raster | CAUGHT | `test_svg_capture_writes_the_file_and_grades_it_in_the_record` |

**Controls (per "a perfect first result deserves suspicion"):** a behaviour-neutral log-wording change
**SURVIVED** (rc=0), and an induced `import` error also scored **SURVIVED** (rc=2, zero named
`FAILED`) — the harness scores named failures, not any non-zero exit.

Gates: `pylint scripts` **10.00/10 exit 0**; `ruff` clean; `sync_agent_conventions.py --check` and
`check_navigation_index.py` exit 0; **132 targeted tests pass**; live end-to-end `--images --svg` run
exit 0 with both legs written and graded.

## What I could NOT verify

1. **The #403 hypothesis itself.** I established the *mechanism* (a hard 2× ceiling; 410 labels in
   1600×1600). I did **not** establish that any customer sign-off was actually given against a missing
   or 192px dashboard reference. That still needs the specific case.
2. **`/pdf` and `/pptx` rendering fidelity.** I proved the PDF is vector by content-stream inspection
   and rendered it with a **temporarily installed** `pypdfium2` (isolated venv, deleted). I did not
   evaluate `/workbooks/{id}/powerpoint` at all.
3. **The +1 px law for worksheets.** Solid for dashboards (52/52). One worksheet (`BAN Hired`, PNG
   1584×1584) reported **792×792 — offset 0**. n=1; the code and docs scope the claim to dashboards.
4. **Font substitution risk at scale.** SVG embeds no fonts. On this Windows box `Trebuchet MS`/`Arial`
   were present so the three routes looked near-identical; on a machine lacking a customer's corporate
   font the SVG will substitute glyphs (positions are absolute, so no reflow). Untested.
5. **Whether SVG survives `?vf_` filter pinning** (#194's state ceiling). Not probed.
6. **Whether `/image?format=svg` exists below Tableau Cloud's June 2026 build**, i.e. on a customer's
   older on-prem Server. This site is Cloud at 3.29; an older Server may simply not offer it. The code
   handles the refusal cleanly, but the *frequency* of that refusal in the field is unknown.
7. **Rate limits.** Tableau Cloud meters exports at ~100/hr/Creator and 20 concurrent. I made ~300
   calls without hitting a 429, but `--svg` doubles the render calls on an estate sweep and I did not
   test the ceiling.
8. **One PAT = one live session** (discovered by accident): two concurrent probes invalidated each
   other's tokens (`401002`, then hard 401). Reproduced once, root-caused, not systematically tested.


Guust-Franssens:
## The question

Follow-up on the one thing I could not verify: **does `?format=svg` exist on the on-prem Tableau
Server versions customers actually run?**

**Answer: no, and it will not for years.** SVG needs REST **3.29 = Tableau Cloud June 2026 / Server
2026.2**. On-prem customers on **2023.x–2025.x (API 3.19–3.25)** have none of it. But `/pdf` reaches
back to **2018** and is genuinely vector — so this is a **two-tier answer**, and the tier is now
**probed, never assumed**.

## Version / capability matrix

**MEASURED** — live, Tableau Cloud `fabric-migration-lab`, one PAT, no concurrency:

| what | result |
|---|---|
| site now advertises | `productVersion 2026.3.0`, `build 20263.26.0824.1544`, **`restApiVersion 3.30`** |
| **the same site, one week earlier** | **2026.2.5 / 3.29** — Cloud force-upgraded mid-investigation |
| `/serverinfo` reachability | HTTP **200 unauthenticated** at client api 2.4, 3.15, 3.21, 3.29 alike; **404** at 3.99; always reports the *server's* number, never the one asked for |
| `?format=svg` at client api 3.15 / 3.21 / 3.25 | **HTTP 400** — *"SVG export requires API version 3.29 or later"* |
| `?format=svg` at 3.29 | **200**, `image/svg+xml` |
| `?resolution=high` at 3.15 / 3.21 / 3.25 / 3.29 | **200**, and identically 2× (1000×800 → 2000×1600) at every one |
| `/pdf?type=Unspecified` at 3.15 / 3.21 / 3.25 / 3.29 | **200**, `application/pdf` at every one |

⚠️ Those old-version rows are a **2026.3 server answering an old client api-version**. They prove the
API contract is stable across versions *on this server*; they are **not** a 2023.x Server.

**INFERRED from Tableau's REST reference** (we have no on-prem site — this half is necessarily
documentation-based):

| route | API floor | Tableau release | on-prem 2023.x–2025.x |
|---|---|---|---|
| `/image` + `?resolution=high` | **2.5** | **Server 10.2** (2017) | ✅ universal |
| `/pdf` | **2.8** | **Server 10.5** (2018) | ✅ universal |
| `/data` | 2.8 | Server 10.5 | ✅ |
| `/crosstab/excel` | 3.9 | Server 2020.3 | ✅ |
| **`/image?format=svg`** | **3.29** | **Cloud June 2026 / Server 2026.2** | ❌ **none** |

> *"Available in API 3.29 (Tableau Cloud June 2026 / Server 2026.2 and later:"* — verbatim, and
> **identical on `Query View Image` and `Get Custom View Image`** (including the unclosed
> parenthesis). So SVG **is** documented for on-prem, just only from 2026.2. Tableau's REST 3.29
> "What's New" does not mention SVG at all — an omission, not a contradiction.

**The API → release map** is committed as `API_RELEASE` in `scripts/tableau_render_capability.py` and
test-gated. Six rows are **Cloud-only** — 3.26, 3.24, 3.22, 3.20, 3.18, 3.16 — unreachable on-prem
ever. On-prem 2025.1 tops out at **3.25**; 2023.3 at **3.21**.

⚠️ **The published table stops at 3.29** while the live site already reports **3.30**. `release_for()`
returns *"not in the published table"* rather than inventing a release.

## Two corrections to what I shipped in the first half

1. **`vizWidth`/`vizHeight` are ignored for DASHBOARDS, not universally.** A dashboard has a fixed
   declared size. A **worksheet** *does* honour `vizHeight`: `Revenue by Region` went **361×835 →
   361×1535**, and → 722×3070 with `resolution=high`. The 2× dashboard ceiling stands; the blanket
   claim did not.
2. **`type=Unspecified` is measured, not documented.** Tableau's documented values are
   `A3, A4, A5, B5, Executive, Folio, Ledger, Legal, Letter, Note, Quarto, Tabloid` and the docs say
   the default is `Legal`. Measured: the default is **612×792 = Letter portrait** (explicit
   `?type=Legal` → 612×1008). `pdf_facts` now records the `MediaBox` actually returned.

## Recommendation

**Cloud → `svg`. On-prem below 2026.2 → `pdf`. Just pass `--reference-best` and let it probe.**

- **PDF is the portable answer** (2018-era floor), genuinely vector — 792 hex `Tj` runs, embedded
  `/FontFile` — and it **embeds its fonts**, which SVG does not, so it is *more* faithful on a machine
  lacking the workbook's typefaces. Cost: a rasteriser dependency (none installed here). That cost is
  the only reason it is not the default where SVG exists.
- **SVG is the premium answer**: no new dependency (Chromium/Playwright already ships) and content as
  machine-readable `<text>` — but 2026.2+ only.
- **PNG is the floor** and always works.

## Detect capability; do not assume it from a version string

Three numbers disagree: what we **ask** (`TABLEAU_REST_API_VERSION` — a *client preference*), what the
server **advertises** (`/serverinfo`), and what the endpoint **does**. Only the third is
authoritative. Measured live on one site:

| client pin | selected tier | warning |
|---|---|---|
| 3.29 | **svg** | — |
| **3.21** (on-prem-shaped) | **pdf** | *"tier 'svg' is supported by this server (advertises API 3.30) but TABLEAU_REST_API_VERSION is pinned to 3.21"* |
| 3.29, **blocked** probe view | **`null`** | *"capability UNDETERMINED … re-probe with a different view"* |

That third row is the one that matters. A workbook with disconnected sources fails **every** route
identically, so a naive probe would report "this site cannot render" for a perfectly capable site —
the unassessable-collapsing-into-clean shape. `classify_probe` keeps `available` / `unsupported` /
`indeterminate` strictly apart, and `--reference-best` tries up to 3 views before giving up.

The tier, per-rung verdicts and all three version numbers land in `oracle-manifest.json` under
`render_capability`. **The grade is stated, not inferred** — #403's surviving half.

## Design fixes taken, not suppressed

Pylint surfaced four real smells; none were waived. The new module's CLI imported back into
`capture_tableau_oracle` (**cyclic** — it now has its own 12-line sign-in and is genuinely
standalone); `RenderTier.min_release` was derivable, so it became a property over `API_RELEASE`, which
is *how the release map got encoded at all*; `capture_view`'s three parallel `want_*` booleans became
one `wants` frozenset; `write_manifest`'s provenance parameters became a `CaptureRun` dataclass. And
`_partition` now builds complete/blocked/failed in one place — the bug it replaces was three
comprehensions where only two had been taught about a new leg.

## Mutation table — 14/14 caught, control survived

| mutation | verdict | named failing test |
|---|---|---|
| string version compare instead of numeric | CAUGHT | `test_api_versions_compare_numerically_not_as_strings[3.9-3.10]` |
| unknown server version reads as 'cannot' | CAUGHT | `test_an_unknown_server_version_is_none_not_false` |
| version gate misfiled as indeterminate | CAUGHT | `test_a_genuine_refusal_of_every_tier_is_reported_as_incapable` |
| disconnected sources misfiled as unsupported | CAUGHT | `test_a_disconnected_source_says_nothing_about_capability` |
| all-indeterminate collapses to 'incapable' | CAUGHT | `test_an_all_indeterminate_walk_reports_UNDETERMINED_not_incapable` |
| ladder ordered worst-first | CAUGHT | `test_the_ladder_is_ordered_best_first` |
| probe every rung, never stopping | CAUGHT | `test_an_on_prem_site_below_329_falls_back_to_pdf` |
| pin warning fires when the server cannot | CAUGHT | `test_no_pin_warning_when_the_server_genuinely_cannot_do_it` |
| cloud-only marker dropped | CAUGHT | `test_cloud_only_releases_are_marked_because_on_prem_can_never_reach_them` |
| unknown API invents a release | CAUGHT | `test_an_api_version_past_the_published_table_says_so_rather_than_inventing_a_release` |
| PDF page geometry not recorded | CAUGHT | `test_pdf_facts_records_the_page_actually_returned_not_the_one_requested` |
| PDF embedded-font count zeroed | CAUGHT | `test_a_pdf_is_always_marked_vector_and_its_embedded_fonts_counted` |
| pdf leg dropped from the render legs | CAUGHT | `test_a_failed_pdf_leg_is_counted_like_the_others` |
| pdf captured with a paper size | CAUGHT | `test_pdf_capture_uses_the_fitted_page_route` |
| **CONTROL: neutral log wording** | **SURVIVED** | `-` |

The control confirms the harness scores **named `FAILED` lines**, not any non-zero exit — the property
that made the first round's induced-import-error control also score SURVIVED.

Live end-to-end: standalone report (exit 0, tier `svg`); 3.21 fallback to `pdf` with the pin warning;
blocked-view `UNDETERMINED` (exit 1); `--reference-best --pdf` writing both legs with correct geometry
(SVG viewport **1001×801** for a declared 1000×800; PDF page **822×672** = 0.75·px + 72).

Gates: `pylint scripts` **10.00/10 exit 0**, `ruff` clean, `sync_agent_conventions --check` and
`check_navigation_index` exit 0, **166 targeted tests** green. Rebased on `origin/master` (picks up
#398, #401).

1. **Any real on-prem Tableau Server.** The entire version-floor table is **documentation-based**. The
   old-api probes ran against a 2026.3 server, which proves contract stability, not old-server
   behaviour. An actual 2023.3 Server could differ and I have no way to know.
2. **Whether the 2× ceiling holds on old Server builds.** Measured identical at client api 3.15–3.29
   *on this server*. Whether a 2023.x Server renders the same is untested.
3. **`type=Unspecified` on any other server.** Undocumented, so it could vanish or behave differently.
   Mitigated (the returned `MediaBox` is recorded, never assumed) but not verified elsewhere.
4. **`/workbooks/{id}/powerpoint`** — still untested.
5. **API 3.30's contents.** The site advertises it; Tableau documents nothing. It may add or change
   render behaviour.
6. **Whether SVG shipped in on-prem 2026.2 in practice.** The method reference names Server 2026.2,
   but REST 3.29's "What's New" never mentions SVG. If it slipped, on-prem reach is *later* still.
7. **Font substitution on a machine lacking the workbook's fonts** — still untested; it is the stated
   reason PDF is preferred there, and that preference is reasoned, not measured.

Guust-Franssens:
## All six addressed. Five reproduced independently before fixing, then proved fixed by the same reproductions.

I reproduced findings 1–5 myself first — **all five confirmed**, no dispute:

| finding | my reproduction (before) | after |
|---|---|---|
| 5 security | token present in report JSON: **True**; `raw_get` body carried it: **True** | **False** / scrubbed to `[REDACTED]` |
| 4 exit 0 | `exit=0 captured_complete=1 image_ok=0 svg_ok=0` | **exit=5**, `reference_missing=true` |
| 3a | `selected='pdf' warnings=[]` | `provisional=True complete=False` + warning |
| 3b | `['no render tier is available on this site']` | `UNDETERMINED` |
| 2a/2b | PNG bytes → `selected='svg'`; `_capture_render` → `status='ok' vector=True` | `indeterminate` / `format_mismatch`, **no file written** |
| 1 | `"is supported by this server (advertises API 3.29)"` | claim only when a **floor re-probe succeeds** |

## Finding 6 — partly disputed on attribution, accepted on substance

**CONFIRMED, and it is a real defect.** `tests/mutation_harness.py:236`:
```python
verdict = "CAUGHT " if rc != 0 else "SURVIVED"
```
`run()` extracts named failures into `detail`; `main()` discards them. Any non-zero exit scores CAUGHT.

**DISPUTED: it did not produce my 14/14.** That run used a scratch scorer whose verdict was
`proc.returncode != 0 **and** names` — named `FAILED` lines required — and whose published control was
an **induced import error, exit 2, scored SURVIVED**. That is precisely the shape
`mutation_harness.py` miscalls. Both facts are in PR #405's own body, in the round-1 comment. I am not
asking you to take that on trust — see the re-score below, which reproduces the property explicitly.

**ACCEPTED, and it is the more important half.** That scratch scorer never verified a **clean
baseline** and never required the failure to be **new** or **relevant**. You are right that a neutral
control tests one direction only. The number was under-evidenced regardless of which harness produced
it, so it has been redone properly.

## Re-scored on a scorer built so its own controls disprove the defect

Aborts unless the baseline is green; **self-tests that a nonexistent test path does not score CAUGHT**;
requires a **NEW** named failure in a **RELEVANT** test file. Exit code is evidence, never the verdict.

```
baseline exit=0 named_failures=0
nonexistent path -> exit=4 named=0 verdict=SURVIVED

real mutations CAUGHT:              24/24
negative controls behaving right:    3/3
  CONTROL neutral log wording   -> SURVIVED  (exit 0, no failures)
  CONTROL induced import error  -> SURVIVED  (exit 2 but ZERO new named failures)
  CONTROL induced syntax error  -> SURVIVED  (exit 2 but ZERO new named failures)
```

The two **exit-2** controls are the decisive ones — `mutation_harness.py` would score both CAUGHT.
The honest number is **24/24**, higher than 14 only because the fixes added gated behaviour.

<details><summary>Full mutation table (24 real + 3 controls)</summary>

| mutation | verdict | named failing test(s) |
|---|---|---|
| string version compare instead of numeric | CAUGHT | `test_api_versions_compare_numerically_not_as_strings[3.9-3.10]` |
| unknown server version reads as 'cannot' | CAUGHT | `test_an_unknown_server_version_is_none_not_false` |
| cloud-only marker dropped | CAUGHT | `test_cloud_only_releases_are_marked_because_on_prem_can_never_reach_them` |
| unknown API invents a release | CAUGHT | `test_an_api_version_past_the_published_table_says_so_rather_than_inventing_a_release` |
| ladder ordered worst-first | CAUGHT | `test_a_clean_win_is_not_provisional` |
| **format signature check removed** | CAUGHT | `test_png_bytes_are_rejected_for_an_svg_request` |
| **content-type mismatch ignored** | CAUGHT | `test_a_mismatched_content_type_is_rejected_even_when_bytes_look_right` |
| **a 200 is trusted without checking the format** | CAUGHT | `test_a_200_carrying_the_wrong_format_is_indeterminate_not_available` |
| **capture persists a mismatched payload** | CAUGHT | `test_wrong_format_bytes_are_refused_rather_than_persisted_as_the_requested_type` |
| version gate misfiled as indeterminate | CAUGHT | `test_a_version_gate_is_a_permanent_no_for_that_tier` |
| disconnected sources misfiled as unsupported | CAUGHT | `test_a_disconnected_source_says_nothing_about_capability` |
| **provisional flag never set** | CAUGHT | `test_a_transient_failure_above_the_winner_makes_the_selection_provisional` |
| **capability_complete always true** | CAUGHT | `test_an_all_indeterminate_walk_reports_UNDETERMINED` |
| **UNDETERMINED collapses into 'no tier available'** | CAUGHT | `test_a_version_gate_plus_a_blocked_view_is_UNDETERMINED_not_incapable` |
| **floor re-probe removed (support inferred again)** | CAUGHT | `test_a_version_gated_tier_is_reprobed_at_its_documented_floor` |
| **pin warning claims support without a re-probe** | CAUGHT | `test_an_advertised_but_absent_feature_is_never_claimed_as_supported` |
| **probe detail reported unredacted** | CAUGHT | `test_the_probe_detail_is_redacted_before_it_can_be_serialised` |
| **requested-but-absent render leg reads as ok** | CAUGHT | `test_a_requested_render_that_never_arrived_is_a_failure` |
| **reference_missing never computed** | CAUGHT | `test_reference_best_that_selected_no_tier_exits_5_rather_than_0` |
| **exit code 5 removed** | CAUGHT | `test_reference_best_that_selected_no_tier_exits_5_rather_than_0` |
| PDF page geometry not recorded | CAUGHT | `test_pdf_facts_records_the_page_actually_returned_not_the_one_requested` |
| PDF embedded-font count zeroed | CAUGHT | `test_a_pdf_is_always_marked_vector_and_its_embedded_fonts_counted` |
| pdf leg dropped from the render legs | CAUGHT | `test_a_failed_pdf_leg_is_counted_like_the_others` |
| pdf captured with a paper size | CAUGHT | `test_pdf_capture_uses_the_fitted_page_route` |
| CONTROL neutral log wording | **SURVIVED** ✓ | exit 0 |
| CONTROL induced import error | **SURVIVED** ✓ | exit 2, zero new named failures |
| CONTROL induced syntax error | **SURVIVED** ✓ | exit 2, zero new named failures |

</details>

## Notes on individual fixes

**5.** `raw_get()` still returns the body **RAW by documented contract** — classification must read raw
text, because `redact` is handed the human-chosen PAT *name* and a short one rewrites Tableau's error
codes (a mangled `401002` stops being recognisable as recoverable session loss). That is the existing
`classify_export_error` rule; the fix applies it to the new fetchers rather than bypassing it.
Redaction happens in `classify_probe` before `detail` exists. Regressions on **both** fetchers, plus
one asserting classification reads raw while reporting redacted.

**1. The re-probe does more than reword the warning — it recovers the tier.** Measured live: a client
pinned to 3.21 against the 3.30 site now selects **`svg`**, where before it fell back to `pdf` on an
inferred claim. And in the unverifiable case (advertises 3.29, SVG absent) it claims nothing and says
**UNKNOWN**.

**4.** A plain data-only run with no `--reference-best` still exits 0 — separately gated, so the new
code cannot creep into ordinary use.

## Live re-verification (Cloud, 2026.3.0 / advertises 3.30)

| scenario | result |
|---|---|
| pin 3.29 | `svg`, `capability_complete: true`, exit 0 |
| pin 3.21 (on-prem-shaped) | **`svg` recovered** by floor re-probe; *"WORKS on this server — proved by re-probing at API 3.29"* |
| `--reference-best` on the all-blocked workbook | `UNDETERMINED`, **exit 5**, loud message |
| `--reference-best --pdf` happy path | `svg`+`pdf`, `reference_missing: false`, exit 0 |

## Design

One waiver, with its reason inline: `TableauSession._request` carries `too-many-arguments` because its
parameters map 1:1 to distinct HTTP concerns (verb, path, entity body, Accept, auth, API-version
segment) and grouping any two would be arbitrary. Everything else structural: `ApiVersions` groups the
two version *claims* so no caller can pass one and forget the other; `_walk_one_tier`,
`_add_pin_warnings`, `_log_versions`, `_build_report` keep `detect`/`main` readable.

Gates: `pylint scripts` **10.00/10 exit 0**, ruff clean, `check_agent_capabilities` /
`check_navigation_index` / `sync_agent_conventions --check` all exit 0, **372 targeted tests** pass.
`_runs/` untouched — read-only throughout, `git status _runs` clean. None of the protected paths were
modified.

Back to you — you have the reproductions, so the re-review is cheapest with you.

Issue #177, opened 2026-08-15
preview-visuals omits every property fidelity review turns on, so a validator using it as its discovery tool raised a false blocker


Guust-Franssens:
Five findings about the **verification chain** — the tools an agent uses to decide whether a report is
correct — from a blind end-to-end "cold run" of this toolkit (a fresh agent given only the committed
docs, migrating a 17-workbook estate; stumbles S23, S25, S26, S29).

They share one shape: **each tool answers a narrower question than the reader believes, and none of
them says so.** One produced a false BLOCK on a correct report; another certifies a hard defect as
clean; a third silently does nothing at all.

Destination: `.github/skills/powerbi-report-gotchas/SKILL.md` (§3 and §7) and
`scripts/capture_powerbi_pages.py`.

> ⚠️ **SEQUENCING.** #165 is in flight **in this same skill file**, on §1, §5 and §8. Land after it and
> rebase. The two issues do not overlap in content, but they will conflict in the file.

---

## Defect 1 (highest priority — it caused a FALSE blocker) — `preview-visuals` hides every property fidelity review turns on

**Measured, cold run S25.** A sign-off validator reported that `book_6-1-Maps` did **not** bind its
viz-in-tooltip page. The fix round found the binding **already correct** and made no edit. A
spot-check adjudicated on file bytes and confirmed:
`visualTooltip {show:true, type:'Canvas', section:'page-tooltip-over-time'}` present, matching the
target page's `name`, with `pageBinding.type="Tooltip"`, `HiddenInViewMode`, 320x240.

**Root cause is tooling, not judgement.** `powerbi-report-author preview-visuals` — the natural
discovery command, *whose blurb says "agents read each `path` for full JSON"* — emits per visual only:

```
{path, pageName, pageDisplayName, visualName, visualType, position, hasFilters, hasFormatting, hasVCOs}
```

Measured on that report: occurrences of `visualContainerObjects` = **0**, of `visualTooltip` = **0**,
**both with and without `--with-derived`**. So **tooltip bindings, custom titles, data colours and
reference lines are all invisible** in the summary, and `hasVCOs` is a bare boolean that says something
exists without saying what. A validator trusting the summary concludes "not bound" from absent data.

**Cost:** ~35 minutes, one wasted item in a fix round, plus the round trip to disprove it.

**What to deliver:** a §7 entry stating plainly that `preview-*` commands are an **inventory**, not a
[...]

Issue #164, opened 2026-08-15
Three green gates, zero rows: no emitted model could load data while EMPTY-MODEL CHECK, openability_selfcheck and probe_bundle all passed


Guust-Franssens:
The single highest-value learning from a blind end-to-end "cold run" of this toolkit: **three
independent validation gates all returned green while not one of 17 emitted semantic models could load
a single row.**

Primary destination for the fix: `.github/skills/powerbi-semantic-model-gotchas/SKILL.md`. That file is
**published to other repositories** via the `powerbi-playbook` / `powerbi-migration-skills` plugin and
is deliberately **source-tool agnostic**.

---

## Defect 1 — three green gates, one blind spot

**Measured, cold run S8.**

What the gates said:

- `run_estate.py` printed `EMPTY-MODEL CHECK … OK - every file-backed Import partition resolves to a
  file with data rows.` and `ESTATE: READY`
- every handover slice reported `openability_selfcheck.ok: true` on **all six** checks
- `probe_bundle.py --check-only` exited **0**

What was actually true: the `.pbip` opens as `Untitled - Power BI Desktop` with **no database**
(`DAX Evaluate queries work only on databases which have at least one table`). Two independent defects,
**both already documented in this very skill file**, neither caught by any gate:

1. **(§7.1)** Every `Orders` partition emits `Source{[Item="Orders", Kind="Sheet"]}[Data]`, but the
   landed files are genuine **BIFF8 `.xls`** — magic bytes `D0 CF 11 E0 A1 B1 1A E1`, verified
   byte-wise — whose navigation table has only `Name`/`Data` columns. Result:
   *"The key didn't match any rows in the table"*; refresh fails outright. Correct form:
   `Source{[Name="Orders"]}[Data]`.
2. **(§6)** `Table.TransformColumnTypes` is emitted with **no culture** on an **en-BE / LCID 4096**
   host (decimal comma, `dd/MM/yyyy`), which silently inflates decimals and transposes dates **while
   refreshing green**.

After both repairs, the model refreshes to **9,994 rows / SUM(Sales) = 2,297,200.8603 /
2015-01-03 to 2018-12-30**, matching the source file and the Tableau oracle exactly.

**The root cause of the blind spot:** `check_empty_model.py` answers *"does the file behind the
partition have rows?"*, which is a **different question** from *"can this partition read it?"*. The
operator runbook presents `EMPTY-MODEL CHECK` as the guard against "it looks finished and shows
[...]

PR #407, opened 2026-08-30
Fixes #312: openability_selfcheck is the engine's CLAIM, not a verified open

### LINKED ISSUE #312: openability_selfcheck.ok is true on 46 of 61 defective workbooks, and two personas cite it as authoritative
**Second-pass audit finding, measured 2026-08-23 across 96 engine JSONs in three bundles.**

## `openability_selfcheck` oversells its scope, and two personas treat it as authoritative

| measurement | value |
|---|---|
| `ok: true` on workbooks that carry defects | **46 of 61** |
| `issues[]` empty | **60 of 61** |
| `checks` silently omits an unrun check (`eager_calc_refs_resolve`) | **44×** |
| `not_evaluated` explains how many of those omissions | **1 check name** |

So a field named *openability selfcheck* reports `ok: true` on a workbook that simultaneously carries dangling bindings and dropped filters, and its `checks` map **omits checks that did not run** rather than marking them unrun. An agent reading `ok` as *"this model is fine"* is reading a false green.

**Two personas cite it as authoritative.** That is the live half of this finding — the risk is not theoretical, it is in our own instructions today.

## What to do

1. **Correct the persona wording** so `openability_selfcheck.ok` is described as what it is: a narrow structural self-check whose `checks` map is **not exhaustive**, and whose `ok` says nothing about bindings, filters, relationships or data.
2. **Never gate on it.** If a future gate consumes it, it must treat an **absent** check as *not evaluated*, never as passed — the same none / not-present / present-and-empty discipline applied three times this week.
3. Consider whether the **absent-check** shape is worth reporting upstream. The engine emitting a `checks` map that silently drops unrun entries, with a `not_evaluated` list that explains only one of 44 omissions, is arguably a defect in the engine's own reporting rather than in our reading of it.

## Provenance

Measured by us on `_bundle-208`, `_bundle-op` and `_sweep-coldrun/bundle`. ⚠️ Those were built by engines **2.126–2.208**; the installed engine is **2.260.0**, so the shape may have changed. Verify before acting on point 3, and check whether the two persona citations are still accurate at the current version.

Full report: `files/unread-signals.md`.

Guust-Franssens:
Closes #312.

`openability_selfcheck` ships in the **engine's** `report.json` — it is not a check this repo runs.
Our tier's failure was citing it as if it were verification. That is the worst shape a signal can
have: a missing gate is a known gap, but a green that is *trusted* converts a defect into a sign-off.

## Ground truth, re-derived — the issue's number is directionally right, its corpus is gone

The issue's corpus (`_bundle-208`, `_bundle-op`, `_sweep-coldrun`, engines 2.126–2.208) **no longer
exists on disk**, so 46/61 cannot be re-derived from its own evidence. Re-measured the same shape on
the current engine, `_runs/estate-2.339.0-20260829/report.json` (45 workbooks, 44 carrying the field):

| measurement | 2.339.0 | issue (2.126–2.208) |
|---|---|---|
| `ok: true` while the same `report.json` records defects for that workbook | **30 of 44 (68%)** | 46 of 61 (75%) |
| `issues[]` empty | 43 of 44 | 60 of 61 |
| `ok` distribution | `true` 43 · `false` 1 · absent 1 | `true` 60 · `false` 1 · absent 16 |

**Confirmed.** Worked example: `Meridian Calc Gauntlet` ships `ok: true`, `issues: []` beside
`viz_dangling_bindings.count = 1`, 3 `pbip_ref_drops`, 7 `pbip_warnings` and a
`pbip_entity_binding_mismatch` — every one reported by the engine itself, in the same file, one key over.

**What it asserts**, from the engine source (`openability_gate.check_model_openability`): a **static
scan of model TMDL text**. It imports `re` and `os` and nothing else, touches no filesystem, loads no
AMO, and is never handed the report. `ok` is `not issues`. So it says nothing about visual bindings,
dropped filters, data, or whether the model opens.

## Two corrections to the issue, both measured

1. **`compatibility_level_current` is no longer silently omitted** — present 44/44 (the issue recorded
   23 unexplained omissions). The engine closed that between corpora.
2. **Two of the four "unexplained omissions" are vacuous, not silent skips.** Predicate test across 44
   models × 2 checks: **88 agreements, 0 disagreements.** `eager_calc_refs_resolve` is present iff the
   model contains a stub partition; `relationship_columns_exist` iff `relationships.tmdl` exists. In
   both cases an absent key means the hazard *cannot exist*. The one genuinely unevaluated conditional
   check is `typed_columns_in_header` (absent 34/44, headers never supplied) — and its hazard remains.
   `absent = not evaluated` is still the right discipline; "absent = alarming" is not.

## Consumers and dispositions

| consumer | disposition |
|---|---|
| `.github/agents/pbi-semantic-builder.agent.md` (×2) | **already correct** since #380 — no change, now pinned by test |
| `.github/agents/pbi-migration-validator.agent.md` | **under-specified, not false** — listed as an input with no adjudication rule, in the persona that signs off. Now a claim in the input list, with a rule in pass 1 |
| `.github/skills/powerbi-semantic-model-gotchas/SKILL.md` (×2) | already honest; the **uncapped bundle now owns the full breakdown**. Republished via `sync_installed_skills.py` |
| `docs/deterministic-tier-integration.md` | `openable_outputs` now labelled *the engine's claim, not a verified open*; upstream register updated |
| `scripts/probe_bundle.py` (×4), `scripts/README.md`, `docs/tableau-dax-translation-guide.md` | **already honest** — unchanged |

So *"two personas cite it as authoritative"* is **stale as of master**: one was fixed by #380, and the
other under-specifies rather than overclaims. The risk was real; the framing had moved.

## Gate decision: no new gate

The issue's own point 2 is *"never gate on it"* — a 20th `check_*.py` consuming this payload is the
thing it warns about. Openability already has a real gate: `check_datamodel.py` hands each model to
`TmdlSerializer.DeserializeDatabaseFromFolder` (the parser Desktop uses) and exits **3 UNASSESSABLE**
rather than 0 when it cannot run — the discipline `openability_selfcheck` fails to apply to its own
skipped checks. What was missing was a **regression guard on the citations**, so this adds a test, not
a script: no CLI surface, no `docs/INDEX.md` row, no `scripts/README.md` row.

`tests/test_openability_claim_citations.py` requires every persona block naming the field to carry a
qualifier, and pins a **citation floor** so the scan can never pass by matching nothing. Skills are
pinned by heading only — the bundle cites the field twice as an *indictment*, and a
"is this critical enough" prose matcher would either false-flag those or be loosened until it matched
anything.

## Mutation table

Byte-level substitutions on tracked files, each **asserted to have applied** (exactly one match
required; bytes must differ) before pytest runs; restore is an in-memory snapshot, verified after.

| mutation | verdict | named test that failed |
|---|---|---|
| `validator-input-block-loses-its-qualifier` (restores the pre-fix wording) | CAUGHT | `test_every_persona_citation_is_qualified[...validator]` |
| `builder-glossary-row-loses-its-qualifier` | CAUGHT | `test_every_persona_citation_is_qualified[...builder]` |
| `builder-second-citation-loses-its-qualifier` | CAUGHT | `test_every_persona_citation_is_qualified[...builder]` |
| `validator-citation-deleted-entirely` | CAUGHT | `test_the_scan_finds_the_known_citations[...validator]` |
| `gotchas-section-heading-removed` | CAUGHT | `test_the_gotchas_skill_owns_the_full_breakdown` |
| `table-row-block-widens-to-whole-file` | CAUGHT | `test_a_table_row_cannot_borrow_a_neighbours_qualifier` |
| `paragraph-block-stops-walking-backwards` | CAUGHT | `test_a_reflowed_paragraph_is_read_whole` |
| `qualifier-check-always-satisfied` | CAUGHT | `test_an_unqualified_citation_is_detected` |

**8/8 — and the first pass was 6/7 with one survivor.** `builder-glossary-row-loses-its-qualifier`
originally deleted only the word "narrow" from a row carrying four qualifiers, so the mutation was too
weak, not the test blind. Strengthened to replace the whole row description; it then failed as
expected. Recording this because a perfect first result is the thing to distrust.

## Exit codes

| gate | exit |
|---|---|
| `sync_agent_conventions.py --check` | **0** |
| `sync_installed_skills.py` | **0** (1 file republished) |
| `ruff format --check` / `ruff check` (`conftest.py scripts tests .github/skills`) | **0 / 0** |
| `pylint scripts` · `pbip-model-refresh/scripts` · `powerbi-ai-readiness/scripts` · `conftest.py` | **0 / 0 / 0 / 0**, all 10.00/10 |
| `pytest -q` (serial, as CI runs it) | 3 failed / **3072 passed** / 9 skipped — the three failures are the known `test_upstream_repro_pins.py` pins |

## Upstream

[`Yarbrdab000/tableau-fabric-skills#183`](https://github.com/Yarbrdab000/tableau-fabric-skills/issues/183)
— the post-wrap re-check added for upstream #177 rebuilds the payload with only
`ok`/`checks`/`issues`/`rechecked_after_row_predicate_wrap`, **dropping `not_evaluated`** (upstream
#141) **and `reference_case_mismatches`** (upstream #164). So the tri-state discipline #141 exists to
provide fails on exactly the payload whose check set was recomputed. Measured: 1 of 44 workbooks is
missing both keys, and it is precisely the one carrying `rechecked_after_row_predicate_wrap: true`.
A fix that silently retracts an earlier fix's disclosure.

## What I could NOT verify

* **The literal 46/61.** Its three bundles are gone from this machine; nothing under
  `vscode-projects/` still holds them. The re-derivation is the same measurement on a different
  corpus, so 68% vs 75% is not a trend.
* **"Defect-carrying" is my predicate, not the issue's.** I used the union of `viz_dangling_bindings`,
  `measure_filters_needs_review`, `pbip_ref_drops`, `pbip_warnings`, `partitions_needs_review`,
  `pbip_entity_binding_mismatches`, `visuals_projecting_stub_measures`, `remediation_worklist` and
  handoff `needs_review`/`stub`. The issue does not state its own, so the two counts are comparable in
  shape, not in definition.
* **Vacuity is measured on this corpus only.** 88/88 says the predicate held on 44 models at 2.339.0 —
  "unobserved in this corpus", not "provably vacuous".
* **A cross-check against Desktop.** Nothing here was opened in Power BI Desktop; the openability claim
  is compared against source and against the TMDL-oracle path by argument, not by an executed open.

## One flake worth naming

Under `pytest -n auto --dist loadfile`, `tests/test_tmdl_oracle.py` produced 16 ERRORs
(`UnicodeDecodeError` in a `subprocess` reader thread) on one run. I chased it rather than assuming:

| run | mode | tmdl_oracle |
|---|---|---|
| stashed baseline | parallel | clean (3049 passed) |
| this branch, run 1 | parallel | **16 ERRORs** (3042 passed) |
| this branch, run 2 | parallel | clean (3058 passed) |
| this branch, alone | serial | 30/30 passed |
| this branch, whole suite | **serial, as CI runs it** | clean (3072 passed) |

So it is a **non-deterministic race** in the dotnet-backed oracle test under xdist, **not** correlated
with this change, and not on CI's path (CI runs `uv run pytest -q`, serially). Flagged rather than
filed — the sample is small and it is not mine to fix here.


Guust-Franssens:
## Both findings confirmed by my own reproduction, then fixed. No pushback — the SKILL.md counterexample is decisive.

### 1 — HIGH, upheld. The fix for a false green introduced a false green.

Verified the counterexample before acting. `.github/skills/powerbi-semantic-model-gotchas/SKILL.md:318-323`, committed:

> `TmdlSerializer.DeserializeDatabaseFromFolder` deserializes this cleanly, but Power BI Desktop **refuses to open the `.pbip`** ("Could not add Measure with the name X because a Measure with the same name already exists"). … **This shipped and broke a Desktop open — do not repeat it.**

And `tools/tmdl_oracle/Program.cs:69` calls `DeserializeDatabaseFromFolder` and nothing else — no open, no commit; its own header says its scope is "deliberately ONE question". So "only the TMDL oracle proves a model opens" was exactly the same move as the defect #312 exists to fix, one layer up.

**Swept all three sites, not just the cited one:**

| site | now reads |
|---|---|
| `pbi-migration-validator.agent.md:216` | *"No static gate settles openability: the TMDL oracle (`check_unit.py --scope data-model`) is the mandatory parser-level gate and is itself necessary, not sufficient — only a cold Desktop open does."* |
| `powerbi-semantic-model-gotchas/SKILL.md` | rewritten around the §4 counterexample, naming what the oracle helper actually calls, and that a cold Desktop open still proves nothing about rows |
| `docs/deterministic-tier-integration.md` | same correction, with the duplicate-measure-name example inline |

Skill republished — `sync_installed_skills.py` now reports `IN_SYNC`.

### 2 — HIGH, upheld. The test did not enforce its invariant.

**Reproduced your bypass exactly** (one match, byte change asserted): your sentence in place of the validator's paragraph gave `citing blocks: [141, 212]`, **`unqualified findings: []`** — green on both assertions. A positive word list cannot express this invariant, because "claim" appears in the sentence that denies it.

Rewritten with two rules where there was one:

- **`DISCLAIMERS`** — every citing block must carry an explicit **negative** assertion (`not exhaustive`, `never verification`, `says nothing about`, `necessary, not sufficient`, `never passed`, `blind to`…).
- **`OVERCLAIMS`** — authority wording (`authoritative`, `verified proof`, `proves the model opens`, `trust it`, `definitive`, `guarantee`) fails a block **regardless of how many disclaimers accompany it**. That is the specific hole: a hedge does not cancel an assertion.

Plus both structural fixes you named: **every `.github/agents/*.agent.md` is discovered from disk** at collection time, and blocks are isolated to **one table row or one list item**, so a qualified neighbour cannot vouch for a bare citation. The failure message additionally detects a citation inside `<!-- BEGIN:shared-conventions -->` and says to fix `AGENTS.md`, never the persona copy.

Nine adversarial fixtures, including yours verbatim, `not exhaustive … but ok is authoritative`, and `A narrow claim: ... proves the model opens`.

**Applying the stricter rule immediately caught a second real site** the old test had passed: the validator's input-list block qualified only on the word "claims". Now: *"its account of itself, never verification."*

### Mutation — 13/13, and the route there is the report

| # | mutation | verdict | named test |
|---|---|---|---|
| 1 | **`reviewer-bypass-sentence`** (your text, verbatim) | CAUGHT | `test_every_persona_citation_is_qualified[pbi-migration-validator.agent.md]` |
| 2 | **`bare-citation-in-a-previously-unscanned-persona`** | CAUGHT | `…[tableau-migrator.agent.md]` |
| 3 | `validator-input-block-loses-its-disclaimer` | CAUGHT | `…[pbi-migration-validator.agent.md]` |
| 4 | `validator-rule-bullet-loses-its-disclaimers` | CAUGHT | `…[pbi-migration-validator.agent.md]` |
| 5 | `builder-glossary-row-loses-its-disclaimers` | CAUGHT | `…[pbi-semantic-builder.agent.md]` |
| 6 | `builder-second-citation-loses-its-disclaimer` | CAUGHT | `…[pbi-semantic-builder.agent.md]` |
| 7 | `validator-citation-deleted-entirely` | CAUGHT | `test_the_scan_finds_the_known_citations[…-2]` |
| 8 | `gotchas-section-heading-removed` | CAUGHT | `test_the_gotchas_skill_owns_the_full_breakdown` |
| 9 | `table-row-block-widens-to-whole-file` | CAUGHT | `test_a_table_row_cannot_borrow_a_neighbours_qualifier` |
| 10 | `bullet-isolation-disabled` | CAUGHT | `test_a_bullet_cannot_borrow_the_previous_bullets_qualifier` |
| 11 | `paragraph-block-stops-walking-backwards` | CAUGHT | `test_a_plain_paragraph_walks_back_to_its_first_line` |
| 12 | `overclaim-denylist-emptied` | CAUGHT | `test_the_reviewers_bypass_sentence_is_rejected` |
| 13 | `disclaimer-requirement-dropped` | CAUGHT | `test_a_bare_citation_is_detected` |

**Row 1 is the requirement you set: your sentence now fails the live persona assertion.**

Two things went wrong on the way, both worth recording:

- **First pass stopped at 3/4.** A mutation literal matched **zero** times — CRLF, the exact trap. The count guard raised `SystemExit` rather than scoring it CAUGHT. The harness now resolves both newline conventions and supports anchor-span replacement.
- **Second pass was 12/13 with `paragraph-block-stops-walking-backwards` SURVIVING.** The backward walk *is* live — the validator's numbered input list is that shape — but every fixture started at a list marker, so nothing exercised it. Added `test_a_plain_paragraph_walks_back_to_its_first_line`. Then 13/13.

### Persona cap

You were right that the correction would not fit in 453 chars. Bought it back rather than expanding:

- the verbatim `LAST` table-calc handover quote → moved to the **`powerbi-report-gotchas`** skill, which already holds it word-for-word (`SKILL.md:680-682`); the persona keeps the mechanism and the "6 other table calcs" number.
- the Playwright bullet → a pointer at `docs/reference-capture.md`, which now carries the full recipe.

`sync_agent_conventions.py --check` exits **0** at **29,592 chars (408 left)**. Nothing inside `<!-- BEGIN:shared-conventions -->` was touched.

### Exit codes

| gate | exit |
|---|---|
| `sync_agent_conventions.py --check` | **0** |
| `sync_installed_skills.py` | **0** (`IN_SYNC`) |
| `ruff format --check` / `ruff check` | **0 / 0** |
| `pylint` — `scripts`, `pbip-model-refresh/scripts`, `powerbi-ai-readiness/scripts`, `conftest.py` | **0 / 0 / 0 / 0**, all 10.00/10 |
| `pytest -q` (serial, as CI runs it) | 3 failed / **3080 passed** / 9 skipped — the three engine pins only |
| CI on the pushed head | **both jobs pass** |

### Not verified

- **Nothing was opened in Power BI Desktop.** The corrected wording asserts a cold Desktop open is what settles openability; that claim rests on the committed §4 report and the oracle's source, not on an open I performed today.
- **The `OVERCLAIMS` list is a denylist**, so it is a false-negative surface by construction — a novel phrasing of authority that avoids all seven patterns would pass the denylist, though it would still need a `DISCLAIMERS` hit to pass at all. Both rules must be satisfied, which is what makes that acceptable rather than merely hopeful.
- **The validator does not declare `powerbi-report-gotchas` in its "Skills you use"** section; the pointer is by name inline, which the `skill` tool resolves. Flagging it because I did not test that invocation from inside the persona.

Issue #218, opened 2026-08-18
Add cross-artifact gate for running-total visual-axis grain


Guust-Franssens:
`check_datamodel.py` currently discovers and checks only `.SemanticModel` folders (see its CLI usage and target discovery), so it cannot compare a `WINDOW(... ORDERBY(<col>))` measure to the PBIR visual category field that binds it.

Follow-up from #175 Defect 1: add a cross-artifact gate that, for each emitted report visual, compares running-total measure addressing (especially `ORDERBY(<col>)`) against the visual's category projection and flags mismatches such as a measure ordered by `'Orders'[Order_Date]` placed on a `'Date'[Month Start]` axis.

Acceptance sketch:
- Parse PBIR `visual.json` category bindings for the handover/report slice.
- Identify bound measures whose DAX contains `WINDOW`/`ORDERBY` (and, if feasible, `ALL(<date columns>)` running-total forms).
- Emit a warning/error when the addressed column/grain is not the visual category grain.
- Keep false positives low; if the report binding cannot be resolved, report `needs live EVALUATE` rather than blocking.

This is intentionally not part of #175 because that change is docs-only and the existing model-only gate lacks report context.

Guust-Franssens:
OBSOLETE — verified on 2026-09-02 at master `57fb161d2f08`. `git show --stat 58f83aa` reports **“DESCOPE #218: delete the running-total axis gate”**; merged PR #408 deliberately deleted `check_running_total_axis.py` and its supporting modules/tests, unwired the gate, and changed the skill to state that nothing verifies this automatically after the pre-agreed descope criterion fired. The proposed surface was explicitly abandoned rather than left half-built.

PR #408, opened 2026-08-30
DESCOPE #218: delete the running-total axis gate (criterion fired at round 7)

Guust-Franssens:
## ⚠️ This PR is now a DESCOPE. It adds no gate; it removes one.

The criterion agreed **before** round 6 began:

> If the next review round finds **any definite verdict — exit 0 or exit 1 — on input the tokeniser
> did not fully consume**, this PR is descoped. Not another round.

Round 7 found **4 findings**: one criterion trigger, plus three further ones — and two of those three
are worse than the criterion, because they undermine the *evidence* the earlier rounds rested on.
Trajectory across all rounds: **5 → 6 → 2 → 0 → 3 → 4 → 4.** Full evidence and every reproduction
from rounds 1–7: **#425.**

### F1 — the criterion trigger

Measured at `2c7bf38` with the **same fixture for every row** (axis `'Date'[Month Start]`, measure
bound in the `Y` role), through `check_running_total_axis.main()`, judged by **exit code**.
Reachability is proven by the positive control, which uses the identical fixture and exits 1:

| input | exit |
|---|---|
| ✅ control — `SUMX(WINDOW(1,ABS,0,REL,ORDERBY('Orders'[Order_Date],ASC)), CALCULATE(SUM('Orders'[Sales])))` | **1** MISMATCH |
| ❌ `RUNNINGSUM(SUM('Orders'[Sales]), ORDERBY('Orders'[Order_Date], ASC))` | **0** NOT_APPLICABLE |
| ❌ `MOVINGAVERAGE(SUM('Orders'[Sales]), 3, ORDERBY('Orders'[Order_Date], ASC))` | **0** NOT_APPLICABLE |
| ❌ `CALCULATE(SUM('Orders'[Sales]), 'Orders'[Order_Date] <= MAX('Orders'[Order_Date]))` | **0** NOT_APPLICABLE |
| ❌ `CALCULATE(SUM(...), CALCULATETABLE(VALUES(d), d <= MAX(d)))` | **0** NOT_APPLICABLE |

Exit 0 prints *"no running-total, window or period-to-date measure in this model"* — for a measure
named `RUNNINGSUM`. That is round 6 finding 1's exact shape at round 7. **F1 alone met the criterion.**

### ⚠️ F2–F4 — three further findings

**F2. The mutation proof was VOID.** `test_every_mutation_in_the_harness_names_a_symbol_that_still_exists`
**failed** at `2c7bf38`: 14 probes name symbols the 853-line pruning deleted. So rounds 5–6's
"mutation-proven" claim never covered the shipped code. **A new vacuity mode** — the harness was
structurally sound and scoring correctly, aimed at targets that no longer existed.

**F3. ⚠️ The exclusion rationale was factually wrong.** `dax_tokens.py:83` excluded
`RUNNINGSUM`/`MOVINGAVERAGE` because "their relation is REQUIRED".
[Learn](https://learn.microsoft.com/en-us/dax/runningsum-function-dax): `RUNNINGSUM(<column>[,
<axis>][, <orderBy>][, <blanks>][, <reset>])`, computed *"along the given axis of the **visual
matrix**"*, `axis` **optional**, *"if omitted, the first axis of the Visual Shape definition is
used."* No relation parameter — the visual **always** decides. The two strongest cases for the gate
were excluded on a false premise.

**F4. Three repo gates were red at HEAD and unnoticed** — `dax_tokens.py` was committed with no
`scripts/README.md` row and no `docs/agent-capability-wiring.md` entry, plus a cascade in
`test_scripts_help.py`.

F2 and F3 are why another round would not have helped: the strongest evidence this PR had was
measuring nothing, and the design decision that produced F1 rested on a false reading of the docs.

### Why not patch it

The template matcher is sound: `dax_tokens.py` genuinely closed all three round-6 lexical defects
(`'Orders--Archive'` → exit 1, `SUM('WINDOW()'[Sales])` → exit 0, `SUM('Orders'[WINDOW(...)])` → exit
0), and held on every round-7 attack (escaped `''`/`]]`, non-ASCII, dead `VAR`, `ORDERBY(YEAR(col))`,
explicit relation, `TREATAS`/`IN {…}`, unterminated identifiers).

The defect is in the **screen**. `dax_grain._signals` decides *"is this an accumulation at all?"* from
an **open enumeration**, and every accumulation outside it is a silent exit 0. Patching the four found
here buys round 8 — the non-termination round 5 named, moved rather than removed.

A gate wrong in **both** directions is worse than no gate: false positives get it switched off, false
negatives grant false confidence.

### What changed

Deletes `scripts/check_running_total_axis.py`, `scripts/dax_grain.py`, `scripts/dax_tokens.py` and
their tests/harness; unwires the `running-total-axis` gate from `scripts/check_unit.py`; removes the
`scripts/README.md`, `docs/agent-capability-wiring.md`, `docs/INDEX.md` and golden-stdout entries.

**⚠️ The claim goes with the gate** (the #399 precedent). `SKILL.md` §4 now states plainly that
**nothing verifies this** and the `EVALUATE` probe is the only check — and carries forward everything
that survives deletion: the two measured mechanisms and the S14 axis-by-axis table, the
coarse-grain-is-not-`dateTime` finding, why an as-of bound is undecidable from the text, the
`RUNNINGSUM` correction with its Learn citation, and the estate sweep showing zero live instances.

**Net diff vs `origin/master` is one file: `SKILL.md`, +41/−3.**

### ⚠️ `scripts/check_unit.py` — the handoff guarantee, verified BY HASH

An earlier revision of this description claimed byte-equality with `master` on the strength of an
empty **three-dot** diff. That was wrong, and review caught it: three-dot answers *"what did this
branch change"* against the **merge base**, so it cannot establish equality with the tip. The branch
was **four commits behind** and HEAD lacked `b39aaae`'s SVG/PDF oracle-evidence handling — a
consolidation agent taking that file would have started on stale code and silently reverted it.

Current `origin/master` is now merged, and the guarantee is verified by **blob hash**, not by an
empty diff:

```
HEAD:scripts/check_unit.py            fa70dad9b7ac292ab08a6473a7640aade4840af9
origin/master:scripts/check_unit.py   fa70dad9b7ac292ab08a6473a7640aade4840af9
git rev-list --count HEAD..origin/master  ->  0
```

`git diff origin/master..HEAD -- scripts/check_unit.py` (**two dots**) is empty as independent
confirmation. Zero collision surface on the contended file.

### Verification (by exit code, on the merged tree)

| gate | result |
|---|---|
| `pytest tests/` | **3 failed, 3560 passed, 6 skipped** — all 3 are the documented `test_upstream_repro_pins.py` engine-drift tripwires (2.260.0 pinned vs 2.339.0 installed) |
| the 4 PR-owned failures at `2c7bf38` | **all gone** |
| `ruff format --check` / `ruff check` / `pylint` on `check_unit.py` | exit **0** / **0** / **0** |
| `check_navigation_index.py` | exit **0** |
| `sync_agent_conventions.py --check` | exit **0** |
| tracked-tree grep for the gate | only the deliberate "it does not exist" line in `SKILL.md` |

**Not verified:** `build_plugin.py --check` exits 1 with *"nothing built at `dist/marketplace`"* —
pre-existing in this worktree (`dist/` is untracked); I did not prove that against `master`. The
installed skill plugin copy was **not** refreshed (`sync_installed_skills.py` is owned by a sibling
agent), so a running session may still serve the pre-descope `SKILL.md`.

Review question for a descope is *"is the removal complete, and is the report now honest about what
it does not check"* — not *"is the gate good"*.


Guust-Franssens:
All five findings **reproduced independently first**, on copies of real estate/committed artifacts, before any code changed — then fixed, then re-verified. Ground truth against the pre-fix code:

| # | reproduction | pre-fix | post-fix |
|---|---|---|---|
| 1 | as-of + `dateTime` bin under the estate Section 12 pivot's `Rows` | exit 1 | exit 1 |
| 1 | …the **same** measure and column under the real `Columns` role | **silently `ok`** | **exit 1** |
| 1 | …the same measure on a card | **exit 0** | **exit 3** |
| 2 | as-of + the engine's own `Month = FORMAT(...,"MMM")` text bin | **silently `ok`** | **exit 1** |
| 2 | …the identical column forced to `dataType: dateTime` | exit 1 | exit 1 |
| 3 | `... <= DATE(2024,12,31)` on a coarse axis | **exit 1 (false positive)** | **not classified** |
| 4 | two `WINDOW`s, the second ordered by an unprojected column | **exit 0** | **exit 1** |
| 5 | committed Superstore `TOTALYTD(..., 'Sample Superstore'[Order Date])` + coarse same-table axis | **`NOT_APPLICABLE`, exit 0** | **exit 3** |

You were right about the through-line, and I fixed the shape rather than five instances.

## The proxies are gone, not patched

| proxy that was used | what is read now |
|---|---|
| a curated axis-role list; empty ⇒ "axis cleared" | **every projected column** groups the query, so all are examined. One shared `_grouping_or_reason` now answers *"was there a recognised grouping?"* for all three shapes; empty ⇒ `unassessable`. `AXIS_ROLES` survives as **presentation only** and its comment says so |
| declared `dataType` is not date ⇒ safe grain | declared type **or transitive calculated lineage** back to the anchor (`ModelFacts.grain_of`) |
| the first window call | every window call; ordered/partition columns unioned, one unreadable call ⇒ unassessable |
| period-to-date ⇒ "by design" ⇒ not assessed | judged against date-table marking; unproven ⇒ `unassessable` |

**On finding 2 your premise is stronger than stated.** It is not that date bins are *sometimes* text — in the engine's output they *usually* are. `Date (Big Data Source).tmdl` in the estate: `Year = YEAR(...)`, `Quarter No = QUARTER(...)`, `Quarter = "Q" & QUARTER(...)`, `Month No = MONTH(...)`, `Month = FORMAT(..., "MMM")` — **not one carries a `dataType`**. Only `*_Start` bins are date-typed. 95 `FORMAT(` columns estate-wide. A type-only check waves through the engine's own default shape, which I've now written into the skill as its own bullet.

## Finding 3 — I agree it needed narrowing, and split it three ways rather than two

You proposed "literals and unresolved parameter bounds are non-running **or** unassessable" and left the split to me. My reading:

- **constant bound** (`DATE(2024,12,31)`, `TODAY()`) → **not classified at all**, and *silently*. It is not an accumulation, so listing it would be the same noise as listing every other non-cumulative measure in the model. This is the only class dropped in silence and the docstring says so explicitly.
- **unresolvable bound** (`<= [As Of Date]`, a what-if parameter, a foreign column) → **`unassessable`**. It may well be an as-of date; nothing static decides it.
- **`VAR`-hoisted bound** → resolved and followed, because `VAR _asOf = MAX(t[Date])` is the *documented fix* for this shape; excluding it would have let the pinned-cutoff exclusion excuse the real thing. Covered by `test_an_as_of_bound_hoisted_into_a_var_is_still_a_running_total`, and by the `f3-var-bound-never-followed` mutation.

## A defect I found while verifying, which your list did not contain

Grading survivors in a dict keyed by `check_field_bindings.FieldRef` raised `TypeError: unhashable type` at runtime. **A crash also exits 1** — indistinguishable from a mismatch — and my own reproduction harness scored two cases as "fixed" on the strength of it. Caught only because the crashing run printed *nothing on stdout*. The harness now rejects empty stdout as "no verdict", and `test_a_grouping_column_is_never_used_as_a_dict_key` plus the `crash-grading-keyed-by-an-unhashable-fieldref` mutation cover it. Judging by exit code is necessary but **not sufficient** when the process can die.

## Module split, not a waiver

The fixes pushed `check_running_total_axis.py` to **1300/1200** lines (`C0302`), with `R0911` on `_judge_as_of`. Of the three sanctioned moves, shaving would have cost measured evidence (32% of the file is docstring), so I **split** at a real seam: new **`scripts/dax_grain.py`** answers *"what does this DAX address, and what grain IS this column?"* from TMDL alone and cannot see a report. **779 + 575 lines**, both well under. `R0911` fixed by extracting `_judge_same_table_survivors`.

## Mutation table — 33 mutations, 33 caught by the **named** test, 0 survivors

`python tests/mutation_harness_running_total_axis.py` → **exit 0**.

**On your harness warning:** I rolled my own scorer, and it never had the `tests/mutation_harness.py` defect — it has always required the **named** test from the table to appear in pytest's `FAILED` lines, and enforces a **green baseline** before running anything (`raise SystemExit` otherwise). I have now made the remaining gap explicit rather than implicit: a non-zero exit with **no named failure** is reported as `ERROR (no test outcome)` and fails the run, alongside misattribution. Plus two **controls that must SURVIVE** — if a no-op patch is ever scored CAUGHT the run aborts as unsound. Both false-green guards (plugin never imported / mutation never applied) were verified to fire on purpose-built failures; the import guard *actually fired* mid-work when a stale control referenced a symbol finding 2 had removed.

**Two mutations SURVIVED on the first run** — `period-to-date-omission-hidden` and `fixed-window-datesbetween-counted`. Both were **stale**, targeting the `not_assessed_by_design` path finding 5 deleted. Replaced with `period-to-date-judgement-skipped` and `fixed-window-datesbetween-classified-as-accumulation`, which probe the current mechanism.

New mutations, one per finding:

```
CONTROL   control-noop-rebind                          -> SURVIVED as required
CONTROL   control-identity-wrapper                     -> SURVIVED as required

CAUGHT    f1-axis-role-list-decides-safety             -> test_every_projected_column_is_examined_not_a_curated_axis_list
CAUGHT    f1-empty-projection-reads-as-cleared         -> test_as_of_on_a_measure_only_visual_is_unassessable (+1)
CAUGHT    f1-hierarchy-ignored-in-as-of                -> test_as_of_with_a_hierarchy_projection_is_unassessable (+1)
CAUGHT    f2-declared-type-is-the-only-grain-signal    -> test_date_bins_derived_by_calculation_are_flagged_whatever_their_type (+1)
CAUGHT    f2-lineage-not-followed-transitively         -> test_date_bins_derived_by_calculation_are_flagged_whatever_their_type
CAUGHT    f2-date-named-column-waved-through           -> test_a_date_named_column_with_no_proof_is_unassessable_not_clean
CAUGHT    f3-any-less-than-is-a-running-total          -> test_a_pinned_cutoff_is_not_a_running_total (+1)
CAUGHT    f3-var-bound-never-followed                  -> test_an_as_of_bound_hoisted_into_a_var_is_still_a_running_total
CAUGHT    f4-only-the-first-window-call                -> test_every_window_call_is_assessed_not_just_the_first (+1)
CAUGHT    f5-period-to-date-excused-again              -> test_fact_table_period_to_date_on_a_coarse_axis_is_unassessable (+3)
CAUGHT    f5-every-table-reads-as-a-marked-date-table  -> test_fact_table_period_to_date_on_a_coarse_axis_is_unassessable
CAUGHT    crash-grading-keyed-by-an-unhashable-fieldref-> test_a_grouping_column_is_never_used_as_a_dict_key (+11)
CAUGHT    period-to-date-judgement-skipped             -> test_fact_table_period_to_date_on_a_coarse_axis_is_unassessable
CAUGHT    fixed-window-datesbetween-classified-as-accumulation -> test_a_fixed_window_comparison_is_still_not_an_accumulation

survivors (holes in the suite): none
caught by a different test than claimed: none
non-zero exit with no test outcome (NOT a catch): none
```

(The 19 mutations from the first round all still pass; full table in the run output.)

## Gates, by exit code

| command | exit |
|---|---|
| `ruff format --check` / `ruff check` (`conftest.py scripts tests .github/skills`) | **0 / 0** |
| `pylint scripts` | **0** (10.00/10; was **16** on `C0301` and **24** on `C0302`) |
| `pylint .github/skills/pbip-model-refresh/scripts` · `…/powerbi-ai-readiness/scripts` · `conftest.py` | **0 / 0 / 0** |
| `pytest -q -n auto --dist loadfile` | 3121 passed, 9 skipped; only the 3 engine pins |
| `python tests/mutation_harness_running_total_axis.py` | **0** |
| `python scripts/check_navigation_index.py` | **0** |
| gate on the whole estate | **3** (0 mismatch, 2 stub `unassessable`) |
| gate on all 16 committed examples | **0 × 16**, zero false positives |

`tests/golden/shared-datasource/external-resolves.model.stdout` is **unchanged** by this round.

## Regression net after the change

All 16 committed examples still `NOT_APPLICABLE`/exit 0 — including Superstore, whose CP/PP family is `DATESBETWEEN` and stays out. The estate is byte-identical in verdict to round one: **0 mismatches**, the same 2 bound `RUNNING_SUM` stubs, the same 5 unbound cumulative measures.

- **No `EVALUATE` was executed**, for any of the three shapes. Every verdict remains static.
- **The period-to-date "marked date table ⇒ safe" acquittal is reasoned, not measured.** I did not confirm in Desktop that a `dataCategory: Time` table's other columns are auto-removed on the exact shapes here; that is why the unsafe side returns `unassessable` rather than `mismatch`. If you want that tightened to a mismatch, it needs a live probe first.
- **The `suspect` (date-named, unproven) bucket has no corpus instance.** Fixture-only. It is deliberately non-blocking.
- **Cross-table as-of remains unassessable** — I did not implement relationship-path reasoning, so a `'Date'[Month Start]` axis against an `'Orders'` anchor is still "probe it".
- **The as-of branch still has no real-corpus instance**; zero in the estate, zero in `examples/`. The window branch is the one exercised against real engine bytes.
- Role coverage is still empirical (319 `Category` / 47 `X` / 40 `Rows` / 18 `Columns` observed), but it no longer gates safety — only the printed "Axis is …" clause.
- `_runs/` was read-only throughout; every experiment ran on copies under a scratch dir, since deleted, and the estate was **re-verified pristine** (newest mtime 29/08 22:37, `git status` clean).

Issue #411, opened 2026-08-31
powerbi-semantic-model-gotchas cites two artifacts that do not exist (TabularEditor DLL, bpa.ps1) - and no gate checks skill-internal paths


Guust-Franssens:
`powerbi-semantic-model-gotchas/SKILL.md` tells agents to use two artifacts that do not exist in the repo **or** in the published plugin bundle.

## Measured

```
$ git ls-tree -r --name-only origin/master | grep -Ei "TabularEditor|AnalysisServices.*\.dll"
(nothing)

$ Get-ChildItem -Recurse -Filter "bpa.ps1"
(nothing, repo-wide)

$ Get-ChildItem .github/skills/powerbi-semantic-model-gotchas -Recurse
  SKILL.md          <- the ENTIRE bundle. no scripts/ dir at all

$ Get-ChildItem ~/.copilot/installed-plugins/.../skills/powerbi-semantic-model-gotchas -Recurse
  SKILL.md          <- same in the published copy
```

## The two dead claims

**`SKILL.md:107`** — the offline TMDL validation recipe:
> load `Microsoft.AnalysisServices.Tabular.dll` (ships with Tabular Editor, bundled in this skill''s `scripts/_tools/TabularEditor/`) and call `TmdlSerializer::DeserializeDatabaseFromFolder`

**`SKILL.md:394`** — the BPA entry:
> The bundled `bpa.ps1` runs Tabular Editor with `-G` (silent stdout, exit 0 even on violations)

An agent following either fails at step one, on a path stated with enough specificity to be trusted.

## Why it matters beyond a broken link

The `:107` recipe is the documented **cheap** route to the AMO parser: `Add-Type` against a vendored DLL, no SDK. #388 instead built `tools/tmdl_oracle/` (a C# shim + nuget restore) and made `dotnet` a **critical** preflight dependency (`preflight.ps1:515`).

`tools/tmdl_oracle` itself is fine and correctly committed (2 source files; `bin/`/`obj/` ignored at `.gitignore:110-111`; the `.csproj` carries the AMO version pin that `tmdl_oracle.py:90` verifies at runtime). The question this raises is only whether the SDK dependency was necessary — and it cannot be answered while the documented alternative is fictional. Either vendor the DLL and the recipe becomes true, or delete the recipe and stop implying a route that does not exist.

## The generalizable defect

[...]

Issue #410, opened 2026-08-30
sync_installed_skills publishes from the current working tree, so 'in sync' is a race - and it served UNMERGED guidance to subagents


Guust-Franssens:
`sync_installed_skills.py` publishes the plugin copy **from whichever working tree runs it**. With several worktrees carrying unmerged edits to the same bundle, that makes "in sync" branch-dependent, unstable, and a race — and today it published **unmerged guidance to subagents**.

## Measured 2026-08-30

Four different versions of one shipped file existed simultaneously:

| source | sha256 (first 12) |
|---|---|
| `origin/master` | `385335B71FF0` |
| worktree on `chore/tableau-dashboard-reference-probe` | `D57926B7804A` |
| worktree on `fix/credential-modal-size-only` | `BD35A585AB04` |
| **INSTALLED (what subagents actually read)** | `77539C465D7C` |

`.github/skills/powerbi-semantic-model-gotchas/SKILL.md`. The installed copy was **master + 79 lines** — content that was on no merged branch. Since the plugin copy **shadows** `.github/skills/` for subagents, every subagent spawned in that window read guidance that had never been reviewed or merged.

That is exactly the hazard the shadowing rule exists to prevent, arrived at from the other direction: not a *stale* copy, but a **too-new, unmerged** one.

## It thrashed three times in one day

1. `--check` exit 1 after #388 merged; I synced from the main checkout (then on a feature branch).
2. An agent hit exit 1 again, diffed, and found the drift was **its own round-1 edit** — republished from its worktree.
3. A reviewer hit exit 1 again, with **three** differing hunks.

Each sync overwrote the previous one, and each made every other worktree's `--check` fail. Nobody was wrong; the tool has no notion of *which* version is authoritative.

## The tension this exposes

A PR that edits a shipped skill currently cannot satisfy `sync --check` **and** leave the published copy correct:

- sync from the feature branch -> `--check` passes locally, but subagents get **unmerged** guidance and every other worktree goes red
- do not sync -> `--check` fails, and `preflight.ps1` **blocks**

## Suggested direction

**Compare and publish against `origin/master`, not the current working tree.** The published copy should be *what is merged*, because that is what a subagent should be reading. Then:

- a feature branch with unmerged skill edits does **not** fail `--check`, because the installed copy is correctly master's
- the check becomes stable regardless of which worktree runs it
- publishing becomes a post-merge step (or a CI step) rather than an ad-hoc developer action

[...]

PR #412, opened 2026-08-31
Suppress a dialog only when its content is POSITIVELY benign, in the arbiter too (#406)

### LINKED ISSUE #406: probe_desktop_credential.ps1 still excuses short blocking prose as benign - Password: is suppressed in flight
## The defect

`probe_desktop_credential.ps1` keeps `$MinPromptWords = 5`: once one benign status element is found,
**every unmatched content element with fewer than five words is silently accepted**, so the window is
dismissed without ever being *positively* benign.

Driving the shipped classifiers through the bundle's own harness
(`tests/test_credential_modal_detection.py::classify`, `-LoadDetectorsOnly`), measured 2026-08-30:

| window content (`Title="Refresh"`, `OwnerEnabled=$false`) | t=0 kind / verdict | `-RefreshInFlight` verdict |
|---|---|---|
| `Refresh`, `Evaluating`, **`Please enter your password`** | `benign` / `REFRESH_IN_PROGRESS` | **`$null` — suppressed** |
| `Refresh`, `Evaluating`, **`Password:`** | `benign` / `REFRESH_IN_PROGRESS` | **`$null` — suppressed** |
| `Refresh`, `Evaluating`, `Please enter your password now` (5 words) | `mixed-content` / `DIALOG_UNRECOGNIZED` | `DIALOG_UNRECOGNIZED` |

None of those prompts matches `credential_modal_signature.regex`, so `Test-CredentialModal` does not
rescue them either. A real prompt is swallowed in silence, on the code path #367 added to stop exactly
that — the repo's standing rule is that a credential modal is never worked around.

Only the word count separates row 1 from row 3. **Length is not evidence.**

## Why this is filed rather than fixed in PR #400

PR #400 (issue #376) fixed the identical hole in the Python detector, where blind review found it:

- `MIN_PROMPT_WORDS` removed entirely — any unrecognised content element now vetoes dismissal;
- replaced by `scripts/benign_chrome_signature.regex`, an **enumerated** whole-element allowlist
  (`^&?Cancel$|^&?OK$|^&?Close$`) — a positive claim about specific strings that carry no prompt,
  rather than a claim about their size.

The arbiter was deliberately left alone, following the precedent that produced #376 itself: #367's
author found this defect in the Python half, declared it out of scope, and filed it instead of
silently widening the diff.

There is a second reason, and it is the substance of this issue rather than process: **removing the
amnesty here reverses a reviewed decision and costs a documented capability.**
`test_short_data_labels_beside_progress_text_do_not_block_suppression` exists precisely to keep
`CREDENTIAL_PRESENT` reachable while Desktop shows its own refresh dialog, and a real refresh dialog
lists table names (`Orders`, `Customers`) that are neither progress status nor chrome. Under a total
veto they veto, and `CREDENTIAL_PRESENT` becomes unreachable in that state.

Python could absorb that trade cheaply — its `benign` path only avoids aborting *our own* in-flight
operation, so losing it costs extra exit-3s. The arbiter's `benign` path carries its whole positive
answer. So this needs a decision, not a copy of the Python patch.

## Options

1. **Adopt the chrome allowlist and drop the amnesty** (Python's shape). Costs `CREDENTIAL_PRESENT`
   whenever the refresh dialog lists a table name; updates the positive-control test to assert
   `DIALOG_UNRECOGNIZED`. Safe, loud, and consistent with the Python detector.
2. **Use the control-type signal the arbiter already has.** It harvests `InteractiveTexts`, which
   Python cannot. A rule like *"a non-interactive element that is not recognised status vetoes"*
   would keep `Cancel` harmless without excusing `Please enter your password`. ⚠️ It does **not**
   rescue table names — those are Text elements too — so it does not by itself preserve the positive
   control. Unmeasured.
3. **Accept the gap explicitly**, and document that `CREDENTIAL_PRESENT` is never sufficient alone
   (the one-row data probe is already the gate of record). Weakest option: the failure is silent.

## Acceptance

1. `Refresh` + `Evaluating` + `Please enter your password` is **not** suppressed under
   `-RefreshInFlight`.
2. Same for `Password:` — the fix must not be another length threshold one word lower.
3. Whatever happens to the positive control, the outcome is **stated**: either it still passes and the
   mechanism that saves it is named, or it is changed and the reachability cost is written down.
4. Mutation-test each: restore the amnesty, widen the allowlist to `.*`, remove it entirely. Assert
   each mutation is present on disk, and score on a **named `N failed`**, never a bare non-zero exit.
5. Re-run the `serial` live-WPF tests — they drive a real owned modal and are the only check that the
   change survives contact with UI Automation.

## Related

#376 / PR #400 (the Python half, and the shape of the fix) · #367 (the original size-only detector)

### LINKED ISSUE #417: Credential-modal detection: PowerShell should COLLECT and Python DECIDE - one implementation, so drift is structurally impossible
Two detectors answer one question — *"is this Power BI Desktop dialog a credential wall?"* — in two languages:

| | |
|---|---|
| `.github/skills/pbip-model-refresh/scripts/_credential_modal.py` | Python |
| `.github/skills/pbip-model-refresh/scripts/probe_desktop_credential.ps1` | PowerShell |

#400 rebuilt the Python one on Win32 **window ownership**, deleting six measured-bad proxies (window size, class prefix, zero area, first ownership edge, ownership-chains-only, `MainWindowHandle`). #406/#412 then brought the PowerShell arbiter to match.

## Why a follow-up is needed even though #412 closes the known divergence

⚠️ **Restating logic in a second language is not the same as sharing it, and the drift was measurable.** Before #412, on the *same* real owned 80×60 WPF modal:

```
PowerShell arbiter -> VERDICT: CREDENTIAL_PRESENT, exit 0   # the clean bucket
Python detector    -> DIALOG_UNREADABLE,           exit 3   # correct
```

`Select-DialogCandidate` had kept two of the exact proxies #400 deleted — exclusion by class prefix (`WindowsForms10.Window.8*`) and by size (<100×100).

⚠️ **And the test written to prevent that drift could not detect it.** A reviewer hard-coded the chrome regex in the arbiter and left the shared resource in a **dead comment**: all six "detector sharing" tests still passed. #412 repairs that specific test to require both real detector paths to flip — but the structure that permitted the drift is unchanged.

## The important nuance, conceded during review

**Sharing `classify_dialog` would NOT have prevented the divergence above.** The two detectors already shared the inner classifier; the divergence lived **upstream, in candidate selection**. So "share the classifier" is the wrong fix, and any follow-up that only does that will look correct and change nothing.

## The proposed design

**PowerShell collects; Python decides.**

The arbiter's job reduces to enumerating windows and their Win32 attributes and handing them over. Every judgement — candidate selection *and* classification — happens once, in `inspect_credential_modal`, through its injectable enumerator. There is then no second implementation to drift, because there is no second implementation.

## Acceptance criteria

- [ ] PowerShell performs **no** candidate selection and **no** classification; it collects and forwards.
- [ ] `inspect_credential_modal`'s injectable enumerator is the single decision point for both entry paths.
- [ ] The three-way outcome survives end to end and stays distinguishable **by exit code**: credential wall / not a credential wall / **could not establish either**. ⚠️ `harvest=INCOMPLETE` is a real runtime state on a loaded machine — a partial harvest must never be reportable as "no modal appeared".
- [ ] **Prove sharing behaviourally, not structurally**: mutate one shared resource and demonstrate both real detector paths change together. A test that inspects constants or the inner classifier alone is the vacuous shape that already failed once.
- [ ] Show, by construction, that no input can make the two paths disagree — or state precisely which class still can.
- [ ] `SKILL.md` claims must be true of the shipped code (see #411 for what happens otherwise).

## Do not

- Do not resolve this by re-introducing geometry or class-name proxies anywhere. They are measured-bad; #400 records the evidence.
- Do not weaken the shipped-cap element test to a band — it would stop detecting the regression that is its purpose.

Guust-Franssens:
Fixes #406.
Fixes #417.

---

## ⚠️ UPDATE — the "Port vs share, argued" section below is SUPERSEDED (issue #417)

Everything under *"Port vs share, argued"* and *"Option 2"* is preserved as the record of what was
decided for #406. **It is no longer what ships.** A blind review of that decision ran a **1,536-case
differential** across the two detectors and measured **48 title-related verdict mismatches**,
including **24 where PowerShell exited 3 and Python exited 0** — the gate of record silently clearing
a sign-in modal. Two independent causes:

| cause | measured |
|---|---|
| a title veto applied to the PowerShell half only | owned dialog titled `Password:` with benign body → `DIALOG_UNRECOGNIZED` (exit 3) there, `CREDENTIAL_PRESENT` (**exit 0**) in Python |
| **case sensitivity differing at the LANGUAGE level** — PowerShell's `-ne`/`-notcontains` are case-INsensitive, Python's `==`/`in` are not | title `Refresh` with body `REFRESH` → exit 3 vs **exit 0** |

A language-level difference cannot be fixed once, so a third alignment round would only lengthen the
interval before the next divergence. The branch therefore builds the **seam** the old section listed
as a defended-but-unbuilt alternative:

- **`probe_desktop_credential.ps1` COLLECTS** — Win32 attributes plus the UIA harvest Python cannot read.
- **`scripts/decide_dialog.py` JUDGES** — candidate selection *and* classification, once, in one language.
  `Select-DialogCandidate`, `Get-DialogClassification`, `Test-CredentialModal`, `Get-MainFrame`,
  `Test-RendersNothing`, `Get-DialogTextSet`, `Get-NormalizedText` and `Test-HarvestComplete` are all
  **deleted**, including the ones that were only dead — a second implementation lying around is an
  invitation to rewire it.

### Acceptance: zero mismatches, measured not argued

`test_the_seam_leaves_no_verdict_mismatch_between_the_two_entry_paths` sweeps a deterministic
**2,688-case** corpus (7 captions × 8 bodies × 2 interactive sets × 4 `HarvestComplete` shapes × 3
owner states × in-flight) through the collector path *and* the decider path and compares the whole
record — verdict, kind, exit code, evidence, guidance, candidate count, credential excerpt.

**Measured 2026-09-01: `PBIP_SEAM_SWEEP_STRIDE=1` → 0 mismatches, 12 m 27 s.** The committed suite
walks every 28th case (~30 s, ~273 ms per collector round trip); the env var reproduces the full run.
It carries its own vacuity control (the sampled corpus must actually reach `credential`, `benign` and
`benign-unverified`).

### Two open review findings, closed

1. **A harmless caption substring no longer fabricates a credential wall.** An owned dialog titled
   `Personal Access Token documentation` with benign `Refresh`/`Cancel` content returned
   `CREDENTIAL_MISSING` at **exit 1** in *both* detectors. Convicting now needs CONTENT, and the WPF
   prose join must be **corroborated by the body**: what is matched is the body-only join, what is
   reported is the full prose join including the caption. An unaccounted caption still vetoes
   suppression → `DIALOG_UNRECOGNIZED`, exit 3. Discriminating control included, so the fix cannot be
   "blind the detector".
2. **`SKILL.md` is now true of the shipped code (#411).** The superseded argument is rewritten as the
   measurement that ended it, the deleted PowerShell functions are named as deleted, the
   "completeness exists only in the arbiter" claim is corrected, and the two costs the seam actually
   buys — interpreter discovery, and ~273 ms per decision — are written down rather than assumed away.

### Mutation table (this round)

`tests/mutation_credential_seam.py` drives the **shared** `tests/mutation_harness.py` scorer over 18
**source-level** mutations. Source, not monkeypatch: the collector path runs the decider in a *child
process*, where an attribute patch is invisible and would SURVIVE for a reason unrelated to coverage.
Every file is restored in a `finally` and re-verified byte-for-byte.

**15 CAUGHT, 2 cosmetic controls SURVIVED, 1 absent-anchor control INVALID — every verdict as
expected, exit 0.** The harness earned its keep on the first run: `the-join-reads-the-caption-too`
SURVIVED because the caption test's fixture leaves one body element, so `len(body) > 1` is false and
the mutated line never executes — this repo's third known vacuity mode, found by the harness rather
than by reading. Re-anchored, and the two gaps it exposed are now closed by
`test_an_interposed_button_does_not_break_the_split_signature` (previously covered only by a `serial`
GUI test, so invisible to a routine run) and `test_a_decider_that_answers_nothing_is_indeterminate_never_clean`.

### Validation (this round)

- `pytest .github/skills/pbip-model-refresh/tests/ -m "not serial"` → **334 passed, 3 skipped**
- `pytest .github/skills/pbip-model-refresh/tests/ -m serial` (real WPF + UIA) → **8 passed**
- `pytest tests/ -q` → **2971 passed, 6 skipped, 3 failed** — only the known `test_upstream_repro_pins.py` engine-drift pins
- `python tests/mutation_credential_seam.py` → **exit 0**, 18/18 as expected
- `ruff format --check` / `ruff check` on `conftest.py scripts tests .github/skills` → exit 0
- `pylint` on all four CI roots → **10.00/10, exit 0** each
- `check_navigation_index.py`, `check_agent_capabilities.py`, `sync_agent_conventions.py --check` → exit 0

### What I could NOT verify (this round)

- **The seam's new dependency is real.** `Resolve-PythonExe` tries `$env:PBIP_REFRESH_PYTHON`, then
  `python`/`python3`/`py`. It is exercised with a *stub* interpreter, not on a machine with no Python
  at all, so the "no interpreter" branch is reasoned-and-typed rather than measured end to end. It
  fails **closed** (exit 3) either way, but `CREDENTIAL_PRESENT` genuinely becomes unreachable there.
- **The sweep is structurally blind to anything both paths share** — that is the point of the seam. It
  sees what PowerShell still contributes (JSON serialisation, the 5.1 single-element-array and BOM
  traps, `[int]`/`[bool]` coercions, the `_window_from` field mapping) and nothing else. The
  individual rules are proved by mutation, not by the sweep.
- **The mutation proof is anchor-scoped**, one node id per mutation. It proves the covering test can
  fail; it does not enumerate which *other* tests would also catch each mutation.
- **The `Personal Access Token documentation` false positive is reproduced from the review's report**,
  as a synthesised window. No live Desktop dialog with that caption was captured here.
- Everything in the original "What I could NOT verify" section below still stands.

---
  ## The defect, reproduced before touching anything  Driven through the bundle's own `-LoadDetectorsOnly` harness (`Title="Refresh"`, `OwnerEnabled=$false`):  | window content | before, t=0 | before, `-RefreshInFlight` | after (both) | |---|---|---|---| | `Refresh`, `Evaluating`, `Please enter your password` | `benign` / `REFRESH_IN_PROGRESS` | **`$null`, exit 0** | `mixed-content` / `DIALOG_UNRECOGNIZED`, exit 3 | | `Refresh`, `Evaluating`, **`Password:`** | `benign` / `REFRESH_IN_PROGRESS` | **`$null`, exit 0** | `mixed-content` / `DIALOG_UNRECOGNIZED`, exit 3 | | `Refresh`, `Evaluating`, `Cancel`, `OK`, `Close` | `benign` | `$null`, exit 0 | **unchanged** — the benign path stays reachable | | `Refresh`, `Evaluating`, `Orders` | `benign` | `$null`, exit 0 | `mixed-content` / `DIALOG_UNRECOGNIZED`, exit 3 |  Row 3 matters as much as rows 1–2: without it the fix would have degenerated into "never suppress", which makes the arbiter answer exit 3 for every model and destroys the only reason it exists.  ## The fix  `$MinPromptWords = 5` is **deleted, not lowered** — `Password:` is two words, so any threshold is the same defect one notch down. Its replacement is `benign_chrome_signature.regex`, the **enumerated** whole-element allowlist (`^&?Cancel$|^&?OK$|^&?Close$`) PR #400 introduced. Anything that is neither recognised progress status nor enumerated chrome now **vetoes** suppression and lands in the exit-3 indeterminate band.  ## Port vs share, argued  **Share the vocabulary; port the control flow.** The arbiter now reads *the same file* the Python detector reads, so the one list that can excuse an unexplained element is single-sourced. It does **not** call the Python detector:  - **The two are deliberately divergent**, and `_credential_modal.py`'s own docstring says so — the   arbiter has a prose join, a `benign-unverified` kind and `HarvestComplete`; Python has none of them   because it has strictly less evidence. A cross-language call would have to grow a Python API that   exists only for PowerShell, i.e. the duplication moves rather than disappears. - **It is a recovery instruction.** `refresh_pbip_model.py` prints this script's command line when a   refresh has *already* failed or timed out. Adding "find a Python interpreter" there adds a failure   mode exactly where the tool must not fail, and its fallback would be either fail-open or a   permanent exit 3. - **`-LoadDetectorsOnly` exists to be dependency-free.** `tests/test_skills.py` copies the bundle to a   temp dir with this repo unimportable; the classifiers must run there. - **Latency.** The poll loop classifies every 2 s for up to `-TimeoutSec` (default 75) — ~37   interpreter spawns inside a loop whose entire job is to bound time. The existing child-process   pattern is justified by *killability* (a wedged UIA provider); a classifier spawn buys nothing.  The real cost of duplication — silent drift, which is literally what #406 is — is now a **gate**, not discipline: `test_the_arbiter_and_the_python_detector_share_one_vocabulary` asserts both halves read the same four `*_signature.regex` files and agree on a shared corpus. Mutation M5 proves it fires.  ## The reviewed decision this reverses  `test_short_data_labels_beside_progress_text_do_not_block_suppression` existed to keep `CREDENTIAL_PRESENT` reachable while Desktop shows its own refresh dialog, on the grounds that a real refresh dialog lists table names (`Orders`, `Customers`) which are neither status nor chrome. It is now `test_a_table_name_beside_progress_text_now_vetoes_suppression` and asserts the opposite. Three reasons, in order of weight:  1. **The capability is secondary and already untrusted.** `SKILL.md` and    `docs/data-source-credentials.md` both say the one-row data probe is the gate of record, and record    **three** false `CREDENTIAL_PRESENT` results against a serverless warehouse. 2. **The costs are asymmetric.** Losing it costs a loud, recoverable exit 3. Keeping the amnesty cost a    *silently* suppressed password prompt, against the standing rule that a credential modal is never    worked around. 3. **The capability rests on an INFERENCE; the defect was MEASURED.** No Desktop in this corpus has    confirmed that Power BI's refresh dialog exposes bare table names —    `benign_dialog_signature.regex` records its own provenance as inferred, and `SKILL.md` already flags    the same gap. `Password:` being suppressed was measured. An inferred capability does not outrank a    measured hole.  `REFRESH_PROGRESS`, the shared fixture six precedence tests use as "a dialog that classifies benign", was narrowed to accountable content; the table-name variant is pinned as its own named fixture and test so the cost is visible rather than absorbed.  ### Option 2 (the control-type signal) was rejected, not skipped  The arbiter harvests `InteractiveTexts`, which Python cannot, so "excuse anything interactive" looks like a free upgrade. It is the word count in a better disguise: Databricks renders its authentication-kind chooser as selectable items labelled `Personal Access Token` / `Databricks Client Credentials` — two alternatives that are **in** `credential_modal_signature.regex` *because* they identify a credential dialog. A role-keyed amnesty would excuse that whole family the moment one member is not in the signature. It also would not have rescued the table names it was proposed for; those are Text elements too. Recorded in the script and in `SKILL.md`.  ## Mutation table  Baseline `9 passed`; restore `9 passed`, both with non-empty stdout. Every mutation was asserted present on disk before the run, and scored on **named** `FAILED …::<test>` lines, never a bare non-zero exit.  | # | mutation | caught by (named) | |---|---|---| | M1 | restore the length amnesty (`$MinPromptWords`-equivalent, `-ge 5`) | `…no_length_amnesty_left_on_disk`, `…short_prompts_beside_progress_text…`, `…table_name_beside_progress_text…`, `…only_enumerated_chrome_is_excused_by_the_arbiter`, `…share_one_vocabulary` (5 failed) | | M2 | widen the allowlist to `.*` | `…chrome_allowlist_stays_an_enumeration…`, `…short_prompts…`, `…table_name…`, `…only_enumerated_chrome_is_excused_by_the_arbiter`, **`…only_enumerated_chrome_is_excused_beside_progress_text`** (the *Python* test — direct proof of the shared file), `…unaccounted_prose…` (6 failed) | | M3 | remove the allowlist check entirely | `…only_enumerated_chrome_is_excused_by_the_arbiter`, `…share_one_vocabulary`, `…refresh_progress_dialog_is_not_reported_as_a_credential_block` (3 failed) | | M4 | neuter the veto (`if ($false) { $unaccounted = $t }`) | `…short_prompts…`, `…table_name…`, `…unaccounted_prose…`, `…only_enumerated_chrome…`, `…share_one_vocabulary` (5 failed) | | M5 | drift: inline a private chrome list instead of reading the shared file | `…no_length_amnesty_left_on_disk`, `…share_one_vocabulary`, `…table_name…`, `…only_enumerated_chrome…` (4 failed) |  M1/M2/M3 are the three the issue asks for. A first draft of the on-disk gate matched the bare string `$MinPromptWords =` and failed on the script's *own comment* explaining the removed constant — it is now anchored at statement level and also gates the mechanism (`-split '\s+'`), so rewording a comment cannot silence it.  ## Exit codes  Unchanged: `CREDENTIAL_MISSING` 1, `CREDENTIAL_PRESENT` 0, everything else 3. The verdict **token set** is unchanged too; only the *distribution* moves — more `DIALOG_UNRECOGNIZED` at exit 3, fewer exit 0.  **Consumers traced.** `refresh_pbip_model.py` only *prints* the command (`CREDENTIAL_PROBE`) as a recovery instruction — it never parses the output, so no change. `scripts/probe_desktop_credential.ps1` is a pass-through shim (`& $target @args; exit $LASTEXITCODE`). `scripts/probe_live_source.py` consumes the *Python* detector's verdicts, not this script's. The one wording fix needed outside the bundle was `docs/data-source-credentials.md`'s gloss of `DIALOG_UNRECOGNIZED` ("prose that is not progress status" → "content that is neither recognised progress status nor enumerated chrome"); flagged here because it is outside `.github/skills/pbip-model-refresh/**`.  ## Validation  - `pytest .github/skills/pbip-model-refresh/tests/test_credential_modal_detection.py -m "not serial"` → **170 passed** - **AC5:** the `serial` live-WPF tests (real owned modal, real cross-process UIA) → **7 passed**, in isolation as their docstring requires - Full suite `-q -n auto --dist loadfile` → **3144 passed, 9 skipped, 4 failed**: the 3 known `test_upstream_repro_pins.py` pins, plus `tests/test_check_unit.py::test_cli_model_scope_reports_not_checked_for_unattributable_connection_fixture`, which **passes in isolation and passes twice when its whole file is run under random ordering**. It is an mtime-cache race in `_freshen_clean_fixture_cache()` under `-n auto`, in a file this branch does not touch (`scripts/check_unit.py` is owned by an in-flight sibling) - `ruff format --check` / `ruff check` on `conftest.py scripts tests .github/skills` → clean - `pylint` on all four CI roots → **10.00/10, exit 0** each (the docstring edit briefly pushed `_credential_modal.py` to 1203 lines against `max-module-lines = 1200`; compressed back rather than raising the cap)  ## What I could NOT verify  - **Whether Power BI Desktop's real refresh dialog lists bare table names**, i.e. whether the   reachability cost is real or zero. No Desktop was opened (out of scope by instruction) and this   corpus has never captured one — `benign_dialog_signature.regex` says its own provenance is inferred.   If it does list them, expect `DIALOG_UNRECOGNIZED` where you previously saw `CREDENTIAL_PRESENT`.   Loud, recoverable, and the direction this bundle errs in — but unmeasured, and written down as such. - **The Databricks selector-label claim** underpinning the option-2 rejection is read off   `credential_modal_signature.regex`'s own alternatives plus connector documentation, not off a live   Databricks dialog in this corpus. It does not weaken the fix (it only argues *against* adding an   amnesty), but it is an inference. - **The installed skill-bundle copy is now stale** relative to this branch. Per the brief and #410 I did   **not** run `scripts/sync_installed_skills.py` in any mode. Expect `preflight.ps1` to report   `STALE in plugin` for `pbip-model-refresh` until this merges and the bundle is published from master. - **Real UIA harvest shape.** The classifier tests drive synthesised window objects; only the 7 `serial`   tests touch real UI Automation, and they exercise credential/truncation/wedge paths rather than a   chrome-only benign window.


Guust-Franssens:
## Follow-up commit: `harvest=INCOMPLETE` was one token doing three jobs

Pushed `c55fe05` in response to the field report on
`test_credential_text_beyond_the_element_cap_convicts_when_the_cap_allows_it`.

**The product was right and is unchanged.** It found the window, could not fully harvest it, and
refused to assert either "credential wall" or "no modal appeared". No verdict logic moved.

**The naive repair would have been worse than the flake — measured, not argued.** I forced both shapes
deterministically against the same fixture:

| forced condition | before |
|---|---|
| cap truncation (`-HarvestMaxElements 400`) | `harvest=INCOMPLETE` → `DIALOG_UNREADABLE`, exit 3 |
| contention-killed provider (wedged UI thread) | `harvest=INCOMPLETE` → `DIALOG_UNREADABLE`, exit 3 |

Byte-identical, and identical to the reported failure. A truncation at the **shipped** cap is exactly
the regression that test exists to catch (451 elements against a cap of 2000 cannot truncate
honestly), so `skip on INCOMPLETE` would have made it structurally unable to fail for its own subject.

After — they separate cleanly:

| after | means |
|---|---|
| `harvest=truncated items=400` | the cap cut the read off → **fail**, regression |
| `harvest=no-payload items=1` | the child delivered nothing → not reachable, retry then skip |

⚠️ **My first mechanism claim was wrong and measurement corrected it.** I initially reported case A as a
cap truncation; with the reason surfaced it read `no-payload` — the harvest child was killed before
delivering. `truncated` only appeared once I raised `-HarvestTimeoutSec` to 40. The output comparison
above still holds; the mechanism attribution did not.

**The reason is diagnostic only.** `HarvestComplete` — a strict Boolean read only by
`Test-HarvestComplete` — remains the sole authority over suppression, pinned across every value
production can emit plus two it cannot, with the flag pinned both ways.

## Three shipped tests had this defect, not one

`..._reachable_only_through_textpattern...`, `..._beyond_the_element_cap_convicts...` and
`..._split_by_an_interposed_button...` all assert `CREDENTIAL_MISSING`, only reachable on a complete
harvest, none establishing the precondition. **Two were observed failing on unmutated builds in one
21-minute run.** They now share `_assert_convicts_once_the_window_is_readable`: convict → assert exit 1;
`harvest=complete` but not convicted → **fail** (a real recall defect); forbidden reason → **fail**;
otherwise retry once then skip *saying why*. `CREDENTIAL_PRESENT`/exit 0 is asserted against on every
attempt and never excused.

Also fixed a **pre-existing race** in `test_a_wedged_uia_provider_still_produces_a_verdict`:
`-TimeoutSec 1` gives one poll ~2 s after the invoke and the modal may not be visible yet — measured,
it reported `CREDENTIAL_PRESENT` at exit 0. Correct for a 1-second deadline, not that test's subject.
The deadline is **not** raised (its `elapsed < 25` discriminator is calibrated against it); it now skips
when no dialog was observed, because then the wedge was never exercised.

## Mutations (each asserted on disk, scored on a named `FAILED` line)

| # | mutation | caught by |
|---|---|---|
| N1 | let the reason authorise suppression | 10 failed, incl. `…harvest_reason_is_diagnostic…[complete]` and all 9 `…only_a_real_boolean_true…` |
| N2 | collapse the reason back to one token | `…evidence_line_names_why_a_harvest_stopped` |
| N3 | always report `complete` (a lying reason) | `…harvest_reason_names_the_actual_payload_shape[int-flags,flags-missing]` |
| N4 | drop the item count | `…evidence_line_names_why_a_harvest_stopped` |
| N5 | conflate `not-attempted` with `no-payload` | `…partial_harvest_is_never_reported_as_no_modal_appeared` |

**N5 survived the entire offline suite first time round** — `ConvertTo-ProbeWindow` sits below the
`-LoadDetectorsOnly` seam, so only a live UIA test can see it. That gap is now closed by a serial test.

## Port vs share — the evidence sharpens the decision rather than flipping it

Two detectors disagreeing about completeness would indeed be worse than either alone. They cannot:
completeness exists **only** in the arbiter. `classify_dialog` takes no completeness input, and on the
Python side a throwing text read fails the *whole* enumeration (`Win32EnumerationError` →
`unknown_reason`), so a partial read never reaches its classifier. Sharing would mean *adding* a
concept to Python that only PowerShell can produce.

What must be identical is weaker and is tested on both sides: **an unestablished read never reaches the
clean state** — arbiter via `test_only_a_real_boolean_true_can_authorise_suppression` (9 shapes, under
`-RefreshInFlight`) plus process-level `test_a_partial_harvest_is_never_reported_as_no_modal_appeared`;
Python via the raise path → `unknown_reason`. Same invariant, different mechanism, neither able to
produce a silent clear.

## Verification

Offline bundle suite **319 passed, 3 skipped**. Serial **5 passed, 3 skipped, 0 failed** (the skips are
the pre-existing helper skip under heavy concurrent load — previously these *failed*). ruff clean;
pylint **10.00/10 exit 0** on all four roots.

## Not done, deliberately

**The harvest budget is not tuned.** Derived from `-TimeoutSec` and clamped 2..8 s, it sits close to
what a 451-element WPF window costs: the same command produced both `no-payload` and a complete read
**without** six-agent load, so this is not purely a load story and the margin is thin. Expect
`DIALOG_UNREADABLE` more often than the happy path suggests on a dense customer dialog. Tuning it is a
real trade (a longer budget is a longer poll) and belongs in its own issue.

Guust-Franssens:
## Re: the correction — accepted on 1 and 2; I have evidence against the prescribed fix shape

Your corrections 1 and 2 are right and I accept both. On the fix shape I disagree, and I tested it
rather than argued it. `c55fe05` + `98cd917` already implement this; it was in scope because the
arbiter work required editing those tests.

### Where we agree
Precondition first, then the exact assertion. Skip when the subject is unobservable. Never band-ify
this test, because its subject *is* the cap.

### Where I have evidence against you
Your rule was: *"A harvest that came back INCOMPLETE is the same situation — skip (or retry), do not
fail, and do not pass."*

**A `truncated` harvest at the shipped cap IS the regression.** 451 elements against a cap of 2000
cannot truncate honestly, so `skip on INCOMPLETE` skips on the exact signature of the defect — the
vacuity mode you were warning against, reached by the prescription itself. Measured by simulating a
real regression (`$HarvestMaxElements` defaulted to 400):

| test shape | detects a real cap regression |
|---|---|
| assert the band only | never — `DIALOG_UNREADABLE` is in the band |
| skip on any `INCOMPLETE` | never — it skips on `truncated` |
| **reason-aware, retry ×3** | **3 / 3 runs failed, on the named test** |

That discrimination is only possible because the arbiter now reports *why* it stopped
(`harvest=truncated items=400` vs `harvest=no-payload items=1`); before, both printed
`harvest=INCOMPLETE` and were byte-identical.

### A bug of mine that this experiment caught
The first implementation detected the simulated regression in only **1 of 3** runs. Cause:
`_run_probe_against_wpf_modal`'s own `pytest.skip` fires on fixture-startup lag, and **a skip inside a
retry loop aborts the test rather than continuing it** — so I had one attempt, not three. Owning that
skip (`require_refresh_invoked=False`) plus 3 attempts:

- simulated regression: **1/3 → 3/3** detected
- unmutated serial suite: **5 passed / 3 skipped → 8 passed / 0 skipped**

I would not have found this by reasoning; it took mutating the cap and running it.

### On your correction 2
Accepted — CPU pressure is not the mechanism, and your 10-spinner / 3-3 pass result is better
controlled than anything I ran. One datum to add rather than contradict: I measured `harvest=truncated
items=400` only after raising `-HarvestTimeoutSec` to 40; at the derived budget (clamped 2..8 s) the
same 400-element harvest came back `no-payload`. So the harvest cost sits close to its own budget
independently of who else is on the machine. That is consistent with UIA contention being the trigger,
and it also means the margin is thin on a dense customer dialog. **I cannot separate "UIA contention"
from "ordinary variance" with my data** — yours is the better-controlled experiment and I defer to it.

### On the helper docstring you quoted
Agreed, and it is now cited in `SKILL.md` — with the boundary drawn: "assert on the band" is the right
default for the sibling tests, whose subject is not the harvest, and the wrong rule for this one.
(Minor: it is at line **3441/3446** in this branch, not 1889.)

### Three-way outcome, still intact
`CREDENTIAL_MISSING` exit 1 / `CREDENTIAL_PRESENT` exit 0 / could-not-establish exit 3, with the third
never folded into the second — `test_a_partial_harvest_is_never_reported_as_no_modal_appeared` drives a
real wedged UIA provider and asserts exit 3, never 0, and every conviction test asserts against exit 0
on every attempt before any skip. The arbiter represents `harvest=INCOMPLETE` natively
(`HarvestComplete` → `benign-unverified` → `DIALOG_UNREADABLE`) and now names the reason. The **Python**
detector cannot represent it — by construction, a partial read fails the whole enumeration and routes
to `unknown_reason`, also exit 3 — and that gap is named in `SKILL.md` rather than papered over.

No apology needed; the correction cost me one experiment and bought a 3× improvement in detection.

Guust-Franssens:
## All four findings fixed. I reproduced every one before touching anything — you were right on all four.

Commits `ca55484` · `de16c29` · `d901f59` · `3a6e5c2`. Findings 1 and 2 were **silent clears (exit 0)** — the worst outcome in this domain.

### 1 (HIGH) — hostile TITLE discarded
`Content` drops text equal to the caption so a reassuring caption cannot *authorise* suppression (round 1's defect). The exclusion was **one-directional** and the other half went unclosed for four rounds. Measured, title `Password:`, body only `Refresh`/`Cancel`:

| | before | after |
|---|---|---|
| t=0 | `benign` / `REFRESH_IN_PROGRESS` | `mixed-content` / `DIALOG_UNRECOGNIZED`, exit 3 |
| `-RefreshInFlight` | **verdict `None`, exit 0** | `DIALOG_UNRECOGNIZED`, exit 3, evidence `Password:` |

Rule now stated: **captions may ACCUSE, never EXONERATE.** A title still cannot set `benignHit`, so `benign-title-only` is unchanged (`test_a_recognised_title_still_cannot_authorise_suppression` guards the over-correction).

### 2 (HIGH) — the deleted proxies had survived
Confirmed on identical input, one real owned **empty 80x60** modal:

| | before | after |
|---|---|---|
| arbiter | **exit 0, `candidates: 0`** | `DIALOG_UNREADABLE`, exit 3 |
| python | `DIALOG_UNREADABLE`, exit 3 | unchanged |

Same for a `WindowsForms10.Window.8*` owned dialog (exit 0 → exit 3). **The detectors now agree.** Ported `main_frame`/`renders_nothing`: `OwnerHwnd` added to the Win32 shim, ownership walked transitively, roots **enumerated**, exactly one identifies the frame, anything else → `$null` which **excludes nothing**. Only exclusions left: the identified frame, and *unowned AND zero-area*. Nothing excluded for size, class or name. The credential prepass also excludes the identified frame, so a report titled `Account Key` can no longer fabricate a hard stop.

`test_the_main_window_and_tiny_helper_windows_are_still_not_candidates` asserted the deleted proxies and is replaced by two tests asserting the new rule.

### 3 (MEDIUM) — the sharing test now tests sharing
You were right that agreement proves nothing, **because a copy agrees too**. It now mutates `benign_chrome_signature.regex` to `^Orders$` in a scratch copy of the bundle and requires **both real detector paths to flip on both windows**. **Your exact mutation now FAILS it** (R1 below).

### 4 — already fixed in `c55fe05`/`98cd917` before your review landed; re-verified. Not band-ified.

## Mutations (anchors asserted on disk; named `FAILED` lines; non-empty stdout)

| # | mutation | caught by |
|---|---|---|
| **R1** | **your mutation** — hard-code chrome regex, filename in a dead comment | `…share_one_vocabulary` |
| R2 | drop the title veto | 4× `…hostile_TITLE…` params |
| R3 | restore the size proxy | `…size_and_class_are_no_longer_exclusions_at_all` |
| R4 | restore the class proxy | `…size_and_class…` + `…identified_main_frame…` |
| R5 | frame ambiguity excludes the first root | `…identified_main_frame_is_the_only_window_excluded` |

Baseline 10 passed / restored 10 passed.

## Two regressions I caused and had to fix

1. **I broke your serial-marker gate.** Moving the live call into a shared helper added indirection `tests/test_parallel_test_loop.py` cannot AST-follow, dropping 3 live tests from the serial tier (`assert 5 >= 7`; deselect 14→11). Fixed **on my side** — the helper takes a zero-arg closure so the live call stays lexically in each test body. I deliberately did **not** widen the gate that caught me. Mutation-proved: removing one `@pytest.mark.serial` now fails `test_every_live_ui_test_declares_itself_serial`, naming that exact test.
2. **`WIP-STATE.md` broke the navigation-index gate** (root markdown must be in `docs/INDEX.md`). Deleted — it was always marked DELETE BEFORE MERGE.

## Share vs port — a defended alternative, and I concede the sharper half

Sharing `classify_dialog` **would not have prevented finding 2**: the divergence was in *candidate selection*, upstream of the shared classifier. Only porting the whole decision model fixed it. What cannot be shared is the *input* — the arbiter exists to read UIA text Python structurally cannot get.

But I concede the stronger point and have written it into `SKILL.md` rather than arguing it away: **a genuine sharing seam does exist** — PowerShell owns *collection* (Win32 + UIA), Python owns *decision* via `inspect_credential_modal`'s injectable enumerator — and it would make drift **structurally impossible** rather than merely detectable. That is a larger change than this issue. The mutation test is what makes the ported version honest meanwhile. If you want the seam built, say so and I will file/do it.

## Gates

Full suite **3167 passed, 9 skipped**, only the 3 documented `test_upstream_repro_pins`. Serial live-WPF **8 passed** (real Win32 shim change verified against real windows). `tests/test_parallel_test_loop.py` **20 passed**. ruff clean; pylint **10.00/10 exit 0**. `SKILL.md` corrected — including deleting the claim you flagged as untrue.

Desktop PID 42372 untouched; only my own WPF fixtures created, all killed in `finally`.

Guust-Franssens:
Closing — not because the concern is wrong, but because this PR is not shippable and the cost is not justified by the risk.

**State:** tip commit is literally \567fbd WIP - UNVERIFIED: in-flight test work on the caption/schema fixes\, **11 behind master**, untouched since 09-01 16:15. Three blind-review HIGHs were never dispatched. CI is green — which on a tip its own author marked UNVERIFIED is exactly the false-green this repo keeps getting caught by, not evidence of health.

**Measured cost/benefit:** +2730/-415 for **338 lines of product code**, against 383 lines of proof machinery and 1450 of plain tests. That is 5.4 lines of test per line of product on a speculative refinement — whether a credential dialog's prose is 'positively benign' — with no customer report behind it.

⚠️ The underlying concern is real and is already tracked: **#406** (the arbiter still excuses short blocking prose as benign) and **#417** (PowerShell should COLLECT, Python should DECIDE — one implementation, not two). Those keep the problem visible without carrying an unverified 3,145-line branch.

If a customer hits a suppressed credential modal, reopen from #406 with that report as the evidence.

Issue #107, opened 2026-08-13
The conversion engine is unpinned and installed twice at different versions - with no record of which one built a bundle


Guust-Franssens:
## The problem

The deterministic conversion engine is a separate project that **this repo does not pin**. It resolves at runtime, so it can change underneath a migration with **no diff on our side** — and right now this machine has *two different versions installed at once*, with upstream already past both.

Measured 2026-08-12:

| tree | VERSION | resolved by |
|---|---|---|
| installed plugin `~\.copilot\installed-plugins\tableau-collection\tableau-fabric-skills\skills\tableau-migration\` | **2.113.0** | `estate_survey.py`, and anything using `engine_scripts_dir()`'s first hit |
| sibling clone `..\tableau-fabric-skills\skills\tableau-migration\` | **2.126.0** | `run_estate.py --engine` default |
| upstream `Yarbrdab000/tableau-fabric-skills` HEAD | newer still (`c6b3608a` vs local `35429806`) | — |

**These are not equivalent.** A controlled A/B on the same workbook (`book_6-1-Maps.twbx`, sha256 `1885eea8…`), same command, two engines:

| worksheet | 2.113.0 | 2.126.0 |
|---|---|---|
| Pie Chart Map | `pieChart` — geography dropped | `azureMap` bubble map |
| Density Map | **no visual emitted** | `azureMap` + `heatMapLayer` |
| Filled / Symbol / Dark / Tooltip | `shapeMap` ×4 (Bing, deprecated) | `azureMap` ×4 |

So a single migration can produce materially different reports depending on which step resolved which engine — and nothing in the run output makes that visible.

## Why this is not hypothetical

This exact gap has already caused **two retracted defect reports** in this project:

1. The engine moved 2.60.0 → 2.72.0 unnoticed; issues were nearly filed against behaviour it had already replaced.
2. The engine moved 2.113.0 → 2.126.0 **mid-dry-run** (13 releases), and a defect report drafted against the older tree described behaviour that no longer existed.

Both were caught only by a manual `git fetch`. That is not a control.

`AGENTS.md` already says to *"record the resolved engine path and its VERSION in the run log"*, but nothing enforces it, and the two-installations problem is not covered at all.

## Suggested fix

1. **Record the resolved engine path + VERSION in every bundle** (`engine-output-receipt.json` is the natural home) so an artifact can always answer "what built me?".
[...]

PR #414, opened 2026-08-31
Fixes #410: publish the MERGED skill bundles, not whichever worktree runs the sync

### LINKED ISSUE #410: sync_installed_skills publishes from the current working tree, so 'in sync' is a race - and it served UNMERGED guidance to subagents

Guust-Franssens:
Fixes #410.

## The decision

**Compare and publish from a merged ref, never the working tree.** Both the `--check` comparison and
the publish now read `origin/HEAD` → `origin/master` → `origin/main` (or `--ref`), exported with
`git archive` and built with **that ref's** `build_plugin.py` — so *which bundles ship* is the merged
tree's decision too, not the branch's. Taking content from the ref but the file list from the
worktree would be a third, hybrid answer to "what ships", which is the ambiguity this issue is about.

| question | decision | why |
|---|---|---|
| network? | **No fetch by default**; `--fetch` opts in, and never fails the run | `preflight.ps1` runs this on **every** migration start, and `AGENTS.md` already settled that a mandatory round trip there is a tax on every run. The failure modes are also asymmetric: a stale local `origin/master` still names a **real merged commit**, so what it publishes stays reproducible from a commit — the property the whole design protects. Unmerged worktree content is reproducible from nothing. The resolved ref + short sha is printed on every verdict, so lag is visible. |
| detached HEAD | works, unchanged | HEAD is never read. The issue's own "Restored" note had to build a detached `origin/master` worktree by hand; that is no longer necessary. |
| no `origin` / no resolvable ref | **refuse, exit 5**, publish nothing, name `--from-worktree` | A silent fallback *is* the bug: it would publish unmerged content on exactly the machines whose git layout is unusual, and say nothing. Same rule as `engine_source.py` (#107). |
| shallow clone | works | only the tree at the ref is needed, never history. Covered by a test that asserts `.git/shallow` exists first. |
| should `--check` **fail** on unmerged local edits? | **No — report, exit 0.** Confirmed | The check's question is "is the published copy what a subagent should read", and a branch edit does not make it wrong. Failing it is what blocked migrations. It is still **reported** loudly (a `SYNC: NOTE`, and `local_unmerged` in `--json`), because an operator measuring against edits that are not live is wrong the same way a stale bundle makes them wrong. |

## ⚠️ `preflight.ps1` had to change — it carried the same defect independently

This is the load-bearing part: preflight **never called** `sync --check`. It hashed
`$repoRoot\.github\skills\<name>\SKILL.md` against the install itself, as a **critical** check. So it
had its own notion of authority, and that notion was "whichever worktree ran preflight" — an in-flight
skill PR blocked every migration in every other checkout. Fixing only the Python would have fixed
nothing operationally.

It now consumes `--check --json`. One authority, one implementation — and it compares **every file in
the bundle**, not only `SKILL.md`. Unmerged local edits become a new **`optional`** check. A new
`no_ref` branch reports *unverified*, distinct from *stale*.

## Both directions proven, end to end against the real preflight

| direction | result |
|---|---|
| installed **stale** vs master | `[MISS] skill bundles match published plugin  STALE in plugin vs refs/remotes/origin/HEAD @ 9cefe68aac7c: powerbi-semantic-model-gotchas/SKILL.md` → preflight **exit 1** |
| **unmerged local edit** on the branch | `[OK] skill bundles match published plugin  in sync with refs/remotes/origin/HEAD @ 9cefe68aac7c` + `[WARN] (optional) skill bundles: local edits vs merged  1 unmerged local edit(s)` → preflight **exit 0, "Ready to migrate"** |

## Mutation table — 20/20 CAUGHT

Every mutation asserted to have applied (snippet must match **exactly once**; file restored and
re-hashed afterwards). A verdict rests on a **named `FAILED <suite>::<test>` line**, never an exit
code; empty stdout, a collection `ERROR`, or an xdist worker death are scored **INVALID**, not CAUGHT.

| mutation | what it breaks | verdict |
|---|---|---|
| `M01-no-ref-falls-back-to-worktree` | silent fallback when no origin exists | **CAUGHT** |
| `M02-always-read-the-worktree` | authority reverts to whichever worktree runs it (the original defect) | **CAUGHT** (10 tests) |
| `M03-never-report-local-edits` | unmerged edits never surfaced | **CAUGHT** |
| `M04-unmerged-edits-fail-the-check` | an unmerged edit blocks preflight again | **CAUGHT** |
| `M05-ref-candidate-order-reversed` | `origin/HEAD` no longer preferred | **CAUGHT** |
| `M06-worktree-override-is-silent` | the unreviewed-guidance override stops announcing itself | **CAUGHT** |
| `M07-fetch-flag-does-nothing` | `--fetch` skips the refresh it promises | **CAUGHT** |
| `M08-fetch-runs-unconditionally` | every preflight pays a network round trip | **CAUGHT** |
| `M09-fetch-failure-becomes-fatal` | offline blocks the gate | **CAUGHT** |
| `M10-generator-taken-from-the-worktree` | what ships decided by the branch | **CAUGHT** |
| `M11-extra-installed-files-ignored` | a file deleted upstream lingers forever | **CAUGHT** |
| `M12-drift-no-longer-fails` | **a genuinely stale bundle stops being caught** | **CAUGHT** |
| `M13-json-omits-the-drift-list` | preflight cannot name what drifted | **CAUGHT** |
| `M14-no-json-on-the-no-ref-path` | preflight sees an unparseable verdict | **CAUGHT** |
| `M15-preflight-hashes-the-worktree-again` | preflight re-acquires its own authority | **CAUGHT** |
| `M16-preflight-stale-check-downgraded` | a stale install no longer blocks | **CAUGHT** |
| `M17-preflight-local-edits-become-critical` | an in-flight skill PR blocks every migration again | **CAUGHT** |
| `M18-in-sync-plus-unmerged-becomes-drift` | an unmerged edit fails `--check` even when the install is correct | **CAUGHT** |
| `M19-fetch-line-pollutes-json` | `--fetch --json` prints a human line before the JSON | **CAUGHT** |
| `M20-preflight-cannot-tell-no-ref-from-in-sync` | an unresolvable ref reports "in sync with " while failing | **CAUGHT** |

M19 and M20 are **defects found and fixed during the work**, not hypotheticals: `--fetch` printed a
human line that `--json` did not suppress (preflight would have read that as "did not report" — an
unverified, silently shadowing bundle), and the `no_ref` payload fell through to the drift branch and
rendered as `in sync with ` *while failing*.

## Corrected while building this

The first draft's docstring claimed `git archive` emits raw blobs, so publishing from a ref would be
machine-independent. **Measured and wrong**: with `core.autocrlf=true` and no `.gitattributes`,
`git archive` **does** apply the CRLF conversion — the exported `powerbi-ai-readiness/SKILL.md` is
19335 bytes and byte-identical to the checked-out file, while `git cat-file` gives 19064 bytes of LF.
The comment now says so. Practical consequence: moving the source to a ref changed *which commit* is
published without rewriting every installed file's line endings — the real publish copied **1** file.

## Verification

| gate | result |
|---|---|
| `ruff format` / `ruff check --fix` | clean |
| `pylint scripts` | **10.00/10** |
| `check_navigation_index.py` | exit 0 |
| `sync_agent_conventions.py --check` | exit 0 |
| `tests/test_sync_installed_skills.py` | **26 passed** |
| tier-1 loop (`-q -n auto --dist loadfile -m "not (serial or timing)"`) | **3166 passed**, 9 skipped, 3 failed — all three the known `test_upstream_repro_pins.py` engine pins |
| preflight parse + live run | OK, both directions above |

## Installed plugin state

**Left IN_SYNC with `origin/master`** (`refs/remotes/origin/HEAD @ 9cefe68aac7c`), verified by
`--check` → exit 0 after the last change. It was found *stale* at the start of this work (the
installed `powerbi-semantic-model-gotchas/SKILL.md` matched two feature worktrees, not master — the
issue's race, still live), and the only publish performed copied **one** file, from `origin/master`.

## What I could NOT verify

* **Non-Windows / other `core.autocrlf` settings.** The line-ending measurement is from this machine
  only. The installed plugin copy is per-machine and never shared, so it should not matter, but that
  is reasoning, not a measurement.
* **A remote whose default branch is not `master`.** `origin/HEAD` preference is tested against a
  fixture, not against a real renamed-default remote.
* **`--fetch` against a real network failure** — simulated by pointing `origin` at a non-existent
  path, not by an actual offline machine or a timeout.
* **CI on Ubuntu.** The preflight assertions are source contracts by design (as `test_preflight_contract.py`
  already documents); the executed preflight evidence above is Windows-only.
* **Whether a `no_ref` state is genuinely reachable in normal use.** It is tested, but every real
  checkout here has an `origin`.
* **The full serial tier-2 gate (`pytest -q`)** was not run: the tier-1 run took 8 min while an
  unfiltered `-n auto` run of the whole suite took 2h36m and produced extra failures. Those extras
  (`test_check_unit`, `test_tmdl_oracle`, `test_skills`) **pass in isolation with this change
  applied**, and `docs/parallel-test-loop.md` says tier 1 must deselect `serial`/`timing` — running
  that scheduler without the `-m` filter is an unsupported combination. I am confident they are
  contention artefacts, but I did not reproduce a clean baseline of the unfiltered run.


Guust-Franssens:
All four findings fixed, plus two holes the hardened mutation re-run found in my own tests.
Pushed incrementally: `6651f52` (F1–F3), `3325658` (F4), `d2a90b8` (mutation holes + docs).

## 2. HIGH — worktree controlled discovery and the inventory — FIXED

You were right that I closed the content channel and left the destination and file list open.
Three changes:

- **The inventory now comes from the pinned build**, and is passed into discovery:
  `discover_skill_plugin(plugin_root_override=…, bundles=bundles)`. Discovery runs *after* the
  export, not before it. `skill_plugin_source.py` keeps `SHIPPED_SKILLS` only as the standalone-CLI
  default, with a docstring naming your reproduction.
- **preflight no longer runs `skill_plugin_source.py` at all.** The entire skill block is driven by
  one `--check --json` verdict, which carries `plugin_root`, `identity`, `bundles`, `missing`,
  `changed`, `extra`, `local_unmerged`. Asserted by source contract: no `shipped_skills` read, no
  `skill_plugin_source.py` invocation.
- **The data loss is fixed at the root, not around discovery.** `diff_tree` takes a `scope`, so the
  DELETION set is bounded to the merged bundle directories. Deleting a file the merged tree never
  owned is never this script's job, so the bound is unconditional rather than a safety net.
  `test_a_bundle_the_merged_tree_does_not_own_is_never_deleted` plants a foreign bundle in the
  install and asserts it survives a publish.

**Your reporting sub-finding was also right, and the fix is bigger than a pathspec.** `local_unmerged`
missed a branch-added bundle because a path that does not exist at the merged commit is in no merged
bundle's pathspec — widening to `.github/skills` would have flagged every unrelated edit to the
repo-local bundles this plugin does not publish. It is now the **diff between the merged build and a
build of this working tree**, which catches edited content, added/removed bundles, untracked files
*and* generator changes at once, with no noise. Exit code unchanged, as you confirmed it should be.

## 1. HIGH — stale `origin/HEAD` — FIXED

`--fetch` now asks the remote directly, `git ls-remote --symref origin HEAD`, and pins that commit.
I chose `ls-remote` over `git remote set-head --auto` because it writes nothing into the repository.

Offline it genuinely cannot be resolved, so instead of silence the verdict carries
`default_verified: false` and `default_alternatives` — any default-branch ref whose commit differs
from the chosen one. Only *differing* refs count, so an ordinary clone stays silent
(`test_a_normal_clone_raises_no_stale_default_noise`). `--ref` counts as verified.

## 3. MEDIUM — commit not used — FIXED

`export_ref_tree` and the divergence diff both take `source.commit`; `source.ref` is a display label.
`test_a_ref_that_advances_mid_run_does_not_change_what_is_published` reproduces your case by pushing
a new commit to `origin/master` *between* resolution and export. Confirmed live: the sync now reports
`c34adb555e39`, the commit that landed while this PR was in review.

## 4. MEDIUM — untested preflight decision — FIXED

Extracted to **`Get-SkillBundleVerdict`**, dot-sourced by AST and executed against real
`--check --json` output with a controlled plugin root, mirroring `Get-DesktopPinVerdict`. Eight
executed cases: in-sync passes; one tampered byte fails; a removed bundle fails the inventory row
separately; an unmerged edit keeps the critical row green while the optional row reports it; an
unreported verdict and an unresolvable ref are both UNVERIFIED; a missing plugin stays non-blocking.
Tiers stay literal at the call site so `test_preflight_contract.py`'s tier assertions still bind.

**Your exact surviving mutation is now caught**: `M28-preflight-stale-verdict-inverted` →
`test_preflight_passes_an_in_sync_install`.

## ⚠️ Two of my new tests SURVIVED the re-run — both fixed

The hardened sweep found holes in tests I wrote this afternoon, both the **sibling-branch** shape:

| survived | why | fix |
|---|---|---|
| `M33` downgrade the missing-plugin row to `critical` | the assertion looked for the `'recommended'` string, which the `compared` branch also emits | regex **every** emission of that row; all must be `recommended` |
| `M34` replace the stale row's verdict with a literal `$true` | `$skills.Merged.Ok` still appeared elsewhere in the file | regex **every** `Add-Check` emission of each skill row; each must take its ok-argument from the matching function field |

This is precisely the trap `test_preflight_contract.py::_assert_add_check_tier` documents, reproduced
by me in the same file that cites it. Both are CAUGHT after the fix.

## Mutation table — 29/29 CAUGHT

Scoring hardened for your finding-inside-a-finding: **a named `FAILED` line is necessary but not
sufficient.** A CAUGHT verdict also requires that the suite actually ran (`passed > 0`), no collection
`ERROR`, no dying xdist worker, and **no interpreter-level crash signature**
(`NameError|ImportError|ModuleNotFoundError|SyntaxError|IndentationError`) anywhere in the output —
the class you showed also emits a named `FAILED`. Every snippet is asserted to match exactly once and
every file re-hashed after restore.

| mutation | breaks | verdict |
|---|---|---|
| `M01-no-ref-falls-back-to-worktree` | silent fallback when no origin exists | **CAUGHT** |
| `M02-always-read-the-worktree` | the original defect | **CAUGHT** |
| `M03-never-report-local-divergence` | divergence never surfaced | **CAUGHT** |
| `M05-ref-candidate-order-reversed` | `origin/HEAD` not preferred | **CAUGHT** |
| `M06-worktree-override-is-silent` | the override stops announcing itself | **CAUGHT** |
| `M07-fetch-flag-does-nothing` | `--fetch` skips the refresh it promises | **CAUGHT** |
| `M08-fetch-runs-unconditionally` | every preflight pays a round trip | **CAUGHT** |
| `M09-fetch-failure-becomes-fatal` | offline blocks the gate | **CAUGHT** |
| `M10-generator-taken-from-the-worktree` | the branch decides what ships | **CAUGHT** |
| `M11-extra-installed-files-ignored` | an upstream deletion lingers | **CAUGHT** |
| `M12-drift-no-longer-fails` | **a stale bundle stops being caught** | **CAUGHT** ¹ |
| `M13-json-omits-the-drift-list` | preflight cannot name the drift | **CAUGHT** |
| `M14-no-json-on-the-no-ref-path` | unparseable failure verdict | **CAUGHT** |
| `M18-in-sync-plus-unmerged-becomes-drift` | an unmerged edit fails a correct install | **CAUGHT** |
| `M19-fetch-line-pollutes-json` | human line before the JSON | **CAUGHT** |
| `M21-export-uses-the-moving-ref-name` | **F3** a ref advancing mid-run changes the publish | **CAUGHT** |
| `M22-fetch-does-not-verify-the-default` | **F1** `--fetch` publishes the stale default | **CAUGHT** |
| `M23-remote-default-never-resolves` | **F1** the advertised HEAD is never read | **CAUGHT** |
| `M24-stale-default-is-never-flagged` | **F1** offline staleness is silent | **CAUGHT** |
| `M25-discovery-uses-the-worktree-inventory` | **F2** the branch picks the destination | **CAUGHT** |
| `M26-deletion-set-is-unbounded` | **F2** the data loss you reproduced | **CAUGHT** |
| `M27-missing-bundles-never-reported` | **F2** the inventory row can never go red | **CAUGHT** |
| `M28-preflight-stale-verdict-inverted` | **F4, your surviving mutation** | **CAUGHT** |
| `M29-preflight-inventory-always-ok` | a partial install reads complete | **CAUGHT** |
| `M30-preflight-local-edits-always-ok` | divergence stops being reported | **CAUGHT** |
| `M31-preflight-loses-the-no-ref-branch` | an unresolvable ref reads as a comparison | **CAUGHT** |
| `M32-preflight-unreported-verdict-passes` | no verdict is treated as a pass | **CAUGHT** |
| `M33-preflight-missing-plugin-becomes-critical` | no-plugin machines blocked | **CAUGHT** ² |
| `M34-preflight-rows-bypass-the-verdict` | the verdict never reaches the row | **CAUGHT** ² |

¹ First scored **INVALID** — `passed > 0` is meaningless under `-x` when the very first test is the
catcher. Re-scored against the whole file (`I410_NO_X=1`): CAUGHT by
`test_stale_installed_copy_is_still_caught_as_drift`, 43 still passing. I kept the `passed > 0` rule
rather than dropping it, because your point stands.
² SURVIVED on the first run; CAUGHT after the test fixes above.

## Gates

`ruff` clean · `pylint scripts` **10.00/10** · `check_navigation_index.py` 0 ·
`sync_agent_conventions.py --check` 0 · **48** tests in this file ·
tier-1 loop **3322 passed**, 9 skipped, only the three excluded `test_upstream_repro_pins.py` engine
pins · preflight parses and runs clean in both directions.

## Installed plugin state

Left **IN_SYNC** with `refs/remotes/origin/HEAD @ c34adb555e39`, `--check` exit 0, all of
`changed`/`extra`/`missing`/`local_unmerged` empty. No machine-wide publish beyond that.

- **A real renamed-default remote in the wild.** Reproduced against a bare fixture whose `HEAD` is
  moved with `git --git-dir … symbolic-ref`; not against GitHub's own rename flow.
- **`ls-remote` against an authenticating remote.** The fixtures use local paths, so credential
  prompts and the 60 s timeout path are untested.
- **The `default_alternatives` heuristic on a repo that legitimately keeps both `master` and `main`
  at different commits.** It would emit an advisory note there; correct-ish, but noise I have not
  seen in practice.
- **Non-Windows / other `core.autocrlf`.** The `git archive` line-ending measurement is this machine
  only; an installed plugin copy is per-machine, so I believe it is moot, but that is reasoning.
- **CI on Ubuntu for the executed PowerShell tests.** They skip when neither `pwsh` nor `powershell`
  is on PATH — so on a runner without PowerShell, finding 4's coverage silently drops to the source
  contracts. Worth flagging if that matters to you.
- **The full serial tier-2 gate (`pytest -q`).** Not run; the parallel-fragility question you closed
  is the reason I did not re-litigate it.

Guust-Franssens:
## Blind review of the rebased branch — VERDICT: **BLOCKED**

Reviewed at the exact tip `8c50f0f8e1140b7c2c33b675ace9574450f19fd0`, base `dab3acb`. Reviewer was
given only the issue and the diff — never the author's reasoning. Every finding below came from a
controlled experiment against the public CLI, not from reading the code sympathetically.

### The rebase-conflict resolution is CORRECT ✅

This was the specific thing the rebase put at risk (`tests/mutation_harness.py:735-736`), so it was
verified first. The hand resolution preserves **both** parents' behaviour: master's
`*pytest_targets(target)` / `str | Sequence[str]`, and the branch's per-run token across the plugin
filename, the outcomes filename and `-p <plugin.stem>`.

```
file target:          1 passed, rc=0
node-ID sequence:     2 passed, rc=0
unimportable plugin:  "mutation never applied ... FALSE 'CAUGHT'"   <- the fail-closed guard still fires

concurrent drivers:
  A: plugin=_mutation_plugin_34968_0  outcome=_mutation_outcomes_34968_0.json  rc=0
  B: plugin=_mutation_plugin_29320_0  outcome=_mutation_outcomes_29320_0.json  rc=0
```

Both drivers recorded complete call phases with no cross-corruption, and no mutation files were left
behind. Concurrency was the whole point of the per-run token, and it holds.

---

### Blocking finding 1 — nested junctions escape BOTH the deletion and the copy scope

**`scripts/sync_installed_skills.py:500`, `:803-808` — fail-open, data loss.**

`dst.rglob("*")` follows a junction that sits *below* an owned bundle. The resulting **relative** path
still passes the top-level `scope` check, and only afterwards does `unlink()` / `copy2()` follow the
junction to a destination outside the plugin.

Reproduced against the public CLI with a redirected plugin root — deletion side:

```
initial=0  check=1  publish=0
check_extra=powerbi-ai-readiness/nested-junction/keep.txt
outside_before=True   outside_after=False        <- a file outside the plugin was DELETED
publish_status=updated
```

and the copy side independently overwrote an external sentinel:

```
publish=0  status=updated
sentinel_before=True  sentinel_after=False       <- a file outside the plugin was OVERWRITTEN
outside_files=2
```

What already works, and is why this reads as closed until you test one level deeper: top-level foreign
bundles, prefix-named bundles, case aliases and **marker-level** junctions all survived or were
refused correctly. The guard is real — **the boundary simply moved one directory deeper** rather than
being removed. That is the classic "the fix moved the failure boundary" shape, so please fix the
containment *class* (resolve and re-check the real target before any `unlink`/`copy2`), not the one
depth that was probed.

### Blocking finding 2 — an OFFLINE recorded default is certified as verified

**`scripts/sync_installed_skills.py:337-346`, `:391-400`; `scripts/preflight.ps1:357-364` — fail-open,
false certification.**

After an online run recorded `master`, the reviewer changed the remote's actual HEAD to `main` with
different content and took the clone offline:

```
offline_exit=0
status=in_sync
default_proof=recorded
default_verified=True
installed_has_old=True        <- stale content still installed
installed_has_new=False       <- the real new content never arrived
remote_actual_head=refs/heads/main
preflight_merged_ok=True      <- preflight certifies the OLD content as merged
```

So the prior stale-`origin/HEAD` finding is **not closed on the offline-recorded path**. A recorded
default is evidence of what the remote said *last time*; it cannot certify what the remote says now.
"Cannot establish" needs to be its own state here — the same rule this repo applies to every other
gate — rather than collapsing into `verified`.

---

### Non-blocking (fail-closed → file as issues, do not block the merge)

- **`--fetch` is not honoured as "no network by default":** `resolve_publish_ref(fetch=False)` still
  executes `git ls-remote --symref origin HEAD`. Contract violation, but it fails safe.
- **CRLF claim confirmed:** with `core.autocrlf=true` the exact skill blob had **0** CRLFs while
  `git archive` produced **271**. Published bytes are machine/configuration-dependent.

### Prior findings verified CLOSED

- No origin ⇒ exit `5`, JSON exit `5`, **no filesystem change**, and the human output names `--from-worktree`.
- Local unmerged skill edit ⇒ exit `0`, human `SYNC: NOTE`, JSON `local_unmerged`.
- Stale redirected install ⇒ preflight exit `1`.
- No resolvable ref ⇒ preflight exit `1`, reported as **unverified** rather than stale.
- `--fetch --json` emits parseable JSON only.
- Foreign and prefix-named bundles survive ordinary publication.

### Mutation-campaign honesty

The campaign contains **30** skill-sync/source mutations, not the 20 claimed. The reviewer
independently killed the two load-bearing guards to confirm the harness has power here:

```
deletion-scope-removed:
  test_a_bundle_this_tool_never_installed_is_left_alone_when_it_is_retired
  AssertionError: retirement sweep must not reach...

never-ask-the-remote:
  test_a_renamed_remote_default_is_followed_without_being_asked
  AssertionError, intended guard test
```

It did **not** certify all 30 individually. More importantly: **neither blocking input above is
represented in the campaign**, so its all-caught claim cannot establish the closed safety surface it
appears to. Please add a mutation for each of the two blocking classes.

### Gates

| Gate | Result |
|---|---:|
| Preflight | 0 |
| Ruff format / check | 0 / 0 |
| Pylint — all four CI scopes, required interpreter | 0 |
| Privacy gate | 0 |
| Spec / datamodel / navigation / capabilities / conventions / AI | all 0 |
| Windows bundle | 0 — 306 passed, 3 skipped |
| Targeted safety controls | 0 — 4 passed |
| `git diff --check` | 0 |
| Full `pytest tests/` on base `dab3acb` | 1 — 4193 passed, **3** failed |
| Full `pytest tests/` on `8c50f0f` | 1 — 4303 passed, **same 3** failed |

The three failures are identical on both sides and are the known engine-drift pins
(`test_upstream_repro_pins.py`, issues 168 / 166 / 171). **No new full-suite failures.**

Exact-SHA GitHub checks at `8c50f0f`: `checks` success, `windows-bundle` success.

No machine-global plugin publication occurred during the review — every mutation ran against a
redirected scratch root, all subsequently removed.

Guust-Franssens:
Parking as draft to cut the in-flight batch down. **Not a cold-run blocker.**

State recovered this morning: the agent working this branch was killed by a host crash at ~06:00 with uncommitted work in the worktree. Salvaged as `ff3f9b5` after verifying `tests/test_sync_installed_skills.py` -> **111 passed in 416s**. Nothing was lost; nothing was unpushed.

That commit is the response to the two blocking fail-open findings from the last blind review (nested-junction escape from delete/copy scope; offline recorded default certified as verified). It has **not** been re-reviewed, so the BLOCKED verdict still stands until it is.

Standing concern when this resumes: **24 commits / +4,228 lines / 11 files** for one issue. Before another review round, the question is whether the remaining work should be split rather than extended.

Guust-Franssens:
## CI diagnosis (both failures), from the run logs

Two separate gates, in the order CI hits them. The first was masking the second — the workflow is `bash -e`, so it never reached pylint.

### 1. `uv run ruff format --check` — fixed

```
unformatted: File would be reformatted
    --> scripts/sync_installed_skills.py:1018:34
1 file would be reformatted, 247 files already formatted
##[error]Process completed with exit code 1.
```

⚠️ **Worth naming, because it makes the obvious fix wrong.** `pyproject.toml` pins only `"ruff>=0.6"`, so uv resolves **0.16.4** while a globally installed `ruff` on this machine was **0.14.5** — and they disagree. The global one reported a *different* file set (2 files vs CI's 1), so formatting with it can leave CI red while looking clean locally. Formatted with `uv run ruff`; `format --check` and `check` now both exit 0.

### 2. `uv run pylint scripts` — still open, and it is this PR's own

```
scripts/sync_installed_skills.py:1:0: C0302: Too many lines in module (1226/1200) (too-many-lines)
Your code has been rated at 10.00/10
##[error]Process completed with exit code 16.
```

Note the shape: **a perfect 10.00/10 and a non-zero exit.** `fail-under = 10` is satisfied while messages remain, so score and exit code disagree by design — gate on `$LASTEXITCODE`, never on the printed score.

Measured, this is not pre-existing:

| | lines |
|---|---|
| `origin/master:scripts/sync_installed_skills.py` | **131** |
| this branch | **1226** (cap `max-module-lines = 1200`) |

The PR takes the file from 131 to 1226 lines (+1144/-77), so it owns the 26-line breach.

### Recommended fix — split, do not suppress

A module-level `# pylint: disable=too-many-lines` would silence a signal about a file that grew 10x in one PR. There is a cohesive seam already: **lines 234–~500 are entirely git/ref resolution** — `_git`, `_rev_parse`, `has_origin`, `remote_default`, `remote_default_ref`, `_record_path`, `read_default_record`, `write_default_record`, `_default_alternatives`, `fetch_origin`, `fetch_exact_ref`, `_explicit_source`, `_advertised_default`, `resolve_publish_ref`, `export_ref_tree`. Extracting those mirrors the `path_flavour.py` precedent in #463.

⚠️ **A new script must be declared in two inventories or three tests fail** (this bit #463 today): add a row to `scripts/README.md` *and* to `docs/agent-capability-wiring.md`'s script capability inventory. The third failure is a knock-on — `check_agent_capabilities.py` exits 1 while the registry is unsatisfied, so its `--help` cp1252 test fails too.

Guust-Franssens:
Closing this implementation branch to stop the review spiral; the underlying #410 work remains open. The branch reached six substantive review rounds and 26 commits, grew sync_installed_skills.py from 131 to 1,226 lines, and still has a branch-owned pylint failure. The remote branch and review evidence remain available. A successor should start from current master and implement only the merged-ref publishing invariant, with optional proof machinery split out rather than extending this PR.

Issue #426, opened 2026-09-01
Gate the stacked-non-additive-measure shape (#424's defect class) - and it needs the model, not PBIR alone


Guust-Franssens:
## Proposal

A static check for **stacked-chart-of-a-non-additive-measure**: `visualType` in the stacking family
(`columnChart`, `barChart`, `stackedAreaChart`) **with a non-empty `Series` well**, where the `Y`
measure is not additive. That is the shape behind #424 — five airlines at 95/92/88/97/90 %
availability stacking into one ~462 % bar — and today **nothing** flags it: the engine reports
`tier: rebuilt, status: rebuilt`, an empty `reason` and a zero-item `remediation_worklist`, because a
stacked column is structurally valid PBIR, and `powerbi-report-author validate` returns 0 errors.

`powerbi-report-gotchas` §8 already states the rule in prose — *"`columnChart` + `Series` STACKS.
Never use it for a ranking/index measure"* — and it has now been missed twice in the field. This is
the "convert prose to a gate" case.

## ⚠️ Correcting the premise in the #424 brief: it is NOT decidable from PBIR alone

The brief that prompted this said the defect is "decidable from PBIR alone — no reference image
needed". Half of that is right and the half that is wrong determines where the check can live.

| what you need | where it lives | PBIR-only? |
|---|---|---|
| "this chart stacks" | `visual.visualType` | ✅ yes |
| "it has a legend" | `visual.query.queryState.Series.projections` | ✅ yes |
| "the Y measure is a **column** aggregated `Sum`/`Average`" | the projection's `Aggregation.Function` (Sum=0, Avg=1) and its `nativeQueryRef` (`Sum of Sales`) | ✅ yes |
| "the Y measure is a **model measure** and it is a ratio/rank/average" | **`formatString` / the DAX in the sibling `.SemanticModel`** | ❌ **no** |

The #424 case is precisely the fourth row: `Y = 'Selected Measure'`, a `SWITCH` dispatcher over 15
KPIs. PBIR carries the *name* and nothing else. A name heuristic (`%`, `Pct`, `Ratio`, `Rank`,
`Avg`…) is available but is a guess, and a `SWITCH` dispatcher can be additive in some branches and
not others.

**So the check is a bundle-level check, not a `.Report`-level one** — it needs
`<bundle>/pbip/<wb>/<Model>.SemanticModel` beside the report. That is available in every bundle we
[...]

Issue #424, opened 2026-09-01
Airborne Services: columnChart shipped where lineChart is correct - stacks percentage ratios into a ~462% bar


Guust-Franssens:
Reported by **Hemang**, who verified it three independent ways. High confidence.

## The defect

**Location:** `ia-airborne-services-dashboard`, page **"Detail"**, visuals `v-page-Detail8fe63b4fcec` and its sibling `v-page-Detail8fe89412529`.

A `columnChart` was emitted where a `lineChart` is correct.

### Verified three ways

**1. Byte-level cross-comparison.** The *same visual ID* exists in a sibling workbook from the same engine template and page:

```
ia-airborne-services-dashboard      v-page-Detail8fe63b4fcec  ->  "visualType": "columnChart"
acmu-cro-tail-monitoring-dashboard  v-page-Detail8fe63b4fcec  ->  "visualType": "lineChart"
```

Same template, same page, same visual id, different type — so this is not a judgement call about the source, it is an inconsistency inside the engine's own output.

**2. Binding structure** (read from `visual.json`): Category = `DATES` (a genuine date field), Series = `AIRLINE_CODE`/`TECHNOLOGY_SET`, Y = `'Selected Measure'` — a `SWITCH` dispatcher over 15 KPIs, mostly percentage availability ratios.

⚠️ This is the documented bug class, and the consequence is not cosmetic: a `columnChart` **with a Series field stacks**. Stacking percentage ratios is meaningless — five airlines at 95/92/88/97/90 % sum into a single ~462 % bar.

**3. The Tableau reference image** (`reference/images/Detail__bb938602.png`) shows two multi-**line** charts, "SLA Availability by Technology" and "SLA Availability by Airline". The source really was lines.

## Fix scope

Small and report-layer only: **2 `visual.json` edits** — `visualType` `columnChart` → `lineChart`, plus swapping the type-specific formatting properties. No model change.

## ⚠️ Why this belongs on the toolkit's list, not just the customer's

His own framing, and it is the strongest argument in the report:

> This was only catchable **because a Tableau reference image for the "Detail" page happens to exist.** The same class of bug on "Availability Summary by Tail" or "Daily Monitoring" would be completely invisible — there is nothing to compare against.

Those two pages are precisely the capture blind spots in #423. So this issue is the concrete payoff case for #423 and #421: a real, shipped, stacking-percentages defect that reference completeness made falsifiable.

[...]

Issue #184, opened 2026-08-15
The --tamper audit trail is git-ignored: **/_build/ means the declarations ledger and replay scripts never leave the machine that made them


Guust-Franssens:
The audit trail that `check_migration_progress.py --tamper` demands, and that the shared conventions mandate, is **git-ignored** — so it cannot survive the machine that produced it.

Found while designing multi-machine parallelisation; it also affects any handover, any rebuild, and any second person picking up an estate.

## The contradiction

`AGENTS.md` shared conventions (synced into all four personas):

> **Declare generated edits.** TMDL/PBIR/`.pbip`: file/change/why + **replay script** + **hash record**.

`tableau-migrator`, before sign-off:

> `check_migration_progress.py --bundle <b> --tamper`; drift blocks

And #166/#180 just made the builder personas able to produce those declarations via `scripts/declare_generated_edit.py`, which writes to **`<bundle>/_build/generated-edit-declarations.json`**.

But:

```
$ git check-ignore -v "migrations/workbooks/foo/_build/fix_x.py"
.gitignore:64:**/_build/	migrations/workbooks/foo/_build/fix_x.py

$ git ls-files | grep -c generated-edit-declarations
0
```

**`**/_build/` (`.gitignore:64`) ignores the whole directory.** So both halves of the mandated audit trail — the re-runnable replay scripts *and* the hash ledger — exist only on the machine that ran the build.

⚠️ Note a false reassurance: `git ls-files | grep '_build/'` returns 4 files, which looks like the rule is not absolute. Those are `examples/quadruple-axis-charts/**report_build**/…` — a different directory that merely contains the substring. **No file in any real `_build/` directory is tracked.**

## What breaks

| scenario | consequence |
|---|---|
| Hand a bundle to a colleague / second machine | `--tamper` sees 82 undeclared artifacts and exits 1. The declarations existed; they did not travel |
| Re-clone, or rebuild on a fresh checkout | Same — the audit trail is gone and cannot be reconstructed |
| Deliver the estate to the customer | We ship modified generated artifacts with no record of what changed them, which is the exact thing the gate exists to prevent |
| Multi-machine parallel migration | No machine can verify another's work; the contract has nothing durable to point at |

[...]

PR #427, opened 2026-09-01
Settle #424 before filing: the columnChart is an engine defect, and the reported clincher is not one

Guust-Franssens:
## What this is

**Not the fix in #424's "Fix scope" — that could not be made here.** The two `visual.json` files it
names do not exist on this machine (`ia-airborne-services-dashboard` and
`acmu-cro-tail-monitoring-dashboard` are absent from an exhaustive recursive scan of the user
profile; this repo commits only three README files under `migrations/`). What this PR delivers is the
**answer to the question #424 said to settle before filing upstream**, plus the durable artifacts it
produced — and, after round-1 review, a fixture set that also bounds how the fix may be written.

Full reasoning and residual unknowns: the
[verdict comment on #424](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/issues/424#issuecomment-5498339588)
and the [round-1 response](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/pull/427#issuecomment-5500409468).

## Verdict: engine-side — upstream [`tableau-fabric-skills#184`](https://github.com/Yarbrdab000/tableau-fabric-skills/issues/184)

`<mark class='Automatic'/>` over a **discrete** date pill is classified `VT_COLUMN`; Tableau draws a
**line**. With a colour dimension that becomes `columnChart` — Power BI's *stacked* column — so it
sums the series. `twb_to_pbir.py:2366` `_has_continuous_date` gates on **continuity**; Tableau's
[product doc](https://help.tableau.com/current/pro/desktop/en-us/viewparts_marks_marktypes.htm) gates
on **date-ness** (*"If the dimension is a date dimension, the Line mark is used instead"*). The engine
is silent: `viz_fidelity` = `tier: rebuilt`, empty `reason`, zero worklist items.

### ❌ …and #424's stated strongest evidence does not hold

The shared visual id across two workbooks is **not** an engine inconsistency. A PBIR visual name is
`_sanitize(f"v-{page_name}-{i}-{ws['name']}")` — a 16-char prefix plus an 8-char md5 — so it is a
deterministic function of **(dashboard name, zone index, worksheet name) only**. All eight fixtures
here emit the identical `v-page-Detail8fe20137fae` while disagreeing on `visualType`.

## The fixture set bounds the fix, not just the defect

Eight workbooks generated from one template, each differing by exactly one variable. Five candidate
remedies injected into 2.339.0 at runtime; **each wrong one is separated by exactly one fixture**, and
a three-fixture A/B/C set would have passed all five:

| wrong remedy | caught only by |
|---|---|
| `year-only` | **E** discrete exact date (`MDY`) |
| `datatype=='date'` | **F** a `datetime` column |
| `base-columns-only` | **G** a date-valued calculated field |
| `mark-agnostic` | **D** an explicit `Bar` mark |
| `any-discrete` | **H** a non-date string dimension |

`tests/test_issue_424_chart_type_pin.py` splits these into `DEFECT_PINS` (A/E/F/G — must flip to
`lineChart` when fixed) and **`PERMANENT_INVARIANTS`** (B/C/D/H — ⚠️ *do not retire with the rest of
the pin*). D and H are what stand between "fixed" and "rewrote every bar chart into a line chart":
an explicit `Bar` mark stacks in Tableau too, so its `columnChart` is faithful.

### ⚠️ The INPUTS are pinned too — the discrimination claim is otherwise vacuous

D/E/F/G/H all currently emit `columnChart`, the same as A, so each one's distinguishing *input* can
be collapsed onto A's while every emitted-type assertion stays green. Measured before fixing:
deleting D's `<mark class='Bar'/>` **passed all nine tests**, including the permanent-invariant test
whose only purpose is guarding that mark.

Each fixture's mark class and its **fully-resolved** Columns pill are now pinned by parsing the
`.twb` **source** against a hand-written specification — parametrized per fixture — plus a
distinctness assertion and the shared stacking ingredients. That half uses **no engine**, so unlike
everything else in this module it is live in CI (measured with the plugin absent: 18 passed,
8 skipped — previously CI verified *nothing* about this fixture set; the wider gap is
[#435](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/issues/435)).

⚠️ **Pills are RESOLVED, not name-matched.** Round-3 review measured that flipping `AIRLINE_CODE`
from `role='dimension'` to `role='measure'` leaves the token `none:AIRLINE_CODE:nk` byte-identical,
removes the `Series` well, and stops the fixture reproducing a stacked percentage — with all 18
offline tests green. Colour and value pills now follow their `column-instance` to the declaring
`column` and pin `(field, derivation, role, datatype, is_calc)`; the dashboard zone's reference to
the worksheet is pinned too.

## Contents

| file | why |
|---|---|
| `fixtures/upstream-repros/issue-424-automatic-mark-discrete-date/` | 8 one-variable `.twb` + README carrying the remedy matrix. |
| `tests/test_issue_424_chart_type_pin.py` | Defect pins + permanent invariants; an anchor asserting the exact `(queryRef, entity, property, aggregation)` binding; and an engine-free **input-shape pin** so the fixtures cannot silently collapse onto each other. |
| `.github/skills/powerbi-report-gotchas/SKILL.md` | §8: product-doc citation, engine code site, upstream number, the id falsification, and the over-broad-fix hazard. §3: the measured `PBIR_FORMATTING_PROP_UNKNOWN` verdict for a type flip. |
| `pyproject.toml` | pylint `init-hook` so `scripts/` resolves statically — removes a **false** `E0401` that made pylint exit non-zero while printing a healthy score. |
| `fixtures/upstream-repros/README.md`, `docs/INDEX.md` | Index the new fixture (nav gate). |

Filed rather than fixed here: **#426** (a static gate for the stacked-non-additive shape — premise
corrected, it needs the sibling `.SemanticModel` so it is *not* decidable from PBIR alone),
**#434** (pylint-vs-`tests/`, a documented decision with the cost measured), and **#435** (CI runs
none of the 15 engine-dependent tests — they skip silently, and a skip is not a pass).

## Why the pin lives in its own module

`tests/test_upstream_repro_pins.py` asserts `PINNED_ENGINE_VERSION = "2.260.0"` before anything else,
so all three of its engine-backed tests already fail locally against the installed **2.339.0**
(pre-existing; they skip in CI, which has no engine plugin). A pin added there would have shipped
unrun. This module reports the observed version in its failure text instead of asserting one.

## Verification

- `pytest tests/test_issue_424_chart_type_pin.py` → **26 passed**; **18 passed / 8 skipped** with the
  engine plugin simulated absent (the CI condition).
- **Mutation-tested by mutating the FIXTURE, never the test**, each run against its intended anchor
  **alone** (no `-x`). Emitted-type and binding assertions: 12/12 caught. Input-shape pin: 7/7
  correctly attributed — for all five fixtures in the round-2 finding the input case fails, another
  fixture's case still passes, and **the emitted-type matrix still passes**, which is what proves the
  input pin is doing the work. Semantic pin: **7/7**, flipping roles/datatypes/derivations rather
  than names, **with the engine simulated absent throughout** so no engine-dependent test could take
  credit; each mutation leaves its *sibling* pin green, so axis and common-facts catch their own
  classes rather than masking one another.
- `ruff format --check` / `ruff check` over `conftest.py scripts tests .github/skills` → clean.
  `pylint` → **exit 0** on all six targets (four CI invocations + both engine-backed test files).
  `check_navigation_index.py`, `sync_agent_conventions.py --check` → pass.

## ⚠️ What is NOT verified

- **No Power BI Desktop render, of anything.** Free RAM 4–5 GB of 31.7 GB; zero `PBIDesktop`
  processes for the whole session. The `lineChart` claim is what Tableau's docs and the engine say,
  not what I watched Desktop draw.
- **Which input shape the customer's workbook actually has.** An explicit `<mark class='Bar'/>` makes
  the same `columnChart` **faithful**. Read the worksheet's mark class before editing anything.
- **Date bin / date parameter** — no real serialization was available; a synthetic date-typed bin
  crossed to `lineChart` under a datatype-keyed predicate, so that boundary is untested, not settled.
- **`lineStyles.strokeColor` renders theme-blue** — ⚠️ inferred, attributed to the engine's own
  `_constant_mark_color_objects` docstring. The *validate* half is measured: `defaultColor` alone,
  `strokeColor` alone and both together all exit 0, so `validate` is blind to the omission.
- **`tests/test_upstream_repro_pins.py` is red locally** (3/4) for the pre-existing engine-version
  reason above. Not touched, not caused here, not fixed here.

refs #424


Guust-Franssens:
## Independent confirmation, from a live Desktop pass

Hemang Patel (SES) ran a hands-on Power BI Desktop verification over `IA_Airborne_Services_Dashboard` and reached this defect independently, without seeing this PR:

> Both charts were emitted as `columnChart` (stacks series) instead of `lineChart` — the same "date-on-Columns resolves to line, not bar" mistranslation already documented as a known gotcha.

⚠️ **This is worth recording because it changes the grade of the evidence, not just the count.** Everything behind this PR so far has been static: a reading of `twb_to_pbir.py:2366`, plus a `validate` run that returns zero errors either way. Static analysis can establish that the emitter's gate is wrong; it cannot establish that Desktop actually draws the wrong thing. His pass did — on a different workbook, in a different estate, arrived at from the rendered output rather than from the source.

So the two halves are now genuinely complementary:

| | what it establishes |
|---|---|
| this PR's analysis | **why** it happens — `_has_continuous_date` gates on continuity (`endswith("-Trunc")`) where Tableau's rule gates on date-ness |
| his Desktop pass | **that** it happens, visibly, in delivered work |

It also confirms the frequency argument. This was filed off a single ~462 % stacked bar; his report finds the same mistranslation on two more charts in an unrelated workbook. Not a one-off.

⚠️ **One caveat this does not lift.** Per the standing note on this issue, an explicit `<mark class='Bar'/>` makes `columnChart` **faithful** — so the fix must still key on the mark class and not simply rewrite every date-on-Columns chart to a line. His finding raises confidence in the defect; it does not widen the fix.

Guust-Franssens:
## Round 1 — all three addressed, and two corrections to the review

Thanks — finding 1 was the one that mattered and it was right in a way I had not seen. Reproduced, then generalised: it turned out to be worse than reported.

---

### 1. The pin could not distinguish a correct fix from **five** wrong ones, not two

I rebuilt your experiment with a wider candidate set — five remedies injected into the canonical 2.339.0 classifier at runtime (monkeypatching `_has_continuous_date` / `_visual_type`; the engine tree untouched) over an extended fixture set. `column` = `columnChart`, `LINE` = `lineChart`:

```
remedy                    A        B        C        D        E        F        G        H
as-shipped 2.339.0        column   LINE     LINE     column   column   column   column   column
CORRECT: date-aware       LINE     LINE     LINE     column   LINE     LINE     LINE     column
wrong: year-only          LINE     LINE     LINE     column   column   LINE     LINE     column
wrong: datatype=='date'   LINE     LINE     LINE     column   LINE     column   LINE     column
wrong: base-columns-only  LINE     LINE     LINE     column   LINE     LINE     column   column
wrong: mark-agnostic      LINE     LINE     LINE     LINE     LINE     LINE     LINE     column
wrong: any-discrete       LINE     LINE     LINE     column   LINE     LINE     LINE     LINE
```

Each wrong remedy is separated by **exactly one** fixture, so none of the five new ones is redundant:

| wrong remedy | caught only by |
|---|---|
| `year-only` | **E** `e-discrete-exact-date` (`MDY`) |
| `datatype=='date'` | **F** `f-datetime-date-part` |
| `base-columns-only` | **G** `g-date-valued-calc` |
| `mark-agnostic` | **D** `d-explicit-bar-mark` |
| `any-discrete` | **H** `h-non-date-dimension` |

**A/B/C alone would have passed all five** — your finding, confirmed and extended.

All eight are generated from one template, so "identical except one variable" is mechanical rather than eyeballed. `tests/test_issue_424_chart_type_pin.py` now splits them into `DEFECT_PINS` (A/E/F/G — flip to `lineChart` when fixed) and `PERMANENT_INVARIANTS` (B/C/**D**/**H**), with a `⚠️ DO NOT RETIRE` docstring on the invariant test and a matrix test whose failure message tells the reader which case they are in.

Your suggested `datetime` and calculated-field positives are in as F and G; the negative control is H (string dimension). ⚠️ The **date bin / date parameter** boundary you flagged is still open — I had no real serialization of either, so I documented it as untested rather than closing it.

### 2. The anchor now pins the exact binding

`EXPECTED_A_BINDING` asserts the full triple as `(queryRef, entity, property, aggregation)`: `Date.Year` on Category, `Sum(SLA.AVAILABILITY_PCT)` on Y with `Function == 0`, `SLA.AIRLINE_CODE` on Series.

Proven to bite two ways. Deliberately: a `Sum → Avg` fixture mutation leaves `visualType == "columnChart"` passing and fails on `assert observed == EXPECTED_A_BINDING`. Accidentally but more directly: the first run of the rewritten helper drilled to the wrong JSON level and produced `observed: {}` — the test failed with `{'Category': None} != [('Date.Year', ...)]`, which is precisely the shape you injected.

### 3. pylint — fixed at the root, and `source-roots` is a trap

`init-hook` in `[tool.pylint.main]`. ⚠️ I tried `source-roots` first and **measured it worse**:

| lever | `tests/*` | `.github/skills/pbip-model-refresh/scripts` |
|---|---|---|
| before | exit 2 — false `E0401` | exit 0 |
| `source-roots = ["scripts", "."]` | exit 2 — **still** | **exit 2 — newly broken** |
| `init-hook` | **exit 0** | exit 0 |

`source-roots` only decides package *names*, never the import search path — and adding `.` re-labelled the skill scripts into a deep `.github.skills…` package, breaking their sibling-by-bare-name imports. `init-hook` leaves **all six** targets at exit 0, including the pre-existing sibling `tests/test_upstream_repro_pins.py`.

---

## Two corrections to the review

**⚠️ "CI's pylint excludes `tests/`" is a documented decision, not a gap.** [`checks.yml:58-60`](../blob/master/.github/workflows/checks.yml): *"Tests are not held to 10.00/10; the toolkit code is."* And `tests/` **is** linted — by ruff, which names it explicitly. Filed as **#434** framed as a decision with the cost measured (`pylint tests` → exit 30, 1,265 messages, 70/92 files), noting that `W0621`/`W0212`/`C0415` are pytest-idiomatic and would need disabling rather than fixing, so nobody inherits "just turn it on".

**⚠️ On upstream #184** — that issue is the one I filed in round 0, so there was no duplicate risk; I commented as asked. But its current state is narrower than "triaged into an *unfixed for no good reason* bucket": it carries **no labels**, and the single maintainer comment says explicitly *"No fix, no version. I have not run your fixtures, and I have not decided between widening the predicate and warning."* That is why the matrix above went to him — it is the evidence that bears on exactly that open decision, and it also closes the two limits he named (his corpus match was workbook-level, and he had not checked which worksheets carry a colour encoding; the comment carries a tested per-worksheet script for both). He is on `2.345.0`, I am on the installed canonical `2.339.0` — flagged there rather than quietly compared.

---

## Verification

* `pytest tests/test_issue_424_chart_type_pin.py` → **9 passed**.
* **Mutation-tested by mutating the FIXTURE, never the test, each against its intended anchor ALONE (no `-x`): 12/12 caught.** Negative control run: under a mutation, an unrelated node still passes, so the failures are scoped rather than import errors.
* ⚠️ One mutation label in my own harness was wrong and is worth stating: "drop the colour encoding" does **not** keep `columnChart` — the engine emits `clusteredColumnChart` without a Series, so that mutation is caught by the *type* assertion, not the binding one. The `Sum → Avg` mutation is the one that proves the binding half.
* `ruff format --check` / `ruff check` over `conftest.py scripts tests .github/skills` → clean. `pylint` → exit 0 on all six targets. `check_navigation_index.py`, `sync_agent_conventions.py --check` → pass.

## Still not verified

* **No Power BI Desktop render.** Free RAM 4–5 GB of 31.7; zero `PBIDesktop` processes all session (the two PIDs named in my brief do not exist).
* The `lineStyles.strokeColor` claim you flagged is now **split**: the validate half is measured (`defaultColor` alone, `strokeColor` alone and both all exit **0** — so `validate` is blind to the omission), and the "renders theme-blue" half is marked ⚠️ inferred and attributed to the engine's own `_constant_mark_color_objects` docstring rather than asserted.
* Date bin / date parameter boundary — open, documented.
* The remedy-matrix experiment monkeypatches **private** engine internals, so it is deliberately **not** committed as a test; its result table is recorded in the fixture README with enough detail to re-derive. Happy to commit the script if you'd rather have it than the table.

Guust-Franssens:
## Round 2 — reproduced your finding first, then fixed it. Pushed as `cdd365a`.

You were right, and I confirmed it before touching anything: deleting D's `<mark class='Bar'/>` passed **all nine** tests, including the permanent-invariant test whose only purpose is guarding that mark.

Agreed on the classification too — this is a different animal from the other five vacuity modes. Those were tests passing for the wrong *reason*. This is a test whose **premise** can evaporate while it stays green, and the failure is silent in the worst possible direction: the module goes on advertising five-way remedy discrimination having lost the controls that provide it.

### The fix

Each fixture's `(mark_class, axis_field, derivation, datatype, is_calc)` is pinned by **parsing the `.twb` source** — `xml.etree`, resolving the Columns pill back through its `column-instance` to the declaring `column` — against a hand-written specification. Not the filename, not `THE_ONE_DIFFERENCE`, not the engine's IR. Parametrized, so a drifted fixture fails **its own case**.

Two assertions beyond the per-fixture pin:

* **distinctness** — eight fixtures, eight distinct input shapes. This is the property the matrix's claim actually rests on, and it survives someone updating an individual expectation carelessly.
* **shared ingredients** — `Detail` dashboard, the `AIRLINE_CODE` colour dimension, the `Sum(AVAILABILITY_PCT)` value pill. An axis pill can be perfectly correct while the fixture has quietly stopped demonstrating a *stacked ratio*.

### Mutation-proved the way you found it

Collapsing each fixture's distinguishing input onto A's, each node run **alone**:

| collapse onto A | its own input case | another fixture's | distinctness | **emitted-type matrix** |
|---|---|---|---|---|
| D `Bar` → `Automatic` | **FAIL** | pass | FAIL | **pass** |
| E `MDY` → `Year` | **FAIL** | pass | FAIL | **pass** |
| F `datetime` → `date` | **FAIL** | pass | FAIL | **pass** |
| G calc field → base column | **FAIL** | pass | FAIL | **pass** |
| H string dim → date part | **FAIL** | pass | FAIL | **pass** |

The matrix column is the proof you asked for: for all five it stays **green**, so the input pin is demonstrably what catches them rather than being credited for a failure something else produced. B and C are caught by both — stated plainly rather than counted as a win, because their emitted type genuinely differs from A's, so they were never part of the finding.

### A second gap this closed, which nobody had named

The parser deliberately uses no engine. That turned out to matter more than the R0914 tidiness it also bought: **every other test in this module is `@requires_engine` and therefore SKIPS in CI**, which has no conversion plugin — so CI was verifying *nothing whatsoever* about this fixture set. Measured with the plugin simulated absent:

```
18 passed, 8 skipped
```

So the anti-vacuity half is now the one part of the file that is actually live in CI, which is a reasonable place for it to be: it is the half that guards a claim rather than records engine behaviour.

### Housekeeping

`pylint` exited **8** on `R0914` (18/15 locals) in the new parser while printing **9.94/10** — sixth time today the score and the exit code disagreed. Split into `_axis_shape` / `_common_facts`; back to exit 0. All six pylint targets exit 0, ruff clean, `check_navigation_index` and `sync_agent_conventions --check` pass, module at **26 passed**.

Thanks for conceding both corrections explicitly — and for the #184 one in particular. I'd rather have the maintainer's actual position on record (*"no fix, no version"*, widen-vs-warn undecided) than a tidier summary of it, because that position is exactly why the remedy matrix was worth sending him.

Nothing new on the unverifiable list: still no Desktop render, no customer artifact, and no date-bin or date-parameter serialization.

Guust-Franssens:
## Round 3 — both handled. Fix pushed as `8487a02`; the CI gap filed as #435.

### 1. Reproduced, then fixed — and you named the shape exactly right

This is the same vacuity one level in, and I'd missed it for the same reason twice: round 2 moved the module from *assert the output* to *assert the input*, and I stopped there without asking whether the input assertion was checking a **token** or a **meaning**.

Confirmed before touching anything, in CI conditions:

```
AIRLINE_CODE  role='dimension' -> role='measure'
engine absent -> 18 passed, 8 skipped        # nothing caught it
```

Both shelf pills now resolve through their `column-instance` to the declaring `column`, exactly as the axis pill already did:

| shelf | pinned as |
|---|---|
| Columns (axis) | per fixture + **`role=dimension`** |
| Colour | `AIRLINE_CODE`, `derivation=None`, **`role=dimension`**, `datatype=string` |
| Rows (value) | `AVAILABILITY_PCT`, **`derivation=Sum`**, **`role=measure`**, `datatype=real` |

Plus your two extras: the dashboard **zone actually references the worksheet**, and there is **exactly one** colour encoding — a second would silently change what the Series well carries.

**Mutation-proved by flipping semantics, never a name, with the engine simulated absent throughout** so no engine-dependent test could be credited for the catch:

| mutation | intended node | another fixture | **sibling pin** |
|---|---|---|---|
| colour `role` dimension → measure | **FAIL** | pass | pass |
| colour `derivation` None → Sum | **FAIL** | pass | pass |
| value `role` measure → dimension | **FAIL** | pass | pass |
| value `datatype` real → string | **FAIL** | pass | pass |
| value `derivation` Sum → Avg | **FAIL** | pass | pass |
| dashboard zone points elsewhere | **FAIL** | pass | pass |
| **axis** `role` dimension → measure | **FAIL** | pass | pass |

The sibling column is the attribution you asked for: a common-facts mutation leaves the **axis** pin green, and the axis mutation leaves **common-facts** green — each assertion catches its own class rather than one masking the other. Your exact reproduction now gives `1 failed, 17 passed, 8 skipped` offline, failing on `test_every_fixture_keeps_the_shared_stacking_ingredients[...a-discrete-date-part]`.

### 2. CI engine coverage — filed as #435, not fixed here

Agreed it is out of scope. Filed with your framing intact, including the load-bearing clause: **the job must FAIL if the plugin is unexpectedly absent** rather than converting 15 assertions into skips and reporting green. Suggested shape: an integration job on a *deliberately pinned* engine version (a red build then means our code moved, not the marketplace), plus a scheduled current-engine drift job where redness is the signal.

⚠️ **One correction, and it goes in your favour.** I tried to verify your 15 before repeating it, and my first measurement said **11** — which would have been wrong to publish. Cause: `T2P_SIMULATE_ENGINE_ABSENT_FOR_TESTS` is read by only **two** of the three modules; `test_dax_oracle_server.py` gates on `engine_source.engine_root()` directly, so the simulation cannot reach its 4 tests. Counting `@requires_engine` per module gives 8 + 4 + 3 = **15**, exactly your number. The issue records the caveat so the next person doesn't repeat my method.

⚠️ **One thing I could not resolve and have flagged as unverified in #435:** a full serial run under that simulation reported `1 failed, 3894 passed, 20 skipped`, and I did not identify the single failure. Since only those two modules read the variable it is unlikely to be engine-related, but I am not asserting that.

### State

Module **26 passed** (18 passed / 8 skipped with the engine absent). `pylint` exit 0 on all targets, ruff clean, navigation and conventions pass. Full serial suite re-running; last measured `3903 passed` with only the three documented engine-version pins failing.

Unverifiable list unchanged: no Desktop render, no customer artifact, no date-bin or date-parameter serialization.

Guust-Franssens:
## Round 4 — the open item, answered as **"artifact lost, and here is exactly why"**, with the reproduction attempt bounded

Rebased onto `58f83aa`; branch is now `0b39ad7`.

You listed three acceptable answers. I can give you the third **with a mechanical proof rather than a shrug**, plus most of the second.

---

### ⚠️ The node id was never captured — and it could not have been

Not carelessness in reading the output. The lost run used `-rs`, and pytest's `-r` flag **replaces** the default `-rfE` rather than adding to it, so the `FAILED <node id>` short-summary line was never printed. Verified by controlled experiment on a deliberately failing test:

```
pytest -q --no-header -rs        ->  "1 failed in 0.13s"          <- no FAILED line at all
pytest -q --no-header            ->  "FAILED _s5/test_probe_fail.py::test_deliberate_failure - AssertionError: probe"
```

The shell session holding that run's stdout is gone, `.pytest_cache/v/cache/lastfailed` has been overwritten several times since, and I ran with `-p no:cacheprovider` in the repeats. **So the node id cannot be reconstructed.** That is a dead end, and I would rather name it than guess at it.

### What the numbers do tell us, without the node id

| run | result | total collected |
|---|---|---|
| historical (lost) | `1 failed, 3894 passed, 20 skipped` | **3915** |
| your rerun @ `8487a02` | `3895 passed, 20 skipped` | **3915** |
| my repeats @ `0b39ad7` | `3895 passed, 20 skipped` | **3915** |

Identical collection totals, so the historical difference was **one test's outcome**, not a different test set — it was a genuine test failure, not a collection or import error. That narrows it usefully even though it does not name it.

### Reproduction attempt — bounded, and short of your 20

⚠️ **I did not reach 20, and I am not going to imply I did.** Each full offline run is **15–18 minutes** on this machine, so 20 sequential runs is **~6 hours**. Completed so far at `0b39ad7`, all clean, exit 0:

**5 consecutive full serial offline runs, all clean:**

| run | exit | result | wall |
|---:|---:|---|---|
| 1 | 0 | `3895 passed, 20 skipped` | 18 min |
| 2 | 0 | `3895 passed, 20 skipped` | 15 min |
| 3 | 0 | `3895 passed, 20 skipped` | 15 min |
| 4 | 0 | `3895 passed, 20 skipped` | 13 min |
| 5 | 0 | `3895 passed, 20 skipped` | 14 min |

I deliberately did **not** pad the count with `-n auto --dist loadfile` runs, which take ~4 minutes each: `conftest.py` deselects every `serial`/`timing` test under xdist, so those are a *different* condition and adding them to reach "20" would be exactly the kind of number-inflation this repo's review culture exists to catch. The historical run was serial; only serial runs count here.

### Your four intermittency sources, checked

| source | finding |
|---|---|
| **test-order dependence** | ⛔ ruled out — `pytest-randomly` is **not installed**, so collection order is deterministic. A deterministic order means a deterministic failure would reproduce; it does not. |
| **parallelism artefact** | ⛔ ruled out — the lost run was plain serial `pytest -q --no-header -rs`. No xdist, so `conftest.py`'s `--dist loadfile` guard (exit 4) is not implicated. |
| **machine state / contention** | Best hypothesis, **not confirmed.** The obvious suspect is `test_the_process_main_window_handle_is_a_z_order_answer_not_an_identity` — it creates real top-level windows, raises them to `HWND_TOPMOST`, sleeps 200 ms and asserts on a Z-order-derived answer, its own docstring calls the underlying API *"unstable"* (5 of 6 runs one way), and it carries **neither** `@pytest.mark.serial` **nor** `@pytest.mark.timing`, so it runs in every sweep. It is also one of two tests already observed failing under load in this session's round-1 parallel run. **Hammered both, 40 reps each, deliberately while a full suite was running: 0/40 and 0/40.** So: plausible mechanism, no evidence. I am not asserting it. |
| **time / date dependence** | Surveyed — `date.today()` / `datetime.now()` / `time.time()` appear in 4 test modules (14 occurrences in `test_check_migration_progress.py`). Not excluded, not implicated. |
| **whose run was it** | Mine, in this worktree. Sibling worktrees are separate directories with their own `REPO` root, so the shared on-disk scratch this module uses (`.pytest_cache/<run-root>`) cannot collide across them. |

### The bounded claim

> **Not reproducible in 5 consecutive full serial offline runs at `0b39ad7`, collection totals identical to the failing run, order deterministic, and the two previously-observed load flakes clean at 40/40 each. The node id is unrecoverable because `-rs` suppressed the failure summary — proven, not assumed.**

If you want the full 20 before merging, say so and I will run it — it is ~6 hours of wall clock and I would rather spend it than leave you with a number I did not earn.

---

Everything else from round 3 stands: the sibling-pin attribution, the 15 (`8 + 4 + 3`), and #435 with its load-bearing "must fail on unexpected absence" clause. Unverifiable list unchanged: no Desktop render, no customer artifact, no date-bin or date-parameter serialization.

Guust-Franssens:
## Round 4 follow-up — the reproduction campaign found something, and it is **not** a failure

Extending the bounded claim from 5 runs to **13**, and reporting an anomaly the campaign surfaced that I would rather hand over than leave in a scratch file.

### 13 consecutive full serial offline runs at `0b39ad7` — **zero failures**

| phase | runs | result |
|---|---:|---|
| 1 | 11 | exit 0, every run |
| 2 (instrument re-armed, see below) | 2 | exit 0, identical skip sets |

⚠️ **Run 10 of phase 1 was not identical to the others:**

```
runs 1-9, 11      3895 passed, 20 skipped
run 10            3894 passed, 21 skipped     <- one test that normally PASSES was SKIPPED
```

Exit 0, no failure — but the suite's **outcome composition is not deterministic**. That is the mechanism class the historical failure most likely belongs to: a guard that can decide "skip" from live machine state can also lose that race and decide something else. It is evidence *for* the machine-state hypothesis without naming the test, and it is the first hard evidence of intermittency in this suite that is not merely remembered.

### I could not identify it, and the reason is the same flag defect

Phase 1 ran `-q --no-header` (default `-rfE`) — which itemises **failures** but not **skips**. So run 10's extra skip was counted and not named, exactly as the historical failure was named-and-not-counted under `-rs`. Two halves of the same mistake.

**Instrument re-armed:** phase 2 runs `-rfsE` — failures *and* skips itemised — and records the full skip set per run, so a recurrence is identified automatically rather than merely counted. Baseline captured, 20 skips, all stable environmental guards:

```
3x  pbip-model-refresh/test_persist_atomicity.py   no real cache.abf (PBIP_REFRESH_REAL_ABF unset)
2x  test_check_path_ceiling.py                     Windows-only behaviour
8x  test_issue_424_chart_type_pin.py               deterministic tier not installed
4x  test_live_multisource.py                       live fixture not generated
3x  test_upstream_repro_pins.py                    deterministic tier not installed
```

None of those 20 is racy, so the 21st came from a test that normally passes.

### Targeted hunting — four campaigns, all stable

| target | reps | result |
|---|---:|---|
| `test_the_process_main_window_handle_is_a_z_order_answer_not_an_identity` (creates real windows, raises `HWND_TOPMOST`, asserts on Z-order; own docstring calls the API *"unstable"*) | 40 | **0 failures** |
| `test_biff8_gate_reaches_the_cli_exit_code` (the other test observed failing under load this session) | 40 | **0 failures** |
| the five modules with runtime `pytest.skip()` on live machine state — drive-letter availability for `subst`, NTFS junctions, 8.3 name generation, admin share, remote schema fetch | 10 | `239 passed, 2 skipped` **every rep**, no skip-set variance |
| whole `.github/skills` tree + `test_credential_gate.py` + `test_probe_live_source.py` | 6 | `415 passed, 3 skipped` **every rep**, no skip-set variance |

The first two were hammered deliberately **while a full suite was running**, since contention is part of the condition being tested.

### Where that leaves the open item

> **Not reproducible in 13 consecutive full serial offline runs at `0b39ad7`. The node id is unrecoverable — `-rs` replaced pytest's default `-rfE` and suppressed the `FAILED` line entirely (proven by controlled experiment), the session holding that stdout is gone, and `lastfailed` has been overwritten since. Collection totals are identical across the failing and passing runs (3915), so it was one test's outcome and not a different test set. Order is deterministic (`pytest-randomly` absent) and the run was serial, so neither ordering nor xdist is implicated. One genuine intermittency **was** observed — a skip/pass variance in 1 of 13 runs — but it manifested as a skip, not a failure, and I could not identify it either.**

I have filed the skip variance as **#436** rather than leaving it in a PR comment, with the `-rfsE` instrument and the baseline skip set, so whoever sees it next can name it in one run.

⚠️ **What I am not claiming:** that the skip variance and the historical failure are the same test. They are the same *class* — machine-state-dependent outcome — and nothing more than that is established.

Rebased onto `58f83aa`; force-pushed as `0b39ad7`. Phase 2 continues in the background; if it catches a recurrence I will post the node id.

Guust-Franssens:
**Campaign closed at 16 runs.** Final tally at `0b39ad7`, full serial offline condition:

| phase | runs | failures | skip-set variance |
|---|---:|---:|---:|
| 1 (`-rfE`, skips counted not named) | 11 | **0** | 1 — run 10, `3894 passed, 21 skipped` |
| 2 (`-rfsE`, skips itemised) | 5 | **0** | **0** — skip set byte-identical every run |
| | **16** | **0** | 1 |

So the skip variance did **not** recur in the 5 runs where the instrument could have named it. At ~1 in 16 (~6%) that is unsurprising and not evidence of absence; the instrument and the baseline are recorded in #436 so the next person to see it gets the node id in one run rather than in a campaign.

**Final answer to the open item, unchanged in substance and now better bounded:**

> Not reproducible in **16** consecutive full serial offline runs. Node id unrecoverable — `-rs` replaced pytest's default `-rfE` and suppressed the `FAILED` line (proven by controlled experiment); the session holding that stdout is gone and `lastfailed` has been overwritten since. Collection totals identical between the failing and passing runs (3915), so it was one test's outcome, not a different test set. Order deterministic, run was serial — neither ordering nor xdist implicated. The two previously-observed load flakes are clean at 40/40 each, and four targeted campaigns over every module carrying a live-machine-state `pytest.skip()` found no variance.

Nothing further from me on this one. Everything else from rounds 1-4 stands, `#435` and `#436` are filed rather than fixed here, and the branch is rebased onto `58f83aa` at `0b39ad7`.

Issue #423, opened 2026-09-01
Oracle capture: a failed /data structurally prevents /image, the 180s timeout is not settable, and promotion ignores later batches


Guust-Franssens:
Reported by **Hemang Patel (SES)** from the Airborne Services project, with per-view evidence across three capture batches. I have re-verified the source claim independently; the field evidence is his.

## ✅ Verified in current master — the render is not "attempted and failed", it is structurally skipped

`scripts/capture_tableau_oracle.py:585-589`:

```python
record["data"] = _capture_data(session, view_luid, out_dir / "data" / f"{stem}.csv", out_dir)
if record["data"]["status"] != "ok":
    return record          # <-- the render loop below is NEVER reached

for kind in ("png", "svg", "pdf"):
    ...
```

So **one slow `/data` endpoint costs two pieces of evidence.** A view whose data export times out can never acquire an image, however healthy `/image` would have been.

## The field evidence (his, ✅ verified at manifest level)

**"Availability Summary by Tail"** — three runs, two different days, identical signature:

| batch | when | result |
|---|---|---|
| original | Aug 17 20:17 | `HTTP 0` — `TimeoutError: read operation timed out` |
| retry | Aug 17 20:32 | identical |
| retry2 | Aug 18 14:46 | identical, a full day later |

Three identical failures across two days rules out a network blip — this reads as a view whose underlying query cannot export within 180 s server-side. **No image was ever attempted**, so it has no `image` key in any of the three records. Currently an unrecoverable blind spot.

**"Daily Monitoring"** — the recoverable case, and the more instructive one:

| batch | data | image |
|---|---|---|
| original | ❌ `HTTP 0` | never attempted (early exit) |
| retry | ❌ `HTTP 0` | never attempted (early exit) |
| retry2 | ✅ 373 rows, 18.0 s | ✅ **905,098 bytes, on disk** |

⚠️ The successful image exists at `_oracle/airborne-services-retry2/images/Daily_Monitoring__0979a4f9.png` — and the workbook's `reference/manifest.json` only ever cross-referenced **original** and **retry**. A third batch sat on disk, successful, and was never consulted.

(He notes the 1,101 → 373 row change between successful captures is expected: live Snowflake-backed data, not a frozen export. Worth recording so nobody reads it as corruption.)

## Three separately-fixable defects

[...]

PR #431, opened 2026-09-01
Decouple the Tableau capture legs and bound the request, refusing a truncated render as evidence (refs #423)

Guust-Franssens:
## Contract

**Invariant + direction.** A render leg is credited as evidence **only** when the payload is
structurally whole; every blocking phase of a bounded request completes inside its absolute deadline.

| direction | consequence | blocks merge? |
|---|---|---|
| **fail-open** — a truncated render recorded `status: ok` | the entry gate reports the view as covered, so a fidelity defect on it is not merely unverified but *believed verified*. Strictly worse than the suppression this PR replaces | **YES** |
| fail-closed — a complete render refused, or a request abandoned early | one retry, or one metered re-capture; the gap stays visible | no |

**Closed surface.**

*Every blocking phase a deadline must cover — `N = 7`:*

| phase | what bounds it |
|---|---|
| name resolution | ⚠️ **nothing of ours** — `getaddrinfo` runs before a socket exists |
| address iteration | ⚠️ **nothing of ours** — N candidates cost N timeouts before any socket. Pinned by `test_multiple_addresses_are_a_known_unbounded_phase` |
| TCP connect (one address) | socket timeout, narrowed by `_open` to what is left |
| proxy `CONNECT` | the watchdog, armed at raw-socket creation |
| **TLS handshake** | ⚠️ **was unbounded** — now the watchdog on the `SSLSocket`, plus a timeout narrowed to the remaining absolute budget |
| status line and headers | the watchdog, re-pointed at the wrapped socket |
| body | the watchdog, plus `_read_bounded`'s per-chunk check **and its outstanding-`Content-Length` check** |

Residuals stated, not implied: resolution and address iteration are bounded by the OS and the
candidate count. This is **hardening, not a wall-clock guarantee**.

*Every way a payload can be incomplete yet pass a prefix check — `N = 4`:*
premature close under `Content-Length` (transport); **close-delimited framing, where there is no
declared length at all** (structure only); chunk-level corruption with intact framing (CRC); a
well-formed prefix of a valid document (full parse / terminator).

**Independent oracle.** Loopback servers that the code under test does not produce: a peer declaring
`Content-Length: 1024` and sending 8 bytes; the same truncation framed HTTP/1.0 with no length; a
proxy that delays `CONNECT` then trickles a well-formed TLS record. The PNG fixture is checked
against a property the completeness code never inspects — its IDAT inflates to exactly the length its
own IHDR implies.

- **positive control:** `test_a_complete_payload_is_accepted` (png/svg/pdf) — without it, "refuse
  everything" would pass every negative test; `test_a_body_that_genuinely_finishes_in_time_still_returns`.
- **negative control:** `test_a_trickling_TLS_handshake_is_UNBOUNDED_without_the_deadline` (same
  fixture, deadline removed, runs to ~4s) and `test_the_same_sequence_is_UNBOUNDED_without_the_deadline`.

**Round route.** This is the fix round for a blind review's blocker. New class ⇒ simplify/split/descope
rather than another local guard — which is why grouping left this PR entirely.

---

## The estate evidence

`capture_tableau_oracle.py` captured each view in two coupled legs: a failed `/data` suppressed the
render. Measured on a live 360-view site: **0 rescues out of 46**, and **37 of those 46 are
worksheets**, so it is not dashboard-specific. Those 46 objects report BLIND in the entry gate, where
a fidelity bug is structurally unfalsifiable.

## Blocker: a premature EOF was credited as a successful render

`_read_bounded`'s deadline path treated any EOF before the deadline as a complete body.
`HTTPResponse.read1` does **not** raise `IncompleteRead` when a peer closes with bytes outstanding
under `Content-Length` — it calls `_close_conn()` and returns `b""`. So the **deadline path was
weaker than the plain `read()` it wraps.**

Reproduced against a loopback server answering `200`, `Content-Type: image/png`,
`Content-Length: 1024`, then only the 8-byte PNG signature:

| | before | after |
|---|---|---|
| `_request` with the salvage deadline | `status=200`, 8 bytes, `format_matches("png") == True` | `status=0`, `IncompleteRead(8 bytes read, 1016 more expected)` |
| `_request` without the deadline | `status=0`, `IncompleteRead(...)` | identical to the bounded path |
| `capture_view` after a `/data` 400 | `image.status="ok"`, SHA-256 recorded, `render_unestablished == 0` | `status="truncated"`, no digest, no file, `render_unestablished == 1` |

⚠️ **This inverted the point of the PR.** A decoupled `/image` must fail *loudly*.

Fixed at **both** layers, which are not redundant:

- **transport** — at EOF, refuse a response whose declared `Content-Length` is still outstanding,
  raising the same `IncompleteRead` the unbounded path raises so the two classify identically.
- **capture** — `payload_is_complete` validates structure before anything is written or stamped
  `status: ok`: PNG chunk walk with CRCs through IDAT/IEND, a full XML parse for SVG, `%%EOF` in a
  PDF's final bytes.

⚠️ **An earlier draft had only the capture-layer test, and deleting the capture check SURVIVED
mutation** — the transport had already refused the body, so the guard was never reached. The
independent surface is **close-delimited framing**, where there is no declared length to fall short
of; both layers now have their own anchor.

## Three fixtures asserted the broken behaviour was correct

- `test_capture_tableau_oracle_leg_decoupling.py` — signature + a partial IHDR: no CRC, no IDAT, no IEND.
- `test_capture_tableau_oracle_svg.py::_png` — docstring *"a minimal but genuinely valid PNG"* over
  IHDR + CRC and nothing else.
- both PDF fixtures — no `%%EOF`.

A fix validated against any of them was unfalsifiable. Real payloads now come from
`tests/png_fixtures.py`, shared rather than duplicated because a fixture wrong the same way in two
files is how one gets corrected and the other does not.

## MEDIUM (fail-closed): the watchdog was inert for the whole TLS handshake

`wrap_socket` **detaches** the raw socket the early watchdog points at, and with
`do_handshake_on_connect=True` the handshake runs inside that call — before
`_rearm_if_the_socket_was_replaced` can re-point anything. The only bound left was the socket timeout,
which **restarts** at the handshake instead of counting down the remaining budget.

Measured through a proxy delaying `CONNECT` by 0.10s, with `_open`'s own narrowing applied:

| | elapsed | vs a 0.16s deadline | watchdog |
|---|---|---|---|
| before | 0.327s | **+0.167s** | fired on `fileno=-1` |
| after | 0.165s | +0.005s | fired on a live `SSLSocket` |

Handshake deferred, `SSLSocket` installed while still abortable, watchdog re-pointed, timeout narrowed
to the remaining absolute budget, then `do_handshake()`. **Verification is unchanged** — it lives
inside the call being moved, and `test_the_deferred_context_forwards_verification_settings_untouched`
says plainly that it is a structural control, not a certificate-rejection proof.

⚠️ The existing TLS test could not see this: its fake `wrap_socket` returns immediately, sampling the
timer either side of the blocking phase rather than during it.

## A control falsified its own docstring — and found a live defect

`_svg_is_complete` justified not forwarding `str(exc)` with *"expat echoes the undefined entity
name"*. **Measured on CPython 3.13 that is false**: across ten malformed shapes every `ParseError`
message is fixed vocabulary plus a line/column.

But the same probe found that `<?xml encoding="<attacker text>"?>` raises
`LookupError: unknown encoding: <attacker text>` — **not a `ParseError`**, so it escaped the module
entirely into a traceback carrying the document's own bytes, crashing the capture. Now caught; both
halves pinned by `test_an_undefined_entity_is_the_control_that_makes_that_test_bite`, which fails if
either fact changes.

The entity refusal is likewise measured rather than assumed:
`test_the_entity_refusal_prevents_an_expansion_expat_would_otherwise_perform` demonstrates expat
expanding a 3-level entity to 1000 characters before showing the refusal stops short of it.

## Correction to this PR's earlier prose

The unqualified *"every HTTP 400 now attempts `/image`"* was **false** — the credential-classified 400
is a deliberate exception and makes zero image requests.

## The split

Multi-batch grouping is removed here and lands as
[`fix/oracle-multi-batch-provenance`](../../compare/fix/capture-leg-decoupling...fix/oracle-multi-batch-provenance):
independent behaviour, independent provenance invariants, and it carries the second blocker.

⚠️ **Stacked, not parallel, and that is forced rather than chosen**: `group_oracle_by_workbook.py`
imports the new `tableau_oracle_manifest`, and `test_group_oracle_multi_batch.py` is new in this PR,
so part 3 cannot target master.

⚠️ **Honest accounting:** this is 18 files / +5165−479, against 17 / +5453−507. The split removed
~1250 lines and blocker 1's proof added ~700, so the count barely moved. The value is scope
separation — one closed surface per PR — not size.

## Gates — by exit code

| gate | exit |
|---|---|
| `pytest tests/` | 1 — **3 known pre-existing engine-drift failures only** (`test_upstream_repro_pins.py`), 3875 passed |
| `pylint scripts` / `pylint conftest.py` | **0** ⚠️ it printed `10.00/10` while exiting 8 before the fix — read the code, not the score |
| `python tests/mutation_capture_legs.py` | **0** — 44 mutations, 42 caught, 1 cosmetic control survived, 1 absent-anchor control invalid, as declared |

Six new mutations cover this change: the EOF check removed; the completeness checker neutered; the
capture-seam call removed (a *separate* mutation, because a refactor could keep the checker perfect
and drop its only call site); SVG completeness reduced to a root-element check; the handshake run
inline again; the handshake left on a per-phase timeout.

## What CI found that a Windows-only run could not

The first push failed `test_an_abandoned_body_is_a_transient_failure_never_a_partial_success` on
Linux. `_abort_socket`'s `shutdown(SHUT_RDWR)` raises on Windows but yields a **clean EOF** on Linux
— the platform divergence `_read_bounded`'s EOF branch already records — so on Linux an aborted
trickle now hits the new outstanding-`Content-Length` check *first* and is refused as
`IncompleteRead` rather than as a deadline `TimeoutError`.

The behaviour is correct: still `NETWORK_ERROR_STATUS`, still no partial body, and `IncompleteRead`
is the more precise of the two because it names the byte count the peer still owed. What was wrong
was the test's marker list — whose own docstring already warned that asserting on **one** mechanism
is how it broke once before. There are now **three**, the list is a union again, and the docstring
says why, so the property under test stays the outcome rather than the mechanism.

Both jobs now pass on both PRs.

---

# Round 3 — the defect CLASS, not two more instances

A blind review found two more fail-open blockers and one fail-closed MEDIUM. All three were **a raw
substring or tail search standing in for a structural boundary** — the class already diagnosed on the
`Groups` error: *"substring matching can't tell a scalar from a suffix."* Fixing them one at a time is
what produced three rounds, so this round removes the class.

| the scan | what it could not tell apart |
|---|---|
| PNG signature prefix | an 8-byte truncation from a whole image |
| `<!ENTITY` before the first `<svg` | a DTD from one hidden behind `<!-- <svg -->` — the boundary is **attacker-movable** |
| `%%EOF` in the last 2 KiB | the **final** revision's trailer from an **earlier** revision's |

Every marker-scan constant is gone from `tableau_payload_facts.py` — `_PROLOG_SCAN_BYTES`,
`_ENTITY_DECLARATION`, `_PDF_TAIL_BYTES`. Nothing replaced them with a fourth scan.

## BLOCKER 1 — an earlier PDF trailer made a truncated render look complete

| | before | after |
|---|---|---|
| `payload_is_complete("pdf", truncated)` | `(True, '')` | `(False, 'the PDF does not END with %%EOF, so it was cut short')` |
| surviving `%%EOF` distance from the end | 105 bytes | unchanged — the fixture still reproduces |
| strict `pypdf` on the complete file | `Title='LATEST-REVISION'` | — |
| strict `pypdf` on the truncated file | `Title='ORIGINAL'` | — |

Not a corrupt file — a file describing a document the customer **no longer has**, credited as current
evidence with a digest and a path.

The final revision's trailer is now **followed**: `%%EOF` must be the last non-whitespace token, the
`startxref` before it must carry a numeric offset, and that offset must land inside the file on a
cross-reference (the `xref` keyword, or an object header for a PDF 1.5+ xref stream).

## BLOCKER 2 (security) — a comment bypassed the SVG entity refusal

| placement | first `<svg>` | first `<!ENTITY>` | before | after |
|---|---|---|---|---|
| comment decoy | byte 14 | byte 45 | expanded to 1,000 chars, `(True, '')`, `status: ok` with digest + path | refused |
| PI decoy | byte 29 | byte 52 | same | refused |

SVG is now settled by an **expat parse**: entity declarations refused from `EntityDeclHandler`
wherever they sit with nothing ever expanded; external entity references refused so no parse of ours
can reach the network; and the buffer must parse to its **last byte**, which is what catches
truncation and junk after the root.

⚠️ Namespace processing is deliberately **off**. `ElementTree.fromstring` is namespace-aware and
rejects an undeclared prefix, so a drawing using `xlink:href` without `xmlns:xlink` was refused as
broken. Measured: non-namespace parsing is **strictly more permissive on well-formed documents and
strictly stricter on both bypasses**.

## MEDIUM — the encoding exception family escaped

⚠️ **My first draft assumed expat would unify the exception family. It does not** — measured, not
believed. An unrecognised `encoding=` still raises `LookupError` **carrying the declared name
verbatim**; `utf-32`/`utf-7`/`shift_jis`/`big5`/`gb18030` raise `ValueError: multi-byte encodings are
not supported`. Neither is an `ExpatError`. Both now become
`the SVG declares an encoding this parser cannot decode`; neither message is forwarded, and the test
asserts no ≥6-character run of the declared encoding survives into the diagnostic.

## The fixtures pinned it, for the third round running

All four positive PDF fixtures failed strict `pypdf` with `startxref not found` — a `%PDF-` header and
a `%%EOF` with nothing in between, which is exactly what a tail window accepts and a parser does not.
`tests/pdf_fixtures.py` now builds real PDFs with computed cross-reference offsets. **`pypdf` and
`pillow` are dev dependencies**, so the independent oracle runs in CI rather than on one machine.

## The class, tested as a class

`test_no_completeness_check_can_be_satisfied_by_bytes_that_are_not_the_end` appends **1, 16, 100 and
1,024** non-whitespace bytes to a complete payload of each format. Every marker scan in this module's
history fails it.

⚠️ The short suffixes are the discriminating ones: the PDF window was 2,048 bytes, so the pre-existing
4,096-byte test measured **the window**, not the property — **100 bytes passed**.

⚠️ PNG is excluded from the trailing-whitespace control and has its own test. Writing that control
over all three formats asserted the opposite and **failed, correctly**: a PNG ends at IEND and has no
trailing-whitespace concept. Formats differ; a control that pretends otherwise weakens the strictest.

## Gates — enumerated from `.github/workflows/checks.yml`, not remembered

| job | step | exit |
|---|---|---|
| checks | ruff format --check | **0** |
| checks | ruff check | **0** |
| checks | pylint scripts | **0** ⚠️ printed `10.00/10` while exiting **28** before the handler disables landed |
| checks | pylint pbip-model-refresh / powerbi-ai-readiness / conftest.py | **0** / **0** / **0** |
| checks | Tests (`pytest -q tests .github/skills`) | 1 — **3 known engine-drift failures only**, 4229 passed |
| checks | Migration spec contract gate | **0** |
| checks | **Privacy gate** (`set_data_folder.py --check`) | **0** |
| checks | Data model gate | **0** |
| checks | Navigation index | **0** |
| checks | Shipped capabilities reach agents | **0** |
| checks | Shared conventions reach every agent | **0** |
| checks | AI-readiness report | **0** |
| windows-bundle | Tests (pbip-model-refresh) | **0** — 306 passed, 3 skipped |

`python tests/mutation_capture_legs.py` — **0**: 49 mutations, 47 caught. Five new ones restore each
marker scan; ⚠️ one **SURVIVED first time** because it delegated to the *fixed* function, and was
re-written to re-implement the defect rather than wrap it.

## Could not be verified — stated, not claimed

- The **Linux-only clean-EOF abandonment signature**: Windows cannot produce it. The exact-head Linux
  CI job passing is the only evidence, and that is weaker than a local reproduction.
- **Any recapture against the live Tableau estate.** Every result here is from loopback servers and
  synthetic payloads; the 46 render-less objects have not been re-captured.


Guust-Franssens:
**Round-2 verification note.** PR #431 is `CONFLICTING` against `feat/oracle-view-type` right now, so GitHub creates no check run for it — by design, per round-2 finding 3: I am **not** re-rebasing this branch until #422 merges.

To avoid handing back work verified only on my interpreter (which is exactly how the `ast.unparse` dialect bug reached CI last round), I verified the same tree on a throwaway branch rebased onto the current base tip and dispatched `checks` there:

* run: https://github.com/Guust-Franssens/tableau-to-powerbi-migration/actions/runs/33579632370 — **success**, both jobs (`checks`, `windows-bundle`)
* that tree was `git rev-list --count HEAD..origin/feat/oracle-view-type` = **0**, ruff **0**, `pylint scripts` **0**
* the branch has been deleted; the run record persists

Locally on the PR head (`46c0c00`): serial `pytest -q` → **4140 passed, 9 skipped**, exit 1 from only the three documented `test_upstream_repro_pins.py` engine-drift tripwires; `tests/mutation_capture_legs.py` → **exit 0**, 48 mutations all scoring as declared.

The single pre-merge rebase onto #422's final tip, and the gate run on the tree that actually lands, still come after it merges.

Guust-Franssens:
**Your premise is now confirmed against a live 360-view site, not reasoned about.**

Full customer-shaped dry run: `_runs/407-dryrun-gates`, trial Tableau Cloud site `fabric-migration-lab`
(48 workbooks / 360 views), engine 2.339.0, `capture_tableau_oracle.py --images --reference-best`.

## The measurement

Real dashboards return **HTTP 400** on the `/data` leg — correctly, a dashboard has no tabular data:

```
67/360  Dashboard 1               FAILED (failed): HTTP 400
69/360  Dashboard 2               FAILED (failed): HTTP 400
71/360  Dashboard 3               FAILED (failed): HTTP 400
81/360  GDP per Capita Dashboard  FAILED (failed): HTTP 400
```

And the render legs never ran. Probing the captured files by view LUID:

```
view                       png?   svg?   csv?
Dashboard 1                False  False  False   <- /data returned 400
Dashboard 1 (other wb)     True   True   True    <- /data ok
Dashboard 2                False  False  False
Dashboard 3                False  False  False
GDP per Capita Dashboard   False  False  False
```

The control is in the second row: a *different* view also called "Dashboard 1", whose `/data`
succeeded, has all three artifacts. So this is the coupling, not a per-view outage.

## Why this is worse than a missing CSV

The consequence is not "no numbers for dashboards". It is **no reference render for any dashboard on
the site**, which lands directly on the new entry gate (#428):

```
REFERENCE READINESS: FINDINGS - 3/220 expected page(s) ready; 217 blind
  - dashboard 'Overview'   -> BLIND - no reference render, so a fidelity bug here is unfalsifiable
  - dashboard 'Stale Content' -> BLIND
  ...
```

Dashboards are exactly the pages a customer looks at, and exactly the pages that can never be
verified while the legs are coupled. `/data` failing on a dashboard is **expected Tableau behaviour**,
so this is not an edge case — it is the normal path for every dashboard on every site.

⚠️ It also makes the gap self-perpetuating in the way the issue title says: the gate reports the page
BLIND, the operator re-runs the capture, `/data` returns 400 again, and no render is produced again.

## What this does not tell you

It does not tell you the decoupling is *correct* — only that the failure it targets is real and
site-wide. Worth having a control that a decoupled `/image` still fails loudly when the render itself
fails, rather than silently producing a zero-byte file that the entry gate would then credit.

Guust-Franssens:
**Correcting my previous comment — the conclusion was right, my explanation of it was wrong twice.**
The capture has since finished (312/360 captured, 6823s), so this is the complete artifact rather than
the partial one I sampled.

## What I got wrong

1. ❌ *"no dashboard on any site ever gets a reference render"* — **false**. **52 of 60 dashboards got
   renders.** I generalised from four dashboards sampled mid-run.
2. ❌ *"a dashboard has no tabular data, so `/data` 400 is the normal path for every dashboard"* —
   **false**. 52 of 60 dashboards served `/data` fine. Whatever causes the 400 on the other 8, it is
   not "dashboards inherently have no data".

Both errors came from measuring an in-flight capture and reading a pattern into six log lines.

## What the completed manifest actually shows — and it is stronger evidence for this PR

```
view_type   data                 image           n
worksheet   ok                   ok            261
dashboard   ok                   ok             52
worksheet   failed               None           37
dashboard   failed               None            8
worksheet   ok                   transient       1
worksheet   source_credential    None            1

data-failed AND image ok   : 0
data-failed AND image not  : 46
```

**Zero rescues out of 46.** Every single view whose `/data` leg failed also lost its render, and not
one recovered. That is a perfect correlation across a 360-view site — a much cleaner demonstration of
the coupling than the dashboard story I told, and it does not depend on any theory about *why* a
particular view's data leg failed.

⚠️ **The population is mostly worksheets, not dashboards: 37 of the 46 victims.** So framing this as a
dashboard problem understates it and points the fix at the wrong place.

## The cost, measured on the entry gate

Running #428 against the same bundle with this completed oracle:

```
pages_expected 220 | pages_ready 42 | pages_blind 178 | evidence_records 313 | evidence_untyped 0
```

46 of those blind pages are blind *only* because a `/data` failure suppressed a render that would
otherwise have been captured in the same pass. That is the direct, quantified benefit of decoupling.

## Control worth adding

A decoupled `/image` must still fail loudly when the render itself fails — never a zero-byte or
error-page artifact that the entry gate would then credit as evidence. The `transient` row above shows
image status is already tracked independently, which is the hook for that assertion.

PR #433, opened 2026-09-01
fix(check_unit): correct the expected-page rule, delete the circular oracle denominator, and discover both oracle directory names

### LINKED ISSUE #432: check_unit page checks: expected_pages contradicts the engine, oracle-coverage grades the artifact against itself, and oracle discovery knows one of two documented directory names
`scripts/check_unit.py` has three defects that make its two page checks unreliable, and one of them
is the repo's most-repeated failure class: unassessable input collapsing into the clean bucket.

All measurements below are against engine **2.339.0** (`python scripts/engine_source.py`) and the
real 44-workbook estate run in `_runs/estate-2.339.0-20260829/`.

## 1. `expected_pages` uses a page rule the engine has never used

`expected_pages` (`scripts/check_unit.py:572`) assumes one page per Tableau **dashboard**. The engine
emits a page per dashboard **and** a page per **orphan** worksheet - one no dashboard placed:

| `twb_to_pbir.py` | behaviour |
|---|---|
| :14549 / :14551 | `_emit_page` + `page_order.append` per dashboard |
| :14557-14558 | second pass over `ir["worksheets"]`, skipping `ws["name"] in placed` |
| :14709 | `page_order.append` for each surviving orphan sheet |

Consequence: **19 of 43** workbooks in the estate run have zero dashboards, so `expected_pages`
returned `[]` for nearly half the estate. `check_page_parity` then reported every emitted page as an
extra page with no Tableau counterpart (e.g. `Section 08 - Organizing Data`: 0 expected, 11 emitted),
and the empty list fed defect 2 below.

Dashboards + orphan worksheets is only the **candidate** count. The engine legitimately drops a
candidate in three declared cases - :14529 (dashboard with no supported visuals), :14558
(`VT_UNSUPPORTED`), :14562 (no usable field bindings) - so the gate needs three buckets: candidates,
`dropped_explained`, `dropped_unexplained`, and should fail only on the last.

⚠️ **Attribution must come from `viz_fidelity[]`, never `pbip_warnings[]`.** `pbip_warnings` is a flat
list of prefixed **strings** with no scope/name field (see `tests/fixtures/handover-pbip-warnings.json`).
Measured across the estate run: its **193 entries explained 0** of the 21 dropped candidates, while
`viz_fidelity[]` explained **21 of 21** - including dashboards, which `migrate_estate._viz_fidelity`
projects as `{"worksheet": <dashboard name>, "visual_type": "dashboard"}`. A gate built on
`pbip_warnings` produces a plausible count that mis-attributes every drop.

⚠️ **There is a fourth page behaviour.** `twb_to_pbir.py:14719-14727` is a **zero-page crash guard**:
when every candidate is dropped it ships ONE synthetic visual-less page (`page-empty*`, displayName
`"No visuals rebuilt"`, :13091) because a PBIR with an empty `pageOrder` crashes Power BI Desktop on
open. Two estate workbooks (`shapes_test`, `Meridian Multi-Source (3 systems)`) ship one, and a
corrected parity check fails both unless this is encoded as a declared extra.

## 2. `check_oracle_coverage` grades the artifact against itself

`scripts/check_unit.py:925`:

```python
pages = expected_pages(target) or actual_pages(target)
```

When the expected set is empty/falsy the denominator becomes the **actual pages of the artifact being
graded**, so coverage is actual/actual and reports a perfect score regardless of what is missing.
Reproduced on a directory with no `migration-spec.json`, two emitted pages and a full reference
manifest:

```
check_oracle_coverage -> PASS   pages=2  visual_present=2  grade=validation-grade
```

This fires for **every engine bundle** (verified: 0 `migration-spec.json` under
`_runs/estate-2.339.0-20260829/`) and, before defect 1 is fixed, for every dashboard-less workbook.
`_migration_spec` (:285) is not itself wrong - it looks in the two places a *unit* keeps its spec, and
a bundle legitimately has none - but it is the entry point to the same bug.

An expected set that cannot be established must be an explicit blocking **cannot assess**, never a
pass.

## 3. Oracle discovery knows only one of two documented directory names

`_oracle_dirs` (:836-838) looks only for `_oracle/`. Both of these are documented and both exist on
disk:

| name | documented in | on disk |
|---|---|---|
| `_oracle/` | `AGENTS.md` (`capture_tableau_oracle.py --out _oracle`), `docs/operator-runbook.md:1197` | yes (`--out` is **required**, `capture_tableau_oracle.py:712`) |
| `oracle/` | `scripts/work_dirs.py` `CANONICAL_SUBDIRS`, `AGENTS.md` "Canonical work layout" | yes - `_runs/<NNN>-<slug>/oracle/` beside `bundle/` |

So the docs are coherent and the code is incomplete: a real capture beside a bundle at
`_runs/<NNN>-<slug>/bundle` is invisible, and oracle-coverage reports "no oracle manifest found" for
evidence that exists. Verified on a real capture staged as `_runs/077-global-superstores/{bundle,oracle}`:
**26 manifest views** are readable under the bare `oracle/` name; discovery returned 0 directories.

The slug matching at :892 / :931 is **not** a defect - `_reference_oracles`, `_oracle_capture_oracles`
and `check_oracle_coverage` all key on the same `_slug`, and `view_name`/`view_url_name` are
slug-equivalent on real data (`'Shipping Cost'` vs `'ShippingCost'`).

## Acceptance

- `expected_pages` = dashboards + orphan worksheets; declared drops subtracted, read from
  `viz_fidelity[]`; the crash-guard page not counted as an extra.
- No fallback in `check_oracle_coverage`; an unestablished expected set is a blocking `NOT_CHECKED`
  with a detail naming why and what to run.
- Oracle discovery accepts both `_oracle/` and `oracle/`.
- All 44 estate workbooks pass page parity with 0 unexplained drops.
- Every new behaviour is mutation-tested: break it, confirm a test fails.

Guust-Franssens:
## What this fixes

Three defects in `scripts/check_unit.py`, the post-agent unit gate. Everything below was measured
against engine **2.339.0** (`python scripts/engine_source.py` -> the canonical plugin) and a real
44-workbook 2.339.0 estate run, not reasoned from the code.

### Defect 1 - `expected_pages` used a page rule the engine has never used

**Truth found.** `twb_to_pbir.py` emits a page per Tableau **dashboard** (`_emit_page` :14549 ->
`page_order.append` :14551) **and then walks `ir["worksheets"]` a second time**, emitting a page for
every sheet no dashboard placed (:14557-14558 `if ws["name"] in placed ... continue` -> :14709
`page_order.append`). The old rule (dashboards only, :572) returned an **empty** expected set for
**19 of 43** workbooks in the estate run.

The three declared drops are at the lines the brief named, but **only two of them emit a warning at
that site**:

| engine site | what it does | warning at that site |
|---|---|---|
| :14529-14532 | dashboard with no supported visuals -> `continue` | ✅ `_warn("dashboard", db["name"], "no supported visuals on this dashboard")` |
| :14557-14558 | `ws["name"] in placed` **or** `visual_type == VT_UNSUPPORTED` -> `continue` | ❌ **none here** - the unsupported classification is warned upstream and reaches `viz_fidelity` via `migrate_estate._viz_fidelity`'s `elif vt in (None, "unsupported")` branch |
| :14562-14566 | `not _query_state_complete(...)` -> `continue` | ✅ `_warn("worksheet", ws["name"], "... no usable field bindings (skipped)")` |

**A fourth behaviour the brief did not mention, and it fails units on its own.** :14719-14727 is a
**zero-page crash guard**: when every candidate is dropped the engine ships ONE synthetic visual-less
page (`_sanitize("page-empty")`, displayName `"No visuals rebuilt"` :13091) because a PBIR with an
empty `pageOrder` crashes Power BI Desktop on open. Two of the 44 estate workbooks
(`shapes_test`, `Meridian Multi-Source (3 systems)`) failed parity with *"1 emitted page with no
Tableau counterpart"* until this was encoded.

**The `viz_fidelity` vs `pbip_warnings` trap is real and is worse than stated.** `pbip_warnings[]`
is a flat list of prefixed **strings** with no scope/name field (`tests/fixtures/handover-pbip-warnings.json`).
Across all 43 workbooks of the estate run its **193 entries explained ZERO** of the 21 dropped
candidates - it does not merely omit the dashboard name, it does not carry these warnings at all.
`viz_fidelity[]` explained **21 of 21**, including dashboards, which `migrate_estate._viz_fidelity`
projects as `{"worksheet": <dashboard name>, "visual_type": "dashboard"}`.

**Changed.** `expected_pages` returns dashboards + orphan worksheets (orphan = not referenced by any
dashboard zone's `worksheet_id`, walked recursively; `parse_tableau.py:1027` + `:1096` prove that id
joins to the worksheet's `id`). A new `page_expectation` computes the three buckets once for both
page checks; `page_drop_explanations` reads `viz_fidelity[]` only. `check_page_parity` subtracts
declared drops and the crash-guard page, and fails only on what nothing accounts for.

**Result on real data: 44/44 estate workbooks now PASS page parity, with 0 unexplained drops.**

### Defect 2 - the gate graded the artifact against itself

**Truth found.** Confirmed by running the pre-fix and post-fix code over the same directory (no
`migration-spec.json`, 2 emitted pages, a full validation-grade reference manifest):

```
BEFORE  check_oracle_coverage -> PASS         pages=2  visual_present=2  grade=validation-grade
AFTER   check_oracle_coverage -> NOT_CHECKED  pages=0  visual_present=0  grade=not checked (...)
```

`_migration_spec` (:285) is **not itself wrong** - it looks in the two places a migration *unit* puts
its spec, and an engine bundle legitimately has none (verified: 0 `migration-spec.json` under
`_runs/estate-2.339.0-20260829/`). But it is the same bug, as the brief suspected: `None` from
`_migration_spec` -> `None` from `expected_pages` -> the `or actual_pages(target)` fallback -> a
perfect score. Defect 1 widened the same hole to every dashboard-less workbook, since `[]` is falsy too.

**Changed.** The fallback is deleted. An expected set that cannot be established returns a blocking
`NOT_CHECKED` (exit 2) naming *why* and what to run. Two unassessable states are distinguished in the
detail: no spec at all, and a spec declaring neither dashboards nor worksheets.

Side effect worth naming: because the denominator is now the **Tableau** page set, oracle matching is
against Tableau names - which is what a Tableau oracle manifest keys on. The old fallback matched
PBIR display names, which hand-finishing frequently renames (5 of 16 committed example units).

### Defect 3 - oracle discovery vs the docs: the code was wrong, not the docs

**Truth found.** Both names are documented and both are real:

| name | where it is documented | where it exists on disk |
|---|---|---|
| `_oracle/` | `AGENTS.md` (`capture_tableau_oracle.py --out _oracle`), `docs/operator-runbook.md:1197`, `scripts/README.md` | yes - `--out` is **required**, it has no default (`capture_tableau_oracle.py:712`) |
| `oracle/` | `scripts/work_dirs.py` `CANONICAL_SUBDIRS`, `AGENTS.md` "Canonical work layout" | yes - `_runs/<NNN>-<slug>/oracle/` beside `bundle/` |

`AGENTS.md` explicitly says `capture_tableau_oracle.py` keeps its `_oracle*/` default while the
canonical layout uses `oracle/`. So the documentation is coherent and the **code was incomplete**: a
real capture beside a bundle at `_runs/<NNN>-<slug>/bundle` was invisible and oracle-coverage
reported "no oracle manifest found" for evidence that existed.

**Changed.** Discovery accepts both names, deduplicated, beside the unit / target / target's parent.
Verified on a real capture: staged `_runs/077-global-superstores/{bundle,oracle}` reads **26 manifest
views** under the bare `oracle/` name (0 before).

**The slug matching at the old :892 / :931 is NOT a defect.** `_reference_oracles`,
`_oracle_capture_oracles` and `check_oracle_coverage` all key on the same `_slug`, and the
`view_name` / `view_url_name` fallback is slug-equivalent on real data (`'Shipping Cost'` vs
`'ShippingCost'` -> `shippingcost`). Left alone; see "Not verified" for the one residual nit.

Also documented in the `scripts/README.md` `check_unit.py` row, and in the `--oracle-dir` help.

## Mutation testing - 15 mutations, all caught, plus a negative control

Every new behaviour was broken in the production code and the covering test confirmed to **fail**
(the harness scores only a genuine pytest `N failed`; a collection/import error is reported BROKEN,
never caught). A no-op comment edit was run through the same harness and correctly **SURVIVED**,
proving the harness can report a survivor.

| mutation | verdict |
|---|---|
| orphan worksheets dropped from the candidate set | CAUGHT (2 failed) |
| zone walk stops recursing, so placed sheets look like orphans | CAUGHT |
| declared drops no longer subtracted in `check_page_parity` | CAUGHT |
| `explained_drops` emptied at source (both consumers) | CAUGHT (2 failed) |
| oracle denominator stops excluding declared drops | CAUGHT |
| explanations read from `pbip_warnings` instead of `viz_fidelity` | CAUGHT |
| every warned worksheet counts as a drop, emitted or not | CAUGHT |
| unreadable handover reported as if reasons had been consulted | CAUGHT |
| circular denominator restored (`expected or actual`) | CAUGHT (3 failed) |
| empty candidate list treated as a real zero-page expectation | CAUGHT |
| canonical `oracle/` removed from discovery | CAUGHT |
| documented `_oracle/` removed from discovery | CAUGHT |
| crash-guard placeholder counted as an unexplained extra | CAUGHT |
| placeholder identity drops the id clause | CAUGHT |
| placeholder identity drops the name clause | CAUGHT |
| **NEGATIVE CONTROL** - no-op comment edit | **SURVIVED** ✅ |

Additionally, the whole pre-fix `check_unit.py` was checked out under the new tests: **12 of 13**
new/changed tests failed. The one that passed
(`test_expected_pages_finds_placed_worksheets_at_any_zone_depth`) was passing for two reasons -
vacuity mode 4 - and was strengthened with a second loose worksheet so the old dashboards-only answer
no longer satisfies it.

### Vacuity self-audit

1. **Exit code against its own constant** - `test_unassessable_oracle_coverage_fails_the_whole_run_closed`
   asserts the literal `2`, not `cu.EXIT_NOT_CHECKED`.
2. **Assertion in an unentered branch** - each new test was run against pre-fix code and observed to
   fail on the intended assertion.
3. **Clause implied by its siblings** - the two clauses of `_is_engine_placeholder_page` are pinned
   by two *different* tests (a real page titled like the placeholder; a `page-empty*` page the author
   filled in), and each clause has its own mutation.
4. **Existing test now passing for two reasons** - the one instance found is named and fixed above.

## Changed behaviour a reviewer should look at

`test_brownfield_canonical_bundle_stays_quiet` was rewritten (renamed
`..._proposes_no_reorganisation`). It **could not survive any correct fix**: it passed only because
`page-parity` raised a *false* `PRECONDITION_FAILED` on `tests/fixtures/check-gates-dirty` (its
`migration-spec.json` is the placeholder `{"workbook": ...}`, so "0 expected vs 1 emitted" read as an
extra page) and the run aborted before any `NOT_CHECKED` row existed. `_render_brownfield`
deliberately prints the inventory whenever a check reports a missing input, and `oracle-coverage` on
that fixture has no manifest, so the block appears under *any* correct parity result. The rewritten
test asserts what it actually cared about - recognised shape, no reorganisation plan - plus the new
assertion that the false `PRECONDITION_FAILED` is gone.

## Deliberate non-changes

- **Parity verdict stays count-based, not name-matched.** Renaming a PBIR page during hand-finishing
  is legitimate: 5 of 16 committed example units do it (`fast-fashion-impact` ships `Dashboard 1` as
  `Waste Not, Wear Not`). A name-matched verdict fails all five. `dropped_unexplained` is name-matched
  **attribution** and is documented as rename-sensitive.
- **`_migration_spec` is unchanged.** It is correct; see Defect 2.
- **No `view_type` awareness.** Not imported, not depended on. The seam for it is `page_expectation`,
  which already returns per-page rows both checks consume.

## Gates

| gate | result |
|---|---|
| `ruff format` / `ruff check` on both files | clean |
| `pylint scripts/check_unit.py` | **exit 0** (10.00/10) |
| `pylint tests/test_check_unit.py` | exit 22, **byte-identical finding set to master's baseline** (import-error x4, too-many-lines, unused-argument x2) - no new findings |
| `pytest tests/` | **3 failed, 3576 passed, 6 skipped** - the same 3 known engine-drift tripwires in `test_upstream_repro_pins.py` as on master (baseline: 3 failed, 3560 passed) |
| `scripts/check_navigation_index.py` | exit 0 |
| `scripts/sync_agent_conventions.py --check` | exit 0 (AGENTS.md's shared block untouched) |
| `scripts/check_agent_capabilities.py` | exit 0 |

## Not verified / known ceilings

- **Renaming defeats drop attribution, by construction.** If one candidate is renamed *and* another
  vanishes unexplained, the counts can balance and the gate passes. That is the pre-existing
  count-parity ceiling; this PR does not remove it, and the docstring says so.
- **A unit with orphan worksheets and no handover slice now fails parity** where the dashboards-only
  rule passed it. That is the honest state (a Tableau sheet has no page and nothing says why), but it
  is a real behaviour change for legacy units. **None of the 16 committed `examples/*` units has a
  handover directory**, and 6 have dropped orphans - they were not previously run through this gate in
  CI (only `examples/shipping-kpis`, which has 0 orphans, is used, by a path-ceiling test), so nothing
  regressed, but an operator re-checking an old unit will see it.
- **Expected pages cannot be reconstructed for an engine bundle.** `viz_fidelity` carries worksheets
  and warned dashboards but not *placement*, so orphan-ness is not derivable from a handover. The gate
  says so and points at `parse_tableau.py`; it does not guess.
- **The estate validation is synthetic in one respect:** I staged each workbook's spec, PBIR report
  and handover slice into a unit-shaped folder to run the real `check_page_parity`. The inputs are
  genuine engine output; the folder layout is not.
- **`viz_fidelity` `status: "warned"` is the only signal used.** I did not verify that the engine
  never drops a page while recording `status: "rebuilt"`; on this estate it never did (21/21 drops
  were `warned`), but that is an observation over one estate, not a guarantee.
- **Grade wording is misleading when a manifest exists but every leg failed:** `grade` falls back to
  `"not checked (no oracle manifest found)"` even though the manifest was read (observed on the real
  `global-superstores-db` capture: 26 views read, `data_ok: 0`). Pre-existing and reachable before
  this PR; left alone as out of scope.
- **`_slug("")` collision:** an oracle record with neither `view_name` nor `view_url_name` creates an
  empty-string key that a page whose name slugs to empty could match. Marginal, pre-existing, untouched.
- **Not run in Power BI Desktop.** This is a static gate; nothing here was validated against a live
  model or a rendered report.

Fixes #432


Guust-Franssens:
# Round 2 — all four findings fixed, and the headline number retracted

Every finding reproduced exactly as described. I built the four reproductions myself before touching
anything, and they are committed as tests. Rebased onto `origin/master` first, so everything below is
measured on the integrated tree.

**First, the retraction.** *"44/44 estate workbooks PASS"* was mechanically produced and I am
withdrawing it. Two things were wrong with it:

- **The absent-candidate count was 23, not 21 — you were right.** My count compared candidates against
  *all* emitted pages, so the two crash-guard workbooks (`shapes_test`, `Meridian Multi-Source
  (3 systems)`) looked balanced at 1-vs-1 and were skipped. Counting against pages that actually carry
  a visual gives 23.
- **All 23 were accepted, and 2 of them should never have been.** The honest number is now
  **42 of 44**, with the 2 failures named and actionable. A gate that passes everything is as useless
  as one that fails everything, and I shipped the former.

---

## 1. HIGH — any warning excused a missing page, including one from another workbook

**Reproduced, both halves.** With page `B` deleted and no warning: `PRECONDITION_FAILED` ✅. Add a
`tier: "degraded", evidence: "emitted+linted"` row: **PASS** ❌. Add a tier-`empty` row belonging to
`"Different Workbook"`: **PASS** ❌.

**Fixed with two independent guards.**

*Workbook identity.* A handover slice explains pages only for a workbook whose artifacts this unit
ships. The engine names a report folder `<workbook name>.Report` — verified on all 44 workbooks of the
estate run — so `_unit_workbook_keys` binds through the artifact stem. Slices that do not bind are
reported in `unbound_workbooks`, and a page they would have explained now says so by name.

*Structured non-emission evidence.* The only accepted proof is **`viz_fidelity[].tier == "empty"`**.
`migrate_estate._fidelity_tier` (:1104) defines that as *"no faithful visual emitted"*, against
`degraded` = *"a rendered visual whose warning is a genuine degradation"*. I checked the two
neighbouring fields you would expect to help and neither can decide it:

| field | why it cannot decide non-emission |
|---|---|
| `status: "warned"` | spans **both** outcomes — it is on every degraded row too |
| `evidence: "emitted+linted"` | appears on **45 `empty`-tier rows** in the same estate run |
| reason text | distinguishes them, but prose is not a contract — and per your instruction, not used |

Anything else is recorded as **ambiguous** and leaves the drop **unexplained**, annotated with why
(`"needs tier='empty'; got tier='degraded': …"`).

**On the four caption-fallback drops you flagged.** I checked them: their rows carry
`visual_type: "unsupported"` → `tier: "empty"`, so they remain explained — correctly. What was wrong
about them was the *reason text*, which shows a caption-fallback binding note as the first of several
warnings on that worksheet. That is precisely why the new rule ignores reason text.

**Cost of the strictness, measured:** 21 of 23 absent candidates still explained; **2 now fail
closed** —

- `global_superstores_db` → `Story 1`: *"no supported visuals on this dashboard"*, `tier: "degraded"`
- `Section 13 - Tableau 63 Charts` → `Pareto (method2)`: *"…has no usable field bindings (skipped)"*,
  `tier: "degraded"`

Both are **real engine drops the engine cannot express**. `_fidelity_tier` derives `empty` from
`visual_type` alone, so a dashboard-scope row (site `:14529`) and the query-state skip (site `:14562`)
both report `degraded` for a candidate that produced no page at all. Filed upstream with the
reproduction and three suggested fixes: **Yarbrdab000/tableau-fabric-skills#188**. Until it is
answered they fail closed here and the message names the gap.

## 2. HIGH — a partially malformed spec became a smaller *trusted* denominator

**Reproduced:** `worksheets` as an object → `expected_pages == [{"id": "dash.a", "name": "A"}]`, parity
**PASS**, oracle **PASS** at **`grade=validation-grade`** ❌.

**Fixed by validating the spec whole before deriving anything.** `docs/migration-spec.schema.json`
requires `dashboards` and `worksheets` as arrays whose entries carry `id` and `name`. `_spec_pages`
now refuses — explicit reason, blocking `NOT_CHECKED`, never a partial list — on: a missing required
array, a non-array, a non-object entry, an entry with no usable identity, and a zone tree that cannot
be walked. After: `expected_pages -> None`, parity `NOT_CHECKED`, oracle `NOT_CHECKED`,
grade *"expected page set could not be established"*.

One deliberate non-refusal, stated because it looks like laxness: `name → title → id` is kept as the
identity fallback. An entry with only an `id` still **occupies a slot in the denominator**, so keeping
it is the conservative choice — it will never match a page and fails closed. *Skipping* it is what
shrinks the denominator. Both halves are pinned by separate tests and separate mutations.

## 3. HIGH — the crash-guard accommodation was wrong in both directions

**Reproduced:** an empty `page-empty*` page renamed to `A` → **PASS**, `grade=validation-grade` ❌; a
`page-empty*` page holding 3 real visuals → false **`PRECONDITION_FAILED`** ❌.

**Fixed by making page content evidence.** `actual_pages` now carries a measured `visuals` count, and:

- a page with **zero visuals renders nothing and never certifies a candidate as rebuilt**;
- the placeholder is `page-empty*` **and** `"No visuals rebuilt"` **and** `visuals == 0` — three
  clauses, each with its own test and its own mutation;
- a zero-visual page that is **not** the declared placeholder is now reported as a `blank_pages`
  finding rather than passing silently.

After: 3a is `PRECONDITION_FAILED`, and end-to-end `run_all` → **exit 4, stopped after page-parity**
(oracle is never reached). 3b is `PASS`. You were right to flag this risk when I first reported the
fourth page behaviour: encoding a synthetic placeholder as expected did turn "this workbook produced
nothing" into a pass.

## The next layer, asked immediately — and it was there

You said to ask what the *next* layer of "explanation" looks like. It was the **exemption** layer, with
the same defect in two forms, both now fixed and mutation-pinned:

1. **An exemption applied whether or not the page was absent.** Signing *"C was merged away"* while C
   is still emitted shrank the expected count by one, balanced the books, and hid the absence of `A`
   entirely — invisible in every field the gate reported. An exemption now only excuses a genuinely
   absent page; one naming a present page is surfaced as `stale_exemptions`.
2. **Missing/extra were named by a POSITIONAL tail slice.** With candidates `[A, B, C]` and only `A`
   absent, the slice named `C` — so an exemption signed for `C` excused `A`, and one signed for `A` did
   nothing. Both sides are now matched by **name** against the rendered pages. The verdict is still the
   rename-robust **count**; names decide only which pages are reported, and therefore which exemptions
   apply.

## Mutation campaign — now committed, and it immediately found 3 survivors in my own tests

`tests/mutate_check_unit.py`, runnable, exit 0 = every mutation caught. **29 mutations**, each naming
the single anchor test it must kill and run against that anchor alone, plus a no-op **negative
control** that must SURVIVE. A mutation counts as CAUGHT only on a genuine pytest `N failed`; a
collection or import error is reported **BROKEN**, never caught.

Committing it paid for itself on the first run — **3 survivors, one per vacuity mode**:

| survivor | mode | fix |
|---|---|---|
| `visual count is fabricated` (`return 1`) | 2 — assertion in a branch the fixture never enters: the fixture had no `visuals/` folder, so the `is_dir()` guard answered first and the counting line was unreachable | fixture now creates an **empty** `visuals/` folder |
| `placeholder identity drops the id-prefix clause` | 3 — masked by a sibling: the anchor page had a visual, so the zero-visual clause already refused it | anchor is now a zero-visual page with an ordinary id and the placeholder's display name, where only the id decides |
| `blank pages no longer fail the gate` | 4 — passed for two reasons: the anchor was also missing a page | new anchor makes a blank page the **only** problem, missing and extra both zero |

A fourth survivor appeared when I added the attribution mutations — the positional mutation survived
its first anchor because that fixture's exemption made both attributions agree. It now has a dedicated
anchor with no exemption involved.

**29/29 caught, control SURVIVED, exit 0.**

## 4. MEDIUM — branch currency

Rebased onto `origin/master` (`58f83aa`). `git rev-list --count HEAD..origin/master` → **0**. All
gates below were run after the rebase.

## Gates (re-run on the integrated tree)

| gate | result |
|---|---|
| `ruff format` / `ruff check` | clean |
| `pylint scripts/check_unit.py` | **exit 0** (10.00/10) |
| `pylint tests/mutate_check_unit.py` | **exit 0** (10.00/10) |
| `pylint tests/test_check_unit.py` | exit 22, finding set **identical to master's baseline** (import-error ×4, too-many-lines, unused-argument ×2) — no new findings |
| `tests/mutate_check_unit.py` | **exit 0** — 29/29 caught, negative control survived |
| `pytest tests/` | see below |
| `check_navigation_index.py` / `sync_agent_conventions.py --check` / `check_agent_capabilities.py` | exit 0 |

**On the suite count — you were right and I was sloppy.** I reported "3576 passed" from a pre-rebase
run and quoted it as if it were the number for this branch. Measured on the integrated tree just now:

```
3 failed, 3591 passed, 6 skipped
```

The 3 are the known `test_upstream_repro_pins.py` engine-drift tripwires, unchanged from master.

## Estate result, restated honestly

```
44 workbooks: 42 PASS, 2 not PASS
absent candidates: 23   explained=21   unexplained=2
blank pages: 0          engine crash-guard placeholders: 2
handover slices that did NOT bind to their unit: none
```

## Still not verified

- **Renaming defeats drop attribution by construction.** If one candidate is renamed *and* another
  vanishes unexplained, the counts can balance and the gate passes. Pre-existing ceiling; the verdict
  is count-based on purpose, and the docstring says so.
- **`tier == "empty"` is trusted as the engine's own statement.** I did not verify the engine never
  emits a page for a tier-`empty` worksheet; on this estate it never did (21/21), but that is one
  estate, not a guarantee.
- **The estate validation stages units synthetically** — real engine spec, PBIR and handover copied
  into a unit-shaped folder. The inputs are genuine; the layout is constructed.
- **A unit with orphan worksheets and no handover now fails parity.** None of the 16 committed
  `examples/*` units has a handover directory and 6 have dropped orphans; they are not run through
  this gate in CI, so nothing regressed, but an operator re-checking an old unit will see it.
- **Not run in Power BI Desktop.** This is a static gate.

Guust-Franssens:
# Round 3 — all three fixed once, on the sibling's identity type. And the 42 is retracted too.

You were right that 42/44 was hollow. **I reproduced your number exactly**: the 9 previously-passing
workbooks omit precisely **18** expected pages. Under the corrected gate:

```
44 workbooks: 33 PASS, 11 not PASS
omissions by disposition:   declared-by-engine-unsigned 21    unexplained 2
blank pages: 0   unaccounted extra pages: 0   ambiguous-attribution units: 0
```

11 = your 9 + the 2 that were already failing. 23 absent candidates = your 18 + those 2 workbooks'
5. **Every one is now reported and blocks**, and each carries the sentence that resolves it.

## The common cause, fixed once

I looked at `scripts/object_identity.py` before writing anything, and it fits. **It is now in this
branch byte-identical** — `git diff FETCH_HEAD -- scripts/object_identity.py tests/test_object_identity.py`
is empty — together with its own tests and its two registration rows in `scripts/README.md` and
`docs/agent-capability-wiring.md` (also verbatim; `check_agent_capabilities.py` fails without them).
I did not modify it and did not re-implement it, so whichever of #428 / #433 merges first the other
rebases clean.

⚠️ **Your caveat holds, and I verified it rather than trusting it.** Both "cannot" claims are still
holes at `037ec64`:

```
ObjectIdentity(kind='unknown', name='x')  ->  ObjectIdentity(kind='unknown', name='x')   # from_engine returns None
Resolution(outcome='ambiguous')  bool() -> True
Resolution(outcome='absent')     bool() -> True
```

I do not own that file, so I did not patch it — **this is a finding for #428**. What I did instead is
consume it so neither hole can reach a verdict here: identities are built only through
`ObjectIdentity.from_engine`, and a resolution is read via `.outcome` **and** `.value()` inside a
`try`, never truthiness. A mutation (`EVIDENCE: an ambiguous resolution is read as its first match`)
pins that.

## 1. HIGH — `tier="empty"` diagnoses; it does not approve

**Reproduced:** parity `PASS`, oracle `PASS`, `compromises=0`, page B simply gone.

**Fixed.** Every expected page is now either rebuilt or carries a **signed** `page-parity` exemption.
An engine-declared omission is reported with its proof and disposition
`declared-by-engine-unsigned`, and it **fails**. The same rule governs the oracle denominator: only a
signed omission is removed, so a unit missing a page can no longer report validation-grade coverage
of what it still has. Only `applied` signatures count as compromises.

**I deliberately did NOT build the "genuinely empty source sheets" carve-out you left open.** I
measured it first: across all 44 workbooks, **zero** worksheets have empty encodings — including the
one the engine itself calls *"empty worksheet (no fields on any shelf or encoding)"*. The branch would
have been unkillable dead code with a fail-open flavour, so it is better absent. If a real case turns
up it belongs on source structure, exactly as you said.

## 2. HIGH — kind binding

**Reproduced:** dashboard `Sales` absent, its placed worksheet `Sales` carrying the empty-tier row →
the worksheet's evidence explained the dashboard.

**Fixed by identity, not by another key.** Candidates carry `kind`; evidence is indexed as
`ObjectIdentity(kind, exact_name)` in an index built `normalized=False`, so an engine-to-engine join
has no lossy table to slip into.

**A measurement that made the rule sharper than "add kind".** `tier == "empty"` occurs **only** with
`visual_type == "unsupported"` — 46/46 rows across the estate — because `_fidelity_tier` returns
`empty` iff `visual_type in (None, "unsupported")`, and `_viz_fidelity` gives a real visual type only
to rows built from `ir["worksheets"]`. So **an empty-tier row can only ever denote a worksheet**, and
proof now requires **both** fields. That second clause is not redundant: its killing anchor is a real
estate shape — a `filter`-scope row named `Location` sitting beside a worksheet also named `Location`
(HR Dashboard) — where only `visual_type` separates them.

## 3. HIGH — rename ambiguity

**Reproduced:** expected `A, B, C`; rendered a renamed `A` plus `B`; signing for `A` turned
`PRECONDITION_FAILED` into `PASS` with no stale exemption reported.

**Fixed.** While any rendered page matches no expected page, that page might BE a renamed candidate,
so **no** absent candidate is treated as genuinely absent and **no** name-only signature is applied —
on either the parity or the oracle side. The explicit way out is an `extra:<page>` signature
accounting for the unmatched page; once given, signatures apply again **and `C` is still reported**.
Exemptions are dispositioned `applied` / `stale` / `ambiguous`, and only `applied` counts as a
compromise — your second observation, `1 accepted` / `compromises=1` for a signature that did
nothing, now reads `compromises=0`.

The **positional attribution is gone entirely** — missing and extra are identified by name — so the
count-based verdict it existed to support is gone with it.

## Mutation campaign — 36, and rebuilding it found 4 more survivors

Rebuilding for the new model surfaced four survivors and one dead anchor, every one a weakness in
tests I had just written:

| survivor | why | fix |
|---|---|---|
| `oracle denominator drops declared omissions` | its anchor snippet had been reflowed onto one line by ruff, so it mutated **nothing** | re-anchored on the current text |
| `evidence identity drops the kind` | a **no-op I wrote by mistake** (a walrus that still produced `KIND_WORKSHEET`) | replaced with one that indexes the row under **both** kinds — the actual defect |
| `visual_type is no longer required` | vacuity mode 3 — the kind guard already refused its dashboard fixture, masking the clause | new anchor is the `filter`/`Location` estate shape, where only `visual_type` decides |
| `rename ambiguity no longer suspends a signature` | the guard exists on **both** sides; the anchor only exercised parity | split into two mutations with two anchors, one per side |

**36/36 caught, negative control SURVIVED, `tests/mutate_check_unit.py` exits 0.**

## Gates, measured on the pushed HEAD (`bcf2c70`)

| gate | result |
|---|---|
| `ruff format` / `ruff check` | clean |
| `pylint scripts/check_unit.py` | **exit 0** (10.00/10) |
| `pylint tests/mutate_check_unit.py` | **exit 0** (10.00/10) |
| `tests/mutate_check_unit.py` | **exit 0** — 36/36, control survived |
| `pytest tests/` | **3 failed, 3609 passed, 6 skipped** — the 3 known `test_upstream_repro_pins.py` tripwires |
| `check_navigation_index.py` / `check_agent_capabilities.py` / `sync_agent_conventions.py --check` | exit 0 |
| `git rev-list --count HEAD..origin/master` | **0** |

**On the count discrepancy.** I cannot reconcile your 3590 against my 3591 — both trees have moved
since, and I did not keep the artifact. What I can give you is a *deterministic* anchor instead of a
run-dependent one: on `bcf2c70`, `pytest --collect-only` reports **3618 collected**, of which 3 fail
and 6 skip. If your run disagrees with 3618 collected, that is an environment difference worth
chasing; if it agrees, the pass count follows. I will quote collected-count plus failures from here on.

## Still not verified

- **`object_identity`'s guarantees are not enforced** (above). This gate is written to be immune, but
  a future join in *this* file could still use the raw constructor. #428's
  `check_identity_normalization.py` is the structural answer and is not on this branch.
- **A same-name collision between two CANDIDATES** (two dashboards, or a dashboard and an orphan
  worksheet, sharing a name) is not specifically detected; the page match would find two claimants for
  one page. Zero occurrences across 44 workbooks, so untested against reality.
- **`tier == "empty"` is trusted as the engine's own statement.** I did not verify the engine never
  emits a page for such a worksheet; on this estate it never did (21/21).
- **The estate staging is synthetic** — real engine spec, PBIR and handover copied into a unit-shaped
  folder.
- **The gate is now materially stricter and no committed `examples/*` unit has been re-signed.** 11 of
  44 workbooks fail where 2 did. That is the point, but it is real operator work that this PR does not
  do.
- **Not run in Power BI Desktop.**

Guust-Franssens:
# Round 4 — all four fixed. The shared type is now gated, and it caught drift on its first run.

You were right on every count, including the one I was most confident about.

## 1. HIGH — byte-identity was false, and the branches were API-incompatible

Confirmed by reproduction: `AttributeError: module 'object_identity' has no attribute 'IdentityIndex'`.
I carried the round-3 module; the sibling had moved to `EngineIndex`/`CandidateIndex`,
`__post_init__` validation and a raising `__bool__` — **the two holes I had found at `037ec64` and
reported for #428**. So "either PR can merge first" was false and nothing would have said so.

**Both files re-taken from the sibling's current head, and it is now a gate in two halves** — because
`actions/checkout@v4` makes a shallow clone, so no other ref exists in CI and a ref-diff cannot run
there:

| gate | what it proves | where it runs |
|---|---|---|
| `tests/test_shared_identity_pin.py` | this branch has not edited a file it does not own (SHA-256 pin, offline) | everywhere, including CI |
| `tests/verify_shared_identity_pin.py` | the copy equals the sibling's, and **exits 2 when it cannot compare** | wherever the ref resolves |

⚠️ **It paid for itself on the first run.** It reported `DRIFTED` — the sibling had moved again from
`2408e88` to `cc70833` *while I was working*. Re-taken; now:

```
IDENTICAL to origin/feat/reference-readiness-gate@cc70833d: scripts/object_identity.py, tests/test_object_identity.py
python tests/verify_shared_identity_pin.py --ref origin/master   ->  exit 2, CANNOT ESTABLISH
```

The second line is deliberate: master does not carry the module yet, and "I could not look" must
never read as "it is fine".

## 2. HIGH — the type guarded one join; three used raw names

Both reproduced: one rendered `Sales` page satisfied a dashboard **and** a worksheet (PASS,
`expected_count=2`, `emitted_count=1`, one oracle row counted as 2-of-2); one exemption named `Sales`
signed both omissions for one compromise.

**All four joins now resolve through the same `EngineIndex`.** The kind-less side — a PBIR display
name, an exemption `item`, an oracle manifest entry — meets identity in exactly **one** documented
adapter, `_claim()`, which asks the index for every identifiable kind and refuses as soon as more than
one expected page answers. `Resolution.value()` stays the only reader; nothing reads truthiness (the
current type raises on that).

After: `contested_names=['Sales']`, both objects reported as omissions, oracle `visual_present=0`,
`applied_exemptions=[]`, `compromises=0`. **Signing the page `id` is the way out** — an id is
per-object, so it accepts exactly one of the two, and that is its own test.

## 3. MEDIUM — resolve before de-duplicate

Confirmed at `_handover_workbooks`: `_unit_dir()` resolves, `target / "handover"` does not, and
de-duplication compared the two *before* resolution — so the documented relative CLI shape scanned the
same directory twice, indexed every evidence row twice, and turned each into an AMBIGUOUS resolution
whose declared reason vanished. Roots are now resolved first.

⚠️ **You were right that my headline rested on it.** This round's estate figure is therefore measured
through the **relative** invocation, deliberately.

## 4. MEDIUM — my "zero source-empty worksheets" was false

It was, and the cause is embarrassing in a useful way: **I ran the scan against a `_specs/` directory
I had already deleted**, so it found no specs at all and reported 0. Both sheets you name have every
shelf empty (`rows/columns/label/detail/tooltip: []`, `color/size/shape: null`) and no filters.

Implemented as your first option — a `no-source-content` disposition derived from the **spec**, never
from an engine tier. A sheet that renders blank in Tableau owes no Power BI page, so it is reported
and does not block. Three guards, each with its own test and mutation: a missing `encodings` key is
**not** empty (unknown is not empty), one populated shelf is content, and a sheet that still carries a
filter is content.

## Estate, measured through the relative CLI shape

```
44 workbooks: 35 PASS, 9 not PASS
omissions:  declared-by-engine-unsigned 19    unexplained 2    no-source-content 2
blank pages: 0   unaccounted extra pages: 0   contested names: none
```

35 rather than 33 because the two source-empty sheets are no longer rebuild gaps. The 19 declared
omissions still block, unsigned, which is the round-2 rule holding.

## Mutation campaign — 45, and re-anchoring was itself a finding

Re-anchoring after the rewrite needed **11 anchors re-pointed**. The harness reported
`ANCHOR-SNIPPET x0` for each rather than scoring them caught — the failure mode it exists to refuse,
working. Nine new mutations cover pairing through the index, oracle evidence on a contested name,
bare-name vs id signature resolution, resolve-before-dedup, and all four halves of the source-empty
rule.

**45/45 caught, negative control SURVIVED, exit 0.**

## Gates on pushed HEAD (`4f96b35`)

| gate | result |
|---|---|
| `ruff format` / `ruff check` | clean |
| `pylint scripts/check_unit.py` + both new test modules | **exit 0** (10.00/10) |
| `tests/verify_shared_identity_pin.py` | **exit 0** — IDENTICAL to `cc70833d` |
| `tests/mutate_check_unit.py` | **exit 0** — 45/45, control survived |
| `pytest tests/` | **3642 collected, 3 failed** — the three documented `test_upstream_repro_pins.py` tripwires |
| `check_navigation_index.py` / `check_agent_capabilities.py` / `sync_agent_conventions.py --check` | exit 0 |
| `git rev-list --count HEAD..origin/master` | **0** |

Quoting **collected + failures** only, as instructed. Passed-count is dropped.

## Still not verified

- **The pin will need updating whenever #428 moves again.** That is the design — the verifier fails
  rather than drifting — but it is manual, and this branch cannot make CI fetch another ref.
- **`check_identity_normalization.py`** (the sibling's rule banning `oid.normalize()` outside the
  module) is **not** on this branch. `check_unit` does not call it, but nothing here enforces that.
  The lossy `_slug()` still keys the oracle-evidence side, which is external-producer names and so
  legitimate — but it is our own lossy function, outside that quarantine.
- **A same-name collision between two candidates is now detected; a same-name collision between two
  emitted PAGES is treated the same way** (contested), which I have tested but never seen in the
  estate — zero occurrences in 44 workbooks.
- **`no-source-content` is derived from our parser's spec**, so it inherits any parser gap in reading
  encodings. It is refused when `encodings` is absent, which is the safe direction.
- **The 19 declared omissions are real operator work this PR does not do.** No committed
  `examples/*` unit has been re-signed.
- **Not run in Power BI Desktop.**

Guust-Franssens:
# Round 4 addendum — the gate failed its own first CI run, and that is a finding worth having

I posted the round-4 response before CI finished. It went **red**, on the byte-identity gate itself,
against files nobody had touched:

```
scripts/object_identity.py       CI d5971e00…   local 58a8e6d3…
tests/test_object_identity.py    CI 7e0856ac…   local 13ed4012…
```

**Cause:** `git checkout` applies `core.autocrlf`, so the same blob is CRLF in a Windows working tree
and LF on the Linux runner. A raw byte hash therefore pinned the **platform**, not the content — a
gate that fails for a reason unrelated to what it is guarding, which is the same class of uselessness
as one that passes for the wrong reason.

**Fixed:** the digest normalises `\r\n` → `\n` before hashing, so it stays sensitive to every real
edit and survives the checkout git performs on our behalf. A test pins that property directly — CRLF
and LF of the same content digest equal; a one-character edit does not. Rebased onto the moved master
and re-pushed.

```
checks         pass   4m53s
windows-bundle pass   2m5s
git rev-list --count HEAD..origin/master  ->  0
python tests/verify_shared_identity_pin.py -> exit 0, IDENTICAL to cc70833d
```

## ⚠️ And it explains the count discrepancies — including the two you flagged

The same CI run reports **4063 passed / 90 skipped** where this Windows machine reports **3642
collected**. The suite is genuinely platform-dependent: ~90 tests skip on Linux and others do not
exist there at all.

So my "quote collected + failures" fix was only half right. **A collected- or passed-count is
comparable only within one platform.** From here I will quote it as
`<count> on <platform>, N failed`, and I would rather you treat the *failure list* as the anchor —
that has been stable and identical every round: the three documented `test_upstream_repro_pins.py`
tripwires.

For this HEAD (`292e58e`):

| where | collected | failed |
|---|---|---|
| Windows (local) | 3642 | 3 tripwires |
| Linux (CI) | 4153 (4063 passed + 90 skipped) | **0** |

Nothing else changed from the round-4 response above.

Guust-Franssens:
# Round 5 — all four fixed, and my own cross-check found a fifth

## 4 (first, because of what it is) — cannot-establish vs an answer

Confirmed and fixed. `git diff --quiet` exits **1** for a difference and **>1** for a comparison
*failure*, and I treated every nonzero as drift — so a simulated exit 128 returned **DRIFTED**, a
confident verdict from a run that had established nothing. In the instrument written to prevent
exactly that. Now `0` identical, `1` drifted, **anything above 1 CANNOT ESTABLISH**, with the git
stderr carried through.

## 1 — signature matching bypassed exact identity

Both reproductions confirmed: `A-B` signed `A B` (slug), and two pages sharing `ws.same` were both
signed by one entry.

Every page-parity item is now resolved **exactly once, by exact id then exact name**, never through
`_slug` — including the `extra:` form, which had the same hole. And `_spec_pages` **refuses a spec
that reuses a page id**: an id is the only way to sign one of two objects that share a name, so a
colliding id makes every signature unattributable and the expected set ungradeable.

## 2 — `no-source-content` accepted unknown or visible content

All three confirmed. The predicate checked only the truthiness of *present* keys, and because the
disposition needs no signature every false positive was a silent PASS.

Proof now requires **complete** structure, and — following your warning about open enumerations —
**the enumeration is read from the committed `docs/migration-spec.schema.json`, not written in the
gate**:

- the worksheet must carry exactly the schema's key set (all eleven), and
- `encodings` exactly the schema's eight channels, and
- every content-bearing key falsy (`title_text` included — that is the visible channel I had missed), and
- `migration_spec_version` must be the version the classification was derived against.

A property added to the schema upstream lands in the **content** set by default, so the rule tightens
rather than opening a hole. Both real estate sheets carry all eleven properties, so completeness costs
nothing on genuine parser output.

## 3 — parity and oracle disagreed, so 35/44 was half a verdict

Confirmed. The oracle denominator now reads the **same** `_disposition_omissions` output that page
parity does, so `{accepted-signed, no-source-content}` is one set and the two halves cannot drift
apart again.

## ⚠️ And that cross-check found a fifth thing, in my own code

I added a both-halves check to the estate harness rather than asserting agreement. It reported
**one workbook disagreeing** — `Meridian Multi-Source (3 systems)`. Cause: when a source-empty page is
the *only* expected page, the denominator empties legitimately and the not-assessable early return
reported `excluded_omissions: []`, so a page **both halves had accepted** read as a disagreement. The
list is now carried through that path. Found by measurement, not by a test; now it has both.

```
44 workbooks: page-parity 35 PASS, 9 not PASS
omissions:  declared-by-engine-unsigned 19    unexplained 2    no-source-content 2
workbooks where parity and oracle disagree about a source-empty page: none
blank pages: 0   unaccounted extra pages: 0   contested names: none
```

Measured through the **relative** CLI shape, and this time reporting **both** halves.

## Campaign — 55, and re-anchoring found 2 more masked clauses

11 anchors needed re-pointing after the rewrite (reported `ANCHOR-SNIPPET x0`, not scored as caught).
Two then **survived**, both masked by a sibling clause: a non-dict `encodings` that the key-set check
already refused, and a partial `encodings` that the worksheet key-set check already refused. Each now
has a fixture where only its own clause can decide — a list carrying exactly the eight channel *names*
for the type guard, and a complete worksheet with partial shelves for the key-set guard.

**55/55 caught, negative control SURVIVED, exit 0.**

## Gates, from the tip I pushed (`efb956d`)

| gate | result |
|---|---|
| `ruff format` / `ruff check` | clean |
| `pylint scripts/check_unit.py` + test modules | **exit 0** — ⚠️ an intermediate commit printed 9.98/10 while exiting **16** (three `line-too-long`); caught by exit code, fixed, re-run |
| `tests/verify_shared_identity_pin.py` | **exit 0** — IDENTICAL to `cc70833d` |
| `tests/mutate_check_unit.py` | **exit 0** — 55/55, control survived |
| `pytest tests/` | **3845 collected on Windows, 3 failed** — the three documented tripwires |
| `check_navigation_index.py` / `check_agent_capabilities.py` / `sync_agent_conventions.py --check` | exit 0 |
| `git rev-list --count HEAD..origin/master` | **0** |

Counts quoted from the pushed tip and labelled by platform, as instructed. CI's Linux figure will
differ again (~90 tests skip there and others do not exist); the stable anchor remains the failure
list, which has been the same three every round.

## Still not verified

- **`_slug` still keys the oracle-evidence side.** That side is external producer names, so a lossy
  compare is legitimate there — but it is our own lossy function, outside the sibling's
  `check_identity_normalization.py` quarantine, which is not on this branch.
- **The source-empty rule trusts our parser to populate the schema's keys.** If the parser stopped
  emitting one, every worksheet would become unclassifiable — the safe direction, but it would be
  silent apart from omissions reverting to rebuild gaps.
- **`migration_spec_version` is compared to a single constant.** A future `1.1` would disable the
  classification until someone re-derives it. Deliberate; also unexercised against a real 1.1.
- **The pin needs a human whenever #428 moves.** The verifier fails rather than drifting, but this
  branch cannot make CI fetch another ref.
- **19 declared omissions remain real operator work** this PR does not do, and no committed
  `examples/*` unit has been re-signed.
- **Not run in Power BI Desktop.**

Guust-Franssens:
# Round 6 — both fixed, and the enumeration you asked for found two more of the same class

Your closing instruction was the actual task: *"enumerate every place an identity becomes a `str` or
enters a `set`."* I did that mechanically over `check_unit.py` rather than by reading, and it found
**two flattenings you had not reported**, plus a third turned up while fixing yours.

## 1. HIGH — resolved per PAGE, not once GLOBALLY

Both reproduced: `collide` as page A's **id** and page B's **name** signed both (PASS, one
compromise); one `extra:Bonus` accepted **two** rendered pages called `Bonus`.

`_resolve_signatures` now resolves each raw item **once against the whole unit**, counted by
**distinct object** across the id and name namespaces together. So an item matching two objects is
contested and applies to none — while a page whose id happens to equal its **own** name is still one
object and still applies (its own test, so the dedupe cannot be dropped).

After: `applied=[]`, `unsigned=['A','collide']`, `compromises=0`; `unaccounted=['Bonus','Bonus']`.

## 2. MEDIUM — exclusions flattened to a name set

Reproduced exactly: a **signed** dashboard `Sales` removed the **unsigned** worksheet `Sales` too,
`pages=0` while listing one exclusion. Removal is now by `_page_key` — `(kind, id, name)` — never by
display name. After: `oracle pages=1`, exclusions `['d.sales']`, and `ws.sales` still owes evidence.

⚠️ **And the cross-check is extended as you required.** It now compares the **denominator actually
graded** (`oracle["rows"]`) against what parity says still owes output — not the reported exclusion
list, which is precisely what made this defect invisible to the round-4 version of the same
instrument. **0 disagreements across 44 workbooks.**

## Found by the enumeration, not by the review

| line | flattening | consequence |
|---|---|---|
| `_unapplied_exemptions` | omitted pages compared by **display name** | a **present** page whose name matched an omitted one was never reported stale |
| `_unit_workbook_keys` / binding | workbook bound by **pure slug** | two workbooks whose names slug alike bound **interchangeably** — the borrowed-evidence defect one level above the page |

Both fixed and pinned. The workbook fallback now binds on an exact stem, and lossily **only when that
lossy match is unique**.

## Found by the round-5 reproduction, while fixing 2

Page parity subtracted accounted-for `extra:` pages before deciding ambiguity; the oracle did not. The
same unit was ambiguous in one half and not the other, so a signed omission left one denominator and
not the other. `_attribution_ambiguous` is now **one predicate for both**, and
`OMISSIONS_OWING_NOTHING` names the shared disposition set **once** so the two halves cannot re-derive
it differently.

## The enumeration, and what remains lossy on purpose

Every surviving `_slug` call, classified:

| site | verdict |
|---|---|
| `_exempted` (stub-measures, scaffold-partitions) | their own item vocabularies; page-parity no longer uses it |
| `described` map / `why_unexplained` | **reporting text only**, never a verdict |
| oracle manifest names (`_reference_oracles`, `_oracle_capture_oracles`, `_oracle_row`) | **external producer names**, where a lossy compare is legitimate — and now guarded by the `contested` check so it can never satisfy two candidates |
| workbook binding fallback | lossy, but **only when unique** |

That is the honest list. `_slug` is still ours rather than the sibling's quarantined `normalize`, and
`check_identity_normalization.py` is not on this branch — stated again below.

## Estate, relative invocation, both halves

```
44 workbooks: page-parity 35 PASS, 9 not PASS
omissions:  declared-by-engine-unsigned 19   unexplained 2   no-source-content 2
CROSS-CHECK - oracle DENOMINATOR vs what parity says still owes output: 0 disagreements
blank pages: 0   unaccounted extra pages: 0   contested names: none
```

## Campaign — 63

9 anchors re-pointed after the rewrite (`ANCHOR-SNIPPET x0`, not scored as caught). Two then survived:
a **no-op mutation I had written** (a duplicated `or` clause that cannot double-count in a generator),
and a fixture whose two workbook names did **not** actually collide on the slug being tested. Both
replaced with mutations and fixtures where only the clause under test can decide.

**63/63 caught, negative control SURVIVED, exit 0.**

## Gates, from the tip I pushed (`c2eb4b8`)

| gate | result |
|---|---|
| `ruff format` / `ruff check` | clean |
| `pylint` — `check_unit.py` + all three test modules | **exit 0** ⚠️ two intermediate runs printed `9.56/10` and `8.98/10` while exiting **16** (`line-too-long`); caught by exit code both times |
| `tests/verify_shared_identity_pin.py` | **exit 0** — IDENTICAL to `cc70833d` |
| `tests/mutate_check_unit.py` | **exit 0** — 63/63, control survived |
| `pytest tests/` | **3853 collected on Windows, 3 failed** — the three documented tripwires |
| nav / capability / convention-sync | exit 0 |
| `git rev-list --count HEAD..origin/master` | **0** |

## Still not verified

- **`_slug` remains our own lossy function**, outside the sibling's `check_identity_normalization.py`
  quarantine, which is not on this branch. The classification above is a hand audit; nothing enforces
  it.
- **The `(kind, id, name)` key assumes ids are unique**, which `_spec_pages` enforces — but only
  within one spec. A unit with two specs is not a shape I tested.
- **The workbook-binding fallback is uniqueness-guarded, not exact.** A sanitised folder name still
  binds lossily; I kept that because the estate depends on it, and it is now visible in
  `bound_workbooks`.
- **19 declared omissions remain operator work** this PR does not do; no committed `examples/*` unit
  has been re-signed.
- **Not run in Power BI Desktop.**

Guust-Franssens:
## Round 6 response — all three fixed, plus two I found while fixing, plus the gate you asked for

All three reproduced first (script `_repro_round6.py`, since cleaned up), all three now refuse. Tip: `930c1c8`.

### The framing first

> *"an enumeration is a hypothesis until each entry is tested"*

That is the finding. The audit made three wrong claims **and** made them checkable — and the answer is not a better-written audit, it is a gate. `tests/test_slug_call_site_census.py` is that gate, described at the bottom.

---

### 1. HIGH — oracle/reference evidence flattened by slug, workbook identity discarded ✅ fixed

**Reproduced both halves.** `A B` gave validation-grade PASS to `A-B`; a `Revenue` record from `"Different Workbook"` gave full visual+numeric PASS.

Records are now `OracleRecord(name, workbook, visual, numeric)` — multiplicity, exact spelling and producing workbook intact — resolved through `OracleEvidence`:

| guard | rule |
|---|---|
| workbook | a record DECLARING a workbook this unit does not ship is discarded (`foreign_workbook_evidence`); one declaring none is admitted, counted in `unattributed_evidence`, and **named in the grade** |
| exact | an exact name match wins outright |
| normalized | applies only when the lossy key resolves to **exactly one producer record AND exactly one expected page** |

⚠️ **Deviation you should judge, stated plainly: I did not route these through `CandidateIndex`.** Oracle records carry **no declared kind** today (PR #422 is adding one). `CandidateIndex` would therefore either match nothing or force me to assert a kind the record does not have — inventing identity to satisfy a type is worse than the flattening. I implemented your *rule* directly instead. If #422 lands a declared kind, this should convert.

**Real oracle manifests carry no `workbook` field**, so refusing unattributed evidence outright would reject every capture in the estate. Admitted-but-disclosed is the compromise; `test_oracle_evidence_declaring_no_workbook_is_admitted_but_flagged` pins it.

### 2. HIGH — `_exempted` decided per finding with no multiplicity check ✅ fixed

**Reproduced exactly**: one signature `A-B`, two exemptions, gate flipped to PASS.

`_exempted` is **deleted**. `resolve_exemptions(entries, check, findings)` resolves each raw entry **once, globally**: exact name/alias first, lossy slug only as fallback, applied only when it identifies **exactly one** finding. Anything else is `contested_exemptions` and reported. Same for `stub-measures`.

### 3. MEDIUM — uniqueness guard on one side only ✅ fixed

**Reproduced**: `Bo ok.Report` + `Bo-ok.Report` → one key `book`, handover `Book` accepted.

`_unit_workbook_keys` now returns the slugged side as a **list**, and `_bindable_workbooks` requires `slugs.count(slug) == 1 and slugged_stems.count(slug) == 1`.

⚠️ **My first version of this fix was wrong and the estate caught it.** Counting artifact FILES made every ordinary unit look collided — a report and its model share a stem — and drop-evidence binding died estate-wide: **19 declared → 21 unexplained**. The key is one entry per **distinct stem**. `test_a_report_and_its_model_sharing_a_stem_are_one_owner_not_a_collision` pins it, and a mutation (`the target artifact keys count files instead of distinct stems`) kills it.

---

### Two more, found while fixing, both self-reported

**A. `_reference_dirs` resolved candidates but never deduplicated them.** One manifest was read twice, producing two records per dashboard, which then refused each other as *"2 producer records are named X"*. The round-3 resolve-before-dedup fix had reached handover and oracle discovery but **not reference discovery** — invisible for exactly as long as multiplicity was being collapsed. Both now share `_resolved_unique`. Two mutations: no dedup at all, and dedup-before-resolve (killed by a **relative-path** test, since only the relative shape distinguishes them).

**B. My own mutation harness gave a confident wrong answer.** It printed *"negative control DID NOT behave as required"* when the control **had** survived and the real problem was two stale anchor snippets. That is tonight's defect inside the instrument written to prove tests can fail — the same shape as your round-4 finding 4. The three states are now reported separately, and an `--only` selection with no control reports **UNVALIDATED**, not failure.

---

### The structural ask: `tests/test_slug_call_site_census.py`

Machine-checked, scoped to **call sites of one function** (not "any lossy comparison" — I took the quarantine-rule lesson).

Every `_slug()` call is enumerated by AST and must be pinned with a classification:

| class | mechanical proof |
|---|---|
| `uniqueness-guarded` | the enclosing function **must** compare something against the literal `1` |
| `multiplicity-preserving` | no `_slug` call in that function may sit inside a set literal/comprehension |
| `reporting-only` | none exists — held by the **pin**: the exact source line is recorded, so editing it fails the gate and forces re-adjudication |

Four failure modes: an unpinned site, a stale row, a **call-count** mismatch (a second `_slug` on a pinned line), and a disproved classification.

⚠️ **The multiplicity check was vacuous in my first draft** and I only found it by asking "what mutation kills this?" — keyed by *line*, collapsing the list to a set also rewrote the line, so the mutated site stopped matching the pin and the detector went silent exactly when it should have fired. It is now keyed by **function**. That is vacuity mode 4, in a test written this round.

All six detectors are pure functions over `(calls, guards, census)` with synthetic **negative** tests, so each is shown to fire; the census currently pins **11 call sites in 7 functions**.

---

### Mutation campaign — 85/85, and it found six of my own vacuous tests

First run of the new mutations: **6 SURVIVED, 1 stale anchor.** Each was a real vacuity, and every one is now fixed rather than re-pointed away:

| survivor | why it was vacuous |
|---|---|
| one entry may exempt every match | anchor's exact match disambiguated first — mode 4 |
| slug fallback runs despite an exact match | anchor never had an exact match — mode 2 |
| exact matching removed | same |
| normalized match ignores the producer side | anchor refused earlier on the exact branch — mode 2 |
| workbook guard rejects own evidence | mutation was a no-op; the slug fallback rescued it |
| dedup-before-resolve | tmp_path is already resolved — the ordering was unobservable |

Fixed by adding the missing cases (`test_two_records_sharing_a_normalized_key_satisfy_nothing`, `test_a_relative_target_reads_one_reference_directory_once`) and re-pointing anchors at scenarios that actually enter the branch.

**Per your instruction, I broke the production code for every new behaviour and confirmed a specific test fails**, each mutation against its intended anchor alone. 85/85 caught, negative control SURVIVED, exit 0. The census gate's five mutations are in the same campaign (both suites are now collected so `-k` can reach them).

---

### Counts — measured at the tip I pushed, with the command

```
$ python -m pytest -q                    # Windows
3 failed, 4185 passed, 9 skipped   (4197 collected)
```

The three failures are the documented `test_upstream_repro_pins.py` tripwires — nothing else.

⚠️ **This does not match your 3853, and the difference is the invocation, not the tip.** `pytest -q` uses `testpaths = ["tests", ".github/skills"]`; your figure looks like `pytest tests/`. Controlled check, same machine, same command: **HEAD (`c2eb4b8`) collects 4173, this tip collects 4197** — exactly the +24 tests I added. So the delta is mine and accounted for; the absolute number is invocation-dependent. Quoting **collected + failures + the command**, as instructed.

| gate | result |
|---|---|
| `ruff format` / `ruff check scripts tests conftest.py` | exit 0 |
| `pylint scripts` | **exit 0** — it printed **10.00/10 while exiting 8** on two R0914s; caught only by the exit code, as instructed |
| `pytest -q` | 4197 collected, 3 failed (documented tripwires) |
| `tests/mutate_check_unit.py` | 85/85, control SURVIVED, exit 0 |
| `verify_shared_identity_pin.py` | IDENTICAL to `origin/feat/reference-readiness-gate@cc70833d`, exit 0 |
| navigation / capability / convention | exit 0 |

### Estate — 51 units, relative CLI paths

```
page-parity: 35 PASS / 9 PRECONDITION_FAILED / 7 NOT_CHECKED
omissions:   19 declared-by-engine-unsigned, 2 unexplained, 2 no-source-content
blank=0  unaccounted_extra=0  contested=0
parity/oracle disagreements: 0
```

Reproduces your 35/44 exactly (44 assessable + 7 datasource-only units that are NOT_CHECKED).

⚠️ **And the honest part: this is byte-identical to the committed baseline.** I ran the same staging against `git show HEAD:scripts/check_unit.py` — same 35/9/7, same 19/2/2, same zero disagreements. **None of this round's fixes changed a single estate verdict.** They close paths this estate does not exercise: it has no punctuation-colliding artifact names, no duplicate-name oracle records, and no cross-workbook oracle manifests. I would rather say that than let the unchanged headline imply the fixes were load-bearing here.

⚠️ **A staging defect of mine, disclosed:** my first re-measurement gave **18 declared / 3 unexplained** because `run_estate` sanitises handover FILENAMES (`Section 12 - Row Level Calculations (Functions)` → `..._Functions_.json`) and my harness looked up the exact name, silently skipping real evidence. Fixed by indexing handovers on the same lossy key. The 19/2 split is the correct one, and the baseline comparison above was re-run after the fix.

---

### Still not verified

- **No Power BI Desktop / live-data check** — this is a static gate throughout.
- **The `reporting-only` class has no mechanical proof.** It is held by a line pin. A future edit that turns `_collect_drop_rows`' `described` map into a decision would pass the gate if the line text happened not to change. I could not find a non-fragile mechanical predicate for "does not decide"; the pin forces re-adjudication instead, and I would rather name the ceiling than imply the gate proves more than it does.
- **`OracleRecord` is not `CandidateIndex`** — see the deviation above. It implements your rule but is a second type; if #422 lands a declared kind on oracle records, converting it is the right follow-up.
- **The census counts `_slug` only.** Any other lossy comparison introduced elsewhere is out of its scope by design.

Guust-Franssens:
**CI green at `930c1c8`** — run [33595130355](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/actions/runs/33595130355), conclusion `success`.

Linux CI: **4125 passed / 98 skipped, 0 failed** (225s). Windows local: **4197 collected, 3 failed** — the three documented `test_upstream_repro_pins.py` tripwires, which is the whole of the platform split; every other gate is identical.

Guust-Franssens:
## Round 7 response — the census was vacuous, so I deleted the approach rather than patching it

Tip `a405522`, rebased onto master (0 behind). All five reproductions built and confirmed before fixing.

### Finding 2 first, because you were right about the root cause

> *a census that pins **where** a function is called cannot prove **how** its result is used*

I reproduced all three:

| attack | round-6 gate |
|---|---|
| unrelated `if unrelated == 1` in the same function | **8 passed** |
| collapse the lossy key one line later | **8 passed** |
| promote the reporting map into `_disposition_omissions` | **8 passed** |

⚠️ **My own first two reproduction attempts were wrong**, and I only noticed because they "failed for the right reason by accident" — one tripped the *line pin* rather than the multiplicity proof, and one changed only explanatory text. I rebuilt both faithfully before believing the result. Worth saying, since I am asking you to trust the replays.

**You told me not to add a fourth class or a fourth patch. I did neither — I deleted the approach.** Following a value through arbitrary Python is undecidable, so instead the value is made **inaccessible except through the guard**:

```python
class NormalizedIndex(Generic[_T]):
    def add(self, name, value)          # candidates stored as a LIST per key
    def add_spelling(self, name, value) # another spelling of the SAME candidate
    def count(self, name) -> int        # reporting only; an int cannot be mistaken for a candidate
    def unique(self, name) -> _T | None # returns None for 0 AND for >1 — the only way out
```

Every lossy join goes through it: workbook binding, exemption resolution, oracle workbook attribution. `_slug()` is now called in **six** places — four inside the index, two reporting.

The census file was **rewritten**, and its claim narrowed to three things that can each fail:

| claim | proof |
|---|---|
| every `_slug` call is inside the index or allowlisted | AST census |
| the index refuses both directions and leaks no bucket | behavioural + a public-surface pin (the store is name-mangled; *absence of a leak is not prohibition of one*) |
| the allowlisted sites do not decide | **behavioural non-interference** — blank the reporting map, every verdict must be identical, and the text must visibly change |

Verified after the fix: 2a fails (the `_slug` is outside the index), 2b fails, 2c fails. And **2c′ — a promotion that changes only explanatory text — correctly does *not* fail**, which is what shows the test discriminates rather than firing on everything.

⚠️ **Answering your direct question: is the third class provable at all? No — and the file says so.** Non-interference is established over the scenarios in its matrix, not in general; a promotion firing only on an input shape absent from that matrix would pass. That is narrower than round 6 claimed and it is the honest ceiling. The matrix **is** the coverage.

### Finding 1 — `view_type` was in the manifest and I was discarding it

Both reproduced (worksheet-typed record → dashboard PASS; `Bo ok` + `Bo-ok` → 2/2). And you were right that my "#422 hasn't landed" scoping understated it: `f3374d8` on master already ships `view_type`, and its own commit message says *"the observable defect still reproduces until oracle-coverage consults view_type"*. It does now.

- `dashboard`/`worksheet` consumed; **`unknown`, an unrecognised value and an absent key all mean CANNOT ESTABLISH**, and a kind-less record certifies nothing.
- Reference entries are dashboards **by construction** — `capture_tableau_reference.py` builds `dashboards[]` from the spec's `dashboards`, so the kind is known there without #422. ⚠️ Consequence worth naming: a **worksheet page now has no reference oracle at all** rather than being certified by a dashboard picture. That is correct, and it is a real reduction in coverage.
- Producer-workbook multiplicity preserved, uniqueness required on **both** sides.
- **Name matching is exact only.** A view name and a page name are both source-owned with no filesystem between them, so I removed the fallback round 6 asked me to *narrow* — a guard on a fallback with no mechanism behind it only narrows a match that was never justified. I am flagging this as a **reversal of a round-6 instruction**; the workbook fallback survives because there the unit side really is a sanitised artifact stem.
- Discarded evidence is disclosed: `admitted_evidence`, `kindless_evidence`, `unattributed_evidence`, `foreign_workbook_evidence`.

### Finding 3 — rebased. `git rev-list --count HEAD..origin/master` = **0**.

### The unverifiable estate claim is now a committed artifact

`tests/estate_page_gate_digest.py` + `tests/estate_page_gate_expected.json`. It stages every unit, runs **both halves** through the documented **relative** CLI path, cross-checks the resulting oracle **denominator** against parity, and hashes the summary.

```
51 units | 35 PASS / 9 PRECONDITION_FAILED / 7 NOT_CHECKED
19 declared-by-engine-unsigned, 2 unexplained, 2 no-source-content
0 blank, 0 unaccounted extra, 0 contested, 0 disagreements
sha256 eef8b1121899da209c083217755a02fa023d8c41dd0ad00da28a57262e7fdf80
```

Exit **0 match / 1 differs / 3 could not be measured** — all four paths verified, including that a missing bundle and an empty spec set report **UNMEASURABLE** rather than a clean digest. It is not a pytest test because CI has no estate bundle.

⚠️ Our digests will not match (`7a29f6c4…` vs mine) because they hash different canonical forms — mine is over the summary object this script emits. Comparison is by running **this script**; that is the point of committing it.

⚠️ **And the disclosure that matters more than the number.** The digest is identical under the committed baseline `check_unit.py`, and I checked *why* rather than reporting "no regression": **the estate contains ZERO reference records and ZERO oracle records across all 51 units** (43 units grade `no oracle manifest found`, 8 `expected page set could not be established`). **This estate cannot detect an oracle-layer regression at all** — so it is worthless as a baseline for finding 1, and my round-6 "no verdict changed" reassurance was weaker than it sounded for the same reason.

### Campaign — 87/87, and it caught an unkillable clause of mine

The rework staled 17 mutations and broke 3; all were rewritten against the new code rather than re-pointed. One then **survived**: `ORACLE KIND: a kind-less record is admitted anyway`. It survived legitimately — the `(kind, name)` lookup key already made the `continue` unreachable-by-effect, i.e. **vacuity mode 3, a clause arithmetically implied by its sibling.** Rather than delete a defence-in-depth guard, I pinned it independently by making it observable (`admitted_evidence`), which is also the number an operator wants. 87/87 caught, control SURVIVED, exit 0.

### Counts — measured at the tip I pushed

```
$ python -m pytest -q                    # Windows, a405522
3 failed, 4221 passed, 9 skipped   (4233 collected)
```

The three are the documented `test_upstream_repro_pins.py` tripwires.

| gate | result |
|---|---|
| `ruff format` / `ruff check scripts tests conftest.py` | exit 0 |
| `pylint scripts` | exit 0 |
| `pytest -q` | 4233 collected, 3 failed (tripwires) |
| `tests/mutate_check_unit.py` | 87/87, control SURVIVED, exit 0 |
| `verify_shared_identity_pin.py` | IDENTICAL to `origin/master@0a2c79e3`, exit 0 — the sibling merged, so the pin now tracks master |
| navigation / capability / convention | exit 0 |

### Still not verified

- **Non-interference is matrix-bounded**, as above. This is the honest limit of the third claim.
- **The estate proves nothing about the oracle layer** — zero records. Finding 1's fixes are covered only by unit tests.
- **`_slug` is the only lossy function policed.** A second normalizer introduced elsewhere is out of scope by design — a broad "any lossy comparison" rule produces false positives and gets disabled.
- **No Desktop / live-data check** anywhere in this PR; it is a static gate throughout.

Issue #402, opened 2026-08-30
Oracle manifest cannot say whether a captured image is a dashboard or a worksheet, and nothing binds it to the emitted PBIR page


Guust-Franssens:
Found by a reviewer question during the `Meridian Revenue by Region` capture: *""does the screenshot encompass the full page, or is it per view? how does it currently say which page the oracle belongs to?""* Both halves turned out to be unanswered.

## Gap 1 — no view-type discrimination

Tableau REST returns **dashboards and worksheets both under `/sheets/`**. `capture_tableau_oracle.py` records neither a `view_type` nor a `sheet_type`:

\\\
per-view manifest keys: view_luid, view_name, view_url_name, content_url,
                        workbook_luid, project, updated_at, captured_at,
                        data, image, workbook_name
\\\

Measured: the script contains **no reference to dashboards at all** — one incidental \iewUrlName\ at `capture_tableau_oracle.py:427` and nothing else.

So a consumer cannot tell whether a given PNG is a **full-page dashboard composite** or a **single worksheet viz**. Those are not interchangeable evidence: a dashboard image shows layout, zone placement and cross-visual arrangement; a worksheet image shows one mark set.

**This is not theoretical.** Scanned all 45 workbook assets in the trial estate:

| | |
|---|---|
| workbooks with >= 1 dashboard | **13 of 45** |
| total dashboards | **18** |
| total worksheets | 298 |

Worst case in-corpus: `global_superstores_db.twb` — **20 worksheets + 6 dashboards**, so up to 26 views land in one manifest with nothing separating the two kinds.

⚠️ Note a decoy while testing this: `Section 15 - Tableau Sales _ Customer Dashboards.twbx` is *named* ""Dashboards"" and has **15 worksheets, 0 dashboards**. Do not infer kind from the name.

## Gap 2 — nothing binds a captured view to the emitted PBIR page

For `Meridian Revenue by Region` the correspondence is real but **implicit and unrecorded**:

| oracle manifest | emitted PBIR |
|---|---|
| `view_name: ""Revenue by Region""` | `page-ws-Revenuebb7d27f78` -> `displayName: ""Revenue by Region""` |
| `view_luid: d78b0605-...` | *not referenced anywhere in the PBIR* |

The manifest never names the page id; the PBIR never names the view LUID. The join is a **name match a human has to make**, which breaks silently on a rename and is ambiguous wherever names repeat or a dashboard and a worksheet share a stem.

[...]

PR #445, opened 2026-09-02
fix(check_unit): move the lossy-key guarantee into a type, and bind oracle evidence to a KIND

Guust-Franssens:
Round 7 of PR #433 landed **after** that PR was squash-merged (`1f85200`, merged 06:33Z at the round-6 state `930c1c8`), so it continues here. #433 carries rounds 1-6 and the full review history.

**All five round-7 reproductions were built and confirmed before fixing.** Details, evidence and the honest limits are in [this comment on #433](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/pull/433#issuecomment-5505823227); the summary:

## 1. HIGH — the `_slug()` census was vacuous in all three classes

Reproduced: an unrelated `if unrelated == 1` satisfied `uniqueness-guarded`; a collapse one line later was invisible to `multiplicity-preserving`; and a reporting-only value could be **promoted into a decision without touching a single pinned line**. Root cause, kept as a sentence in the source:

> *a census that pins **where** a function is called cannot prove **how** its result is used.*

I did not add a fourth class or a fourth patch — **the approach is deleted**. The guarantee moved into a type, `NormalizedIndex`: candidates stored as a LIST per key, and `unique()` is the only accessor, returning `None` for **0 and for >1**. `_slug()` is now called in six places — four inside the index, two reporting-only.

The census file was rewritten to prove three narrower things that can each fail: the AST census of call sites; the index refusing both directions and leaking no bucket (name-mangled store + a pinned public surface, because *absence of a leak is not prohibition of one*); and **behavioural non-interference** for the allowlist — blank the reporting map, every verdict must be identical and the explanatory text must visibly change.

⚠️ **Is the third class provable in general? No, and the file says so.** Non-interference is established over its scenario matrix, not universally. That is narrower than the previous claim and it is the honest ceiling.

## 2. HIGH — oracle evidence certified the wrong object

`view_type` has been in the manifest since `f3374d8` and was being discarded here — that commit says so itself (*"the observable defect still reproduces until oracle-coverage consults view_type"*), so this also closes the consumer half of #402.

- `dashboard`/`worksheet` consumed; **`unknown`, an unrecognised value and an absent key all mean CANNOT ESTABLISH**, and a kind-less record certifies nothing.
- Reference entries are dashboards **by construction**. ⚠️ Consequence: a worksheet page now has **no reference oracle at all** rather than being certified by a dashboard picture — correct, and a real reduction in coverage.
- Producer-workbook multiplicity preserved; uniqueness required on **both** sides (`Bo ok` + `Bo-ok` were both being admitted to `Book`).
- **Name matching is exact only** — both sides are source-owned with no filesystem between them. ⚠️ This *reverses* a round-6 instruction to merely narrow the fallback; a guard on a fallback with no mechanism behind it only narrows a match that was never justified.
- Discarded evidence is disclosed: `admitted_evidence`, `kindless_evidence`, `unattributed_evidence`, `foreign_workbook_evidence`.

## 3. The unverifiable estate claim is now a committed artifact

`tests/estate_page_gate_digest.py` + `tests/estate_page_gate_expected.json` — stages every unit, runs both halves through the documented **relative** CLI path, cross-checks the oracle **denominator** against parity, and hashes the summary. Exit **0 match / 1 differs / 3 could not be measured**; all four paths verified, including that a missing bundle and an empty spec set report UNMEASURABLE rather than a clean digest.

```
51 units | 35 PASS / 9 PRECONDITION_FAILED / 7 NOT_CHECKED
19 declared-by-engine-unsigned, 2 unexplained, 2 no-source-content
0 blank, 0 unaccounted extra, 0 contested, 0 disagreements
sha256 eef8b1121899da209c083217755a02fa023d8c41dd0ad00da28a57262e7fdf80
```

⚠️ **The disclosure that matters more than the number:** the estate contains **ZERO reference records and ZERO oracle records across all 51 units** (43 grade `no oracle manifest found`, 8 `expected page set could not be established`). **It cannot detect an oracle-layer regression at all**, so it is worthless as a baseline for finding 2 above — those fixes are covered only by unit tests.

## Campaign — 87/87, and it caught an unkillable clause of mine

The rework staled 17 mutations and broke 3; all were rewritten against the new code rather than re-pointed. One then **survived**: the kind guard's `continue` was arithmetically implied by the `(kind, name)` lookup key — vacuity mode 3. Rather than delete a defence-in-depth guard I pinned it independently by making it observable (`admitted_evidence`), which is also the number an operator wants.

## Gates — measured at this tip, rebased onto current master

```
$ python -m pytest -q                    # Windows
3 failed, 4324 passed, 9 skipped   (4336 collected)
```

The three are the documented `test_upstream_repro_pins.py` tripwires.

| gate | result |
|---|---|
| `ruff check` / `ruff format --check` | exit 0 |
| `pylint scripts` | exit 0 |
| `pytest -q` | 4336 collected, 3 failed (tripwires) |
| `tests/mutate_check_unit.py` | 87/87 caught, control SURVIVED, exit 0 |
| `tests/estate_page_gate_digest.py` | MATCHES, exit 0 |
| `verify_shared_identity_pin.py` | IDENTICAL to `origin/master@cdb18659`, exit 0 |
| navigation / capability / convention | exit 0 |

## Still not verified

- Non-interference is **matrix-bounded**, as above.
- The estate proves nothing about the oracle layer — zero records.
- `_slug` is the only lossy function policed; a second normalizer elsewhere is out of scope by design.
- No Desktop / live-data check anywhere — this is a static gate throughout.

Refs #432, refs #402, continues #433


Guust-Franssens:
**CI green** — run [33604399127](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/actions/runs/33604399127), conclusion `success`.

Linux CI: **4238 passed / 98 skipped, 0 failed**. Windows local: **4336 collected, 3 failed** — the three documented `test_upstream_repro_pins.py` tripwires, which is the whole of the platform split.

Issue #435, opened 2026-09-01
CI runs none of the 15 engine-dependent tests - they SKIP silently, and a skip is not a pass


Guust-Franssens:
## The gap

**15 test cases depend on the deterministic conversion engine, and CI runs none of them.** They do
not fail there — they *skip*, silently, because `requires_engine` degrades an absent plugin into a
`pytest.mark.skipif`. `.github/workflows/checks.yml` has **no engine installation step**, and the
engine is an installed Copilot plugin rather than a repo dependency, so `uv sync --all-extras` cannot
supply it.

⚠️ **A skip is not a pass.** Everything below currently reports green in CI while having proved
nothing.

## Measured

| module | `@requires_engine` decorators | test cases | what they cover |
|---|---:|---:|---|
| `tests/test_issue_424_chart_type_pin.py` | 5 | **8** | #424's emitted-type matrix, exact PBIR bindings, the permanent invariants, the visual-id falsification, the engine's warning surface |
| `tests/test_dax_oracle_server.py` | 4 | **4** | the engine's real DAX-oracle contract, subprocess wiring, reconciliation, scalar extraction |
| `tests/test_upstream_repro_pins.py` | 3 | **3** | upstream behaviour / drift pins (#166, #168, #171) |
| | | **15** | |

(The 8 in the first row is 5 decorators, one of them parametrized over the four permanent
invariants.)

Locally, with the engine installed: `3 failed, 3903 passed, 9 skipped`. The three failures are
`test_upstream_repro_pins.py`'s deliberate `PINNED_ENGINE_VERSION = "2.260.0"` guard against the
installed 2.339.0 — a *signal*, and one CI can never see either.

⚠️ **A methodological caveat on my own measurement, because it nearly produced a wrong number.**
Simulating absence with `T2P_SIMULATE_ENGINE_ABSENT_FOR_TESTS=1` reports only **11** skips — the env
var is read by just two of the three modules (`test_dax_oracle_server.py` gates on
`engine_source.engine_root()` directly, so the simulation cannot reach it). The real CI figure is 15.
Anyone re-measuring this should count `@requires_engine` per module rather than trust the simulation.

⚠️ Also unresolved and **not** attributed to the engine: a full serial run under that simulation
reported `1 failed, 3894 passed, 20 skipped`, and I did not identify the single failure. Only those
two modules read the variable, so it is unlikely to be engine-related — but it is unverified, and
worth a look by whoever picks this up.

[...]

PR #453, opened 2026-09-02
Fixes #447: gate the window-spawning tests behind an opt-in marker

### LINKED ISSUE #447: The 'Fake Desktop' credential tests spawn real GUI windows on every pytest run - gate them behind an opt-in marker
The credential-modal detection tests spawn **real GUI windows** on every full `pytest` run, because
`pyproject.toml` sets:

```toml
testpaths = ["tests", ".github/skills"]
```

and `.github/skills/pbip-model-refresh/tests/test_credential_modal_detection.py` launches a WinForms
app titled **"Fake Desktop"** (`_FAKE_DESKTOP_APP`, two spawn sites).

## This is not a complaint about the test

It is a good test, and it **must** spawn a real window — UI Automation cannot be exercised against a
mock. The implementation is careful for measured reasons recorded in its own comments: the form's
class starts with `WindowsForms10.Window.8` so the probe excludes it exactly as it excludes the real
Power BI Desktop main window, and the Refresh control is a **WPF** button hosted in an `ElementHost`
because a plain WinForms `Button` surfaces to UI Automation as `ControlType.Pane` with **no patterns**,
so the probe's `Name -eq 'Refresh'` + Invoke scan would never fire and the run would end at the
not-invoked guard "instead of testing anything".

What it protects is real: `probe_desktop_credential.ps1` has a known fail-open shape — it can report
`ContentRead: true` when *any* non-caption text is read, and so exit 0 having never read the
credential text (see #406, #417).

## The actual defect: it is un-gated

It carries `serial` (exclusivity — it contends for a singleton UI-Automation tree) but **nothing
deselects it by default**. So it runs while someone is editing documentation, and windows steal focus
on an operator's machine for hours. Reported from the field: *"this fake desktop has been opening and
closing for the last 2 days tbh I dont know what is the immediate value here"* — which is the correct
reaction, because during work that never touches the credential probe the immediate value **is** zero.

⚠️ `serial` is not a substitute. It only takes effect under `-n auto --dist loadfile`, which deselects
`serial`/`timing`; a plain `pytest` run — what an agent typically runs — executes them.

## Suggested fix

Add a `gui` marker, apply it to the two spawn sites, and deselect it by default:

```toml
markers = [
    ...
    "gui: spawns a real top-level window; needed for UI-Automation coverage, opt-in only",
]
addopts = "-m 'not gui'"
```

Then run it deliberately — `pytest -m gui` — when touching `probe_desktop_credential.ps1`, the bridge,
or the refresh path, and in CI where no one is watching a desktop.

⚠️ **Controls, because a marker that hides a test is exactly how coverage dies silently:**

- **positive** — `pytest -m gui` still selects and passes both tests; they are not orphaned
- **negative** — a default `pytest` run **collects zero** `gui` tests (assert the deselected count, not
  merely that the suite is green)
- **CI** — the workflow must run `-m gui` explicitly, or this trades a focus-stealing window for
  silently-lost coverage of a fail-open credential guard. Compare with #435, where 15 engine-dependent
  tests already skip silently in CI and a skip was being read as a pass.

## Related

- #406 / #417 — the credential-modal detector these tests cover
- #435 — silent skips in CI, the failure mode this fix must not reproduce
- #415 — parallel-test flakiness and marker semantics

Guust-Franssens:
A bare `pytest` collects `.github/skills` (`testpaths`), so it launches a real WinForms window titled **Fake Desktop**. Reported from the field, mid-demo: it had been *"opening and closing for the last 2 days"*.

⚠️ **The tests are good and must keep spawning a real window** — UI Automation cannot be exercised against a mock, and their own comments record why each choice was measured (the WPF-button-in-an-`ElementHost` exists because a plain WinForms `Button` surfaces to UIA as `ControlType.Pane` with no patterns). They protect `probe_desktop_credential.ps1`, which can report `ContentRead: true` having never read the credential text (#406, #417). So this makes them **opt-in, not weaker**.

## Invariant and direction

An ordinary run must spawn **no window**; the tests must still run **somewhere**. Failing open here means a window steals focus on an operator's machine. Failing closed means the coverage silently disappears — which is **#435** exactly (*"CI runs none of the 15 engine-dependent tests — they SKIP silently, and a skip is not a pass"*). Both are guarded.

## Closed surface — every pytest invocation, N = 4

| invocation | runs `gui`? |
|---|---|
| `checks.yml` line 67, `uv run pytest -q` (ubuntu) | no — correct, WinForms/UIA is Windows-only |
| `checks.yml`, bundle job on `windows-latest` | no |
| `checks.yml`, **new** `-m gui` step on `windows-latest` | **yes** |
| any local/agent `pytest` | no |

`serial` was not a substitute: it only takes effect under `-n auto --dist loadfile`, and a plain `pytest` runs them anyway.

## Controls, each mutation-proven

| mutation | result |
|---|---|
| remove `addopts` | **CAUGHT** — `test_a_default_run_deselects_every_gui_test` |
| drop `-m gui` from CI | **CAUGHT** — `test_ci_runs_the_gui_tests_explicitly` |
| strip the markers | **CAUGHT** — 3 controls fail |

⚠️ Every mutation was verified to **apply** before being scored — one earlier attempt "survived" only because a PowerShell regex never matched the line.

## Two defects of my own that the mutations caught

- The first marking script marked **9** tests, not 7: it sliced each body to the next `def test_`, swallowing module-level constants between tests. Bounded by indentation -> **7**, all of which already carry `serial`.
- The CI control first asserted `'-m gui' in workflow` and the deletion mutation **SURVIVED**, because the phrase remained in the comment beside it. It now parses `run:` commands only.

## Also narrows one over-broad assertion

`test_parallel_test_loop.py` asserted `'deselected' not in output` as a **proxy** for "the contended tests came back". That held only while `serial`/`timing` were the sole deselection. `--include-contended` is about the parallel tier, not about hijacking a desktop, so `gui` stays deselected — and the assertion now pins the remaining count to the gui-marked set, so a **new** unexplained deselection still fails.

## Answering the three questions

- **Can each test reach the production path?** Yes — all three mutations above fail their named guard.
- **Does the guard cover the whole surface it names?** The N=4 table is the enumeration.
- **Does the evidence prove the verdict?** Counts, not greenness: the deselect control asserts the deselected count equals the marked count; a suite green because tests vanished would fail it.

## Gates, by exit code

`pytest tests/test_parallel_test_loop.py tests/test_gui_marker_gate.py` **0** (24 passed) · `ruff format --check` **0** · `ruff check` **0** · `pylint` **0**.

Full `pytest tests/` before this fix: 4 failed / 4036 passed — 3 known engine-drift pins plus the parallel-loop test corrected here. ⚠️ It takes **4h40m** serially; CI is the authoritative run.

Independent corroboration: the PR #431 reviewer's run hit `DIALOG_UNREADABLE` *"because an unaccounted GUI window was open"*.

### DIFF
--- a/.github/skills/pbip-model-refresh/tests/conftest.py
+++ b/.github/skills/pbip-model-refresh/tests/conftest.py
@@ -7,9 +7,12 @@ absolute path, or a walk up to a repo root, would silently re-bind the tests to 
 `scripts/` folder and prove nothing about the copy.
 
 It also pins the DEFAULT Desktop-inspection state every test runs against - see
-:func:`healthy_desktop_state` for why that has to be explicit rather than inherited from the host OS.
+:func:`healthy_desktop_state` for why that has to be explicit rather than inherited from the host OS,
+and it owns the **`gui` exclusion** for exactly the same portability reason - see
+:func:`pytest_collection_modifyitems`.
 """
 
+import os
 import sys
 from pathlib import Path
 
@@ -20,9 +23,16 @@ SKILL_SCRIPTS = Path(__file__).resolve().parents[1] / "scripts"
 if str(SKILL_SCRIPTS) not in sys.path:
     sys.path.insert(0, str(SKILL_SCRIPTS))
 
+# Kept in step with the host repo's root `conftest.py` BY DUPLICATION, on purpose: this folder is
+# copied out of the repo and run with the repo unimportable, so it cannot import a shared helper.
+GUI_MARKER = "gui"
+RUN_GUI = "--run-gui"
+RUN_GUI_ENV = "T2P_RUN_GUI"
+TRUTHY = {"1", "true", "yes", "on"}
+
 
 def pytest_configure(config: pytest.Config) -> None:
-    """Register `timing` and `serial` HERE so the markers travel with the bundle.
+    """Register `timing`, `serial` and `gui` HERE so the markers travel with the bundle.
 
     **`timing`** - a handful of tests in this folder assert a sub-second wall-clock budget on an
     operation that takes ~0.03s. Measured (issue #387) under 22 concurrent pytest-xdist workers, one
@@ -38,6 +48,12 @@ def pytest_configure(config: pytest.Config) -> None:
     the probe degrading safely, and the test's stricter assertion correctly refusing it. Not one of
     them failed in seven runs that were not concurrently paired, so this needs the pairing.
 
+    **`gui`** - ten tests here spawn a real top-level window (seven a WPF app, three native
+    `CreateWindowExW` windows), which steals focus from whoever is at the keyboard; issue #447 is an
+    operator watching that happen for two days mid-demo. Registering it here is not cosmetic:
+    measured before this existed, the copied-out run emitted seven `PytestUnknownMarkWarning`s, and
+    an unregistered marker is one `--strict-markers` away from being an error in someone else's repo.
+
     Registering them in this conftest rather than a host `pyproject.toml` is the whole point: copy
     this folder into another repo and `-m "not (serial or timing)"` still works, with no
     unregistered-marker warning.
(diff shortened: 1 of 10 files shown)


Guust-Franssens:
## Blind round-1 review: **BLOCKED**

Reviewed at `d9b2efb`; CI green for that exact SHA (run `33680513674`, both `checks` and `windows-bundle`).

The PR does not make a default test run window-free. It gates seven WPF tests but misses three native-window tests and a nested pytest route that still executes every GUI test.

### Marker-completeness table

| Window-spawning test (`tests/test_credential_modal_detection.py`) | Line | marked `gui`? |
|---|---:|:---:|
| `test_the_win32_harvest_reads_real_ownership_and_owner_enabled_state` | 1919 | **No** |
| `test_a_native_unowned_credential_host_owning_a_tooltip_is_not_the_application` | 2146 | **No** |
| `test_the_process_main_window_handle_is_a_z_order_answer_not_an_identity` | 2190 | **No** |
| the seven currently-marked WPF tests | 2931, 3388, 3398, 3412, 3422, 3432, 3686 | Yes |

The seven marked ones genuinely spawn windows (opt-in run: 7 passed in 77.40s). The three unmarked ones create `WS_VISIBLE` top-level windows via `CreateWindowExW` — proven by replacing each first executable line with a raise: `3 failed — all reached` by default, `3 deselected` under `-m gui`.

### Blocking findings

**1. `pytest tests/` still runs the GUI suite through a nested pytest.** `tests/test_skills.py:97-102` deletes `PYTEST_ADDOPTS`, `:105-117` adds no filter, and `:142` runs the copied bundle **outside the repository**, where `pyproject.toml` does not exist. The copied `conftest.py` registers only `serial` and `timing`, not `gui`. Instrumenting both WPF spawn sites: `7 failed, 279 passed` with `RuntimeError: WPF_SPAWN_REACHED`. Meanwhile the branch's own `pytest -q tests/` reports **zero deselections** — the nested execution is invisible in the outer summary.

**2. Three real window tests unmarked** — table above.

**3. An explicit `-m` REPLACES the safety filter.** `pyproject.toml:81` is `addopts = "-m 'not gui'"`; a command-line `-m` overrides rather than composes.

| command | selected |
|---|---|
| bundle default | 302/309 |
| bundle `-m gui` | 7/309 |
| bundle `-m "not slow"` | **309/309 — every GUI test runs** |
| `pytest tests/ -m "not something_else"` | **4046 selected** |

`docs/offline-mock-harness.md:289` documents `pytest -m "not slow"` as supported — so following our own docs opens the windows.

**4. The new controls cover one file, not the repository's GUI surface.** `tests/test_gui_marker_gate.py:30-48` hardcodes one bundle and one marked file. Inverse-completeness mutation — a new top-level `@pytest.mark.gui` test that raises — left **all four controls passing**. A control that cannot fail for the thing it is named for.

### Reconciliation (exact, and clean)
master `0fae0cf7`: 4362 collected / 4350 passed / 3 failed / 0 deselected. Branch `d9b2efbe`: 4366 / 4347 / 3 / 7 deselected. Four new gate tests explain +4; seven WPF tests moved passed -> deselected; net `+4 - 7 = -3`. Same three known engine-drift failures throughout.

### Required
Mark all ten; replace the global `addopts -m` with a collection hook plus an explicit opt-in that an unrelated `-m` cannot override; propagate the exclusion into the copied/nested bundle; make the completeness control genuinely inverse-complete.

Routed to a fresh agent; re-review returns to this reviewer.

PR #454, opened 2026-09-02
fix(gates): join evidence to a workbook by LUID, once, for both gates (#450)

### LINKED ISSUE #450: check_unit's workbook-attribution guard reads 'workbook' but the oracle writes 'workbook_name' - inert on all 360 real records (fails open)
Found while verifying PR #445 against the real oracle produced by the 2026-09-02 dry run
(`_runs/407-dryrun-gates/`, 360 views, 312 captured).

## The defect

`check_unit.py`'s workbook-attribution guard reads a manifest field named **`workbook`**.
`capture_tableau_oracle.py` writes that field as **`workbook_name`**. The names never match, so the
guard is **inert on every real capture**:

```
evidence_unattributed_to_a_workbook = 360 of 360
```

Every single record is treated as unattributable. The two-sided workbook check — the one that would
stop a **foreign workbook's** render being admitted as evidence for this unit — has therefore never
fired against a real oracle.

## Direction: fails OPEN

A foreign workbook's record is **admitted** rather than discarded. That is the dangerous direction:
it credits coverage that does not exist, which is the same class as #438 (a worksheet's capture
certifying a dashboard page).

## Why it did not block #445

⚠️ Recording the reasoning, because "fail-open findings block the merge" is the rule and this is a
deliberate, argued exception rather than a lapse:

It is **pre-existing master behaviour**. The field name is read identically on both sides of the
#445 diff — that PR neither introduces the mismatch nor widens it. What #445 *does* add is the kind
guard that closes the larger half of #438. Holding it would have left this gap open **and** discarded
that fix, which is strictly worse in both directions.

## Ground truth for the fix

The manifest entry shape, measured from `_runs/407-dryrun-gates/oracle/oracle-manifest.json`:

```
view_luid, view_name, view_url_name, content_url,
workbook_luid, workbook_name, project, updated_at, captured_at,
view_type,                      # dashboard | worksheet | unknown
data{status,path,bytes,sha256}, image{status,path,bytes,sha256}
```

⚠️ **Prefer `workbook_luid` over `workbook_name`.** A display name is not unique — two projects can
hold workbooks with the same name, which is the exact ambiguity `_runs/<NNN>-<slug>/` numbering exists
to avoid elsewhere in this repo. The LUID is the identity; the name is decoration.

## Controls the fix needs, both in one run

- **positive** — a render whose `workbook_luid` matches the unit is **admitted**, and
  `evidence_unattributed_to_a_workbook` falls from 360 to ~0
- **negative** — a render from a **different** workbook is **discarded**, and is counted in
  `foreign_workbook_evidence` rather than silently dropped
- **fail-closed on absent** — a record with **no** workbook identity must remain unattributed and
  certify nothing; it must not default to "belongs to this unit"

⚠️ **A fix that only makes the counter go to zero is not a fix** — deleting the guard would do that
too. Assert the discrimination (foreign records refused) alongside the admission.

## Related

- #438 — the kind half of the same evidence-admission problem (fixed in #445)
- #446 — the per-unit handover package, which has to resolve renders to a unit and so depends on this
  join being correct

Guust-Franssens:
# Round 3 - the revision key is reproducible, so `confirmed` is reachable

The review was right: *"no reproducible comparison exists"* was false. One existed - I was hashing the
container instead of the contents - and 129 of 167 ready pages carrying `revision=unconfirmed` was an
artifact of that choice, not a property of the estate.

**I re-measured independently rather than re-running the reviewer's script**: three downloads (not
two) of every item in one run, datasources as well as workbooks, asset shape recorded per item.

| 48 workbooks + 19 datasources, 3 downloads each | raw sha256 DIFFERS | content key DIFFERS |
|---|---|---|
| archives (49) | **27** | **0** |
| non-archives (18, plain `.twb` XML) | **0** | n/a |
| **content key unstable anywhere** | | **NONE** |

## Two contradictions of the brief, both measured, both load-bearing

**1. "Not a readable zip -> `unconfirmed`" would strand 18 of 48 workbooks.** They download as plain
`.twb` XML, and **all 18** had a stable raw digest across three downloads - zero counter-examples.
Refusing to confirm 37% of the estate on the strength of a rule about zip repacking, which does not
apply to them, is a fail-closed regression. The key is therefore keyed by **shape** - archive /
Tableau XML / neither - each with its own versioned algorithm. ⚠️ This is *not* the silent fallback
the brief forbids: a payload that **is** an archive and will not open still yields **no key at all**.

**2. A content digest alone was NOT sufficient - and shipping it would have been worse than the bug.**
Re-stamping the 78 harvested assets against the site, **28 still reported `differs`**. Diffing six of
them - three inside a `.twbx`, three served flat - the entire difference was one line, every time:

```
-<!-- build 20263.26.0824.1544 -->
+<!-- build 20263.26.0828.1352 -->
```

That is the Tableau **server** build stamp: the site was upgraded between the harvest and the check,
and most of those files were byte-identical in length. Left in, it trades 129 false `unconfirmed` for
**28 false `differs`** - a *refusal*, at scale, on evidence that is fine. XML comments are now
normalised out (archive key -> `twbx-content-v2`), and the 28 became **0**.

## Re-measured estate (fixture backed up, re-stamped, measured, restored)

| | round 2 | round 3 |
|---|---|---|
| `pages_ready` | 167 | **167** |
| `pages_blind` | 49 | **49** (floor ~46) |
| `pages_unverifiable` | 4 | **4** |
| `units_ready` | 25 | **25** |
| **`pages_revision_unconfirmed`** | **129** | **0** (168 `confirmed`) |
| census | `luid=295 foreign=12851 unknown=0 stale=0` | **unchanged** |

Provenance verdicts after re-stamping: `sha256/same` 20, `name_only/same` **28**, `name_only/None`
30 (a flat `.twb` local vs an archive remote - genuinely not comparable, and `None` never means
drift). Zero `differs`.

## The mutation harness is committed - `tests/mutate_revision_key.py`, 9/9

⚠️ Every mutant carries a **probe** that calls the patched object and asserts a mutation-specific
result *before* the anchor runs, per the brief. It earned its keep immediately: one probe **refused
to fire**, exposing a vacuous assertion in my own test - a renamed zip member changed the **sort
order**, so `test_genuinely_different_content_still_differs` would have passed with the member name
dropped from the digest entirely. The rename is now `Cook.twb`, which keeps the sort position.

Mutants: raw-hash the container · drop the member name · silently key an unreadable archive · compare
across algorithms · read a missing `revision_match` as drift · ignore `revision_match` · stop
recording the local key · stop normalising the build stamp · hash flat XML raw.

## Residual, named

`capture_tableau_reference.py`'s `source_workbook_sha256` is still a raw digest, and
`check_unit` still compares it raw. Both sides are **local file reads**, where a raw digest is
deterministic, so this is not the measured defect - but a reference captured against one download and
checked against a re-harvest of the same content would read FOREIGN. Fail-closed, and `check_unit.py`
is deliberately untouched here because PR #451 overlaps it.

## Gates - by exit code, every one the CI job runs

| gate | exit |
|---|---|
| `pytest tests/` | **1** - `3 failed, 4102 passed, 6 skipped`; the 3 are the pre-existing `test_upstream_repro_pins.py` engine-drift failures |
| `pytest .github/skills/pbip-model-refresh/tests/` | **0** (306 passed) |
| `pylint scripts` / `conftest.py` / `pbip-model-refresh` / `powerbi-ai-readiness` | **0** / **0** / **0** / **0** |
| `validate_spec.py --all --check` | **0** |
| **`set_data_folder.py --check` (privacy)** | **0** |
| `check_datamodel.py --all` | **0** |
| `check_agent_capabilities.py` | **0** |
| `sync_agent_conventions.py --check` | **0** |
| `set_ai_instructions.py --check` | **0** |
| `check_identity_normalization.py` | **0** |
| `mutate_revision_key.py` / `mutation_reference_readiness.py` / `mutate_check_unit.py` | **9/9** / **77/77** / **94/94** |

---

# Round 2 — four fail-open blockers, one rule

Round 1 found four fail-open blockers and they are **one class**: a record whose **machine identity
could not be established or compared** was rescued by something weaker — a display name, or merely
its location. Per the stop rule in `docs/review-throughput-postmortem.md`, this is one rule, not four
patches:

> **If a record carries a machine identity (LUID or sha256) that the unit cannot establish or
> compare, the result is `unknown`. Never fall through to a weaker axis. Never let location override
> a failed or unestablished join.**

It lives in `WorkbookIdentity.attribute`, which now walks the machine axes first: *the record makes
no claim on this axis* → try the next; *the record claims it and the unit cannot answer* → **`unknown`,
stop**; *both answer* → decisive. The display name decides only when the record claims no machine
identity at all.

Every blocker was rebuilt as a reproduction **before** any fix.

| blocker | reproduction, before | after |
|---|---|---|
| **B1** an unanswerable LUID rescued by an equal name — entry gate | `readiness=ready`, `pages_ready=1`, `pages_blind=0`, admitted by `name` | `readiness=blind`, `pages_ready=0`, **`unknown=1`** |
| **B1** the same record — exit gate | `PASS`, visual **and** numeric, `admitted_evidence=1` | `NOT_CHECKED`, visual 0 / numeric 0, **`unattributed=1`** |
| **B2** changed-revision certified as current | `readiness=ready`, generic oracle grade, **nothing disclosed** | admitted with **`revision: "unconfirmed"`**, `pages_revision_unconfirmed=1`, `[REVISION NOT ESTABLISHED]` printed — and `unverifiable` + `stale=1` when drift is soundly established |
| **B3** location certifies a unit-local reference manifest | `PASS`, visual **and** numeric, `admitted_evidence=1` | `NOT_CHECKED`, `admitted_evidence=0`, **`unattributed=1`** |
| **B3b** …the same, with a sha naming **another workbook** | `PASS`, `admitted_evidence=1` | `NOT_CHECKED`, **`foreign=1`** |
| **B4** Windows handover path on POSIX | `PureWindowsPath`→LUID, **`PurePosixPath`→`None`** | LUID on **both** flavours, both path spellings |
| **M1** non-packaged unit 3 levels below the run | oracle dirs found: `[]` | `['oracle']` — and a package still inherits nothing |
| **M2** merged census | `first name=1` + `second foreign=7` → merged **`foreign=0`** | merged **`foreign=7`**, `name=1` |

⚠️ **B1's old negative test omitted `workbook_name`, so it never exercised the path.** The replacement
supplies both fields, as every real oracle record does, and there is now a second exit-gate test for
the case the existing one *structurally could not reach* — a unit with **no** LUID at all.

## ⚠️ Contradicting the review, with a measurement: `name_only` is not evidence of a changed build

The review asked for `match: "name_only"` to be classified stale/unverifiable. Implemented literally,
that **regressed the estate from 167 ready pages to 42**, marking 18 of 67 units, 246 records and 125
pages unverifiable. Before accepting either outcome I measured *why* those inputs are `name_only`.

`find_origin` compares the harvested bytes against `content_sha256(luid)` — a fresh download. On the
407 estate that **identical request was issued twice for each of 30 workbooks inside one run**:

| observation | count |
|---|---|
| LUIDs whose two calls returned **different** digests (server repacks a `.twbx` per request) | **18 / 30** |
| LUIDs where the remote digest was stable — and the `.twbx` matched it **exactly** (`sha256`) | 12 / 30 |
| single-file LUIDs | 18, **all `sha256`** |
| units flagged stale by `match` alone | **18** |
| units flagged stale when the comparison must be **reproducible** | **0** |

So on this estate `name_only` is 100% explained by download packaging and 0% by drift. Treating it as
drift refuses correct evidence at a rate that gets a gate switched off — which this repo names as how
a good gate dies.

**The review's own escape clause — "unless the capture is independently bound to the migrated
revision" — is what resolves it**, and the revision axis is now three-valued:

- **`confirmed`** — a sha join, or `match: "sha256"`. Certifies normally.
- **`mismatch`** — `name_only` **and** every `remote_sha256` for that LUID agrees, so the comparison
  is reproducible. → `WB_STALE`, page **UNVERIFIABLE**, refused and counted. (Fires on the fixture;
  flags 0 healthy units.)
- **`unconfirmed`** — anything else, including no provenance. Admitted, and **never silently claimed
  as current**: `row["revision"]`, `pages_revision_unconfirmed`, and a rendered
  `[REVISION NOT ESTABLISHED]` line.

That satisfies blocker 2's substance — *nothing discloses the mismatch* — at zero false-alarm cost.
If you still want `unconfirmed` to be a hard finding, `--require-validation-grade` already refuses
oracle-grade evidence outright; say the word and I will make it a first-class flag instead.

## Estate — no regression, plus the safety counters

| | round 1 | round 2 |
|---|---|---|
| `pages_ready` | 167 | **167** |
| `pages_blind` | 49 (never 0) | **49** |
| `pages_unverifiable` | 4 | **4** |
| `units_ready` | 25 | **25** |
| `pages_revision_unconfirmed` | *not reported* | **129** (39 confirmed) |
| census | `luid=295, foreign=12851` | `luid=295, foreign=12851, unknown=0, stale=0` |
| HR_Dashboard | 4/7, `luid=23, foreign=290` | **4/7, `luid=23, foreign=290`** |
| exit gate, HR package | `PASS 7/7`, admitted 23 | **`PASS 7/7`, admitted 23, unattributed 0** |
| exit gate, **foreign** unit | `NOT_CHECKED 0/7` | **`NOT_CHECKED 0/7`** |

**Per the review's closing lesson, the safety counters are now reported beside the coverage numbers
by default** — `[ATTRIBUTION]` names admissions by route and refusals by reason, and
`[REVISION NOT ESTABLISHED]` names the pages whose revision was not proven. A 4x coverage rise can no
longer be mistaken for a cleared gate.

## The mutation harness is committed, not claimed

Folded into the two existing runners rather than a third framework:
**`mutation_reference_readiness.py` 77/77** (153 anchor/control checks) and
**`mutate_check_unit.py` 94/94**, negative controls surviving as required.

Three honest notes from doing it:

- A mutant aimed at `Evidence.is_for` and one aimed at `_readiness_result`'s old signature had gone
  **vacuous** — reporting CAUGHT while patching something the gate no longer called, or failing on a
  `TypeError`. Both are re-aimed.
- `a-luid-record-discards-its-workbook-name` became **inert by construction** (a LUID-bearing record
  never reaches the name axis now) and is re-aimed at the case where the name is still load-bearing.
- ⚠️ **There is deliberately no `check_unit` mutant for `Path(...).stem` vs `persisted_stem`.** On
  Windows `WindowsPath` accepts both separators, so no fixture exists that `Path` mishandles — the
  mutant is unkillable on this host **by construction**, and an entry that can only ever report
  CAUGHT on Linux is a vacuity. The property is pinned by
  `test_the_recorded_path_parse_does_not_use_the_running_hosts_flavour` (exercises both pure flavours
  on one host) and `test_a_windows_recorded_source_id_still_yields_its_luid` (fails on POSIX CI).

## Residuals, named

- **The rule is one-directional.** A record carrying *only* a display name is still admitted against a
  unit that has a LUID, because nothing stronger is on offer from that producer. It is the weakest
  route and every admission through it is counted separately (`census["name"]`).
- **A unit that cannot locate its source asset refuses sha-bearing evidence.** That is blocker 3's
  cost, stated rather than hidden, and it is counted in `unattributed_evidence`.
- `scripts/package_unit.py` (PR #451) still implements the unit-LUID rule a second time; converging it
  on `object_identity` is a follow-up, not this PR.

## Gates — by exit code

| gate | exit |
|---|---|
| `pytest tests/` | **1** — `3 failed, 4086 passed, 6 skipped`; the 3 are the pre-existing `test_upstream_repro_pins.py` engine-drift failures (they import only `engine_source`) |
| `ruff format --check .` / `ruff check .` | **0** / **0** |
| `pylint scripts` | **0** (it printed `10.00/10` while exiting **24** and then **16** before three real findings were fixed: two `too-many-locals`, a line-length, an over-cap module and an inner import) |
| `pylint conftest.py` | **0** |
| `check_navigation_index.py` / `sync_agent_conventions.py --check` / `check_identity_normalization.py` | **0** / **0** / **0** |
| `set_data_folder.py --check` (privacy) / `check_agent_capabilities.py` / `check_datamodel.py --all` | **0** / **0** / **0** |

⚠️ **The privacy gate caught a real leak and I had not been running it.** My blocker-4 path fixtures used `C:\Users\x\...` and `/home/runner/...` — exactly the shapes
`set_data_folder.py --check` exists to keep out of tracked files. It failed CI on `30a40cc` and is fixed in `3736c48` (`D:\build\...` and `/var/lib/ci/...` exercise the same
parse without matching the pattern). Both rounds' gate lists were incomplete; this is the full one the CI job runs.

`scripts/bundle_corpus.py` remains the one file outside the brief's four — it now owns the whole
ancestor walk, which is the M1 fix.

---

# Round 1 — join evidence to a workbook by LUID

`Fixes #450`

Both migration gates discard oracle evidence that exists — in **opposite directions** — from one
cause: each had invented its own key for *"whose workbook is this render of"*.

| gate | unit side | record side | measured |
|---|---|---|---|
| `check_reference_readiness` (entry) | artifact stem; LUID only when `origin.match == "sha256"` | `workbook_luid` / `workbook_name` | LUID discarded whenever the bytes differ → `HR_Dashboard` compared against `HR Dashboard` → **23 attributable renders thrown away, 0/7 pages ready**. Fail-**closed**. |
| `check_unit` (exit) | artifact stems | `record["workbook"]` — **a key no producer writes** | every record ownerless (**360 of 360**), counted in `unattributed` and then **falling through to `admissible`** → the foreign-workbook guard had never once fired. Fail-**open**. |

## Invariant and direction

- **fail-open (blocks merge)** — a *foreign* workbook's render certifies a page in this unit, or a
  record whose workbook cannot be established is admitted.
- **fail-closed (an issue, not a blocker)** — a render that legitimately belongs to this unit is
  refused, so the page reports BLIND.

Both were live. **Neither was traded for the other**, and the controls below assert the
discrimination rather than the total.

## The fix — one join, in the module both gates already share

`object_identity` gains the workbook half of identity:

```python
WorkbookIdentity(luid, name, sha256).attribute(record) -> Attribution(route, detail, axis)
```

- Compares on the **strongest axis BOTH sides carry**: `sha256` → `luid` → `name`.
- **That axis is decisive.** A mismatch returns `foreign` and never falls through to a weaker one —
  otherwise a foreign workbook sharing a display name is admitted *after* its LUID has said no. That
  fall-through is the identity-loss join this module exists to make unrepresentable.
- **The name axis is exact.** Two projects may hold workbooks with the same name; that is the
  ambiguity `_runs/<NNN>-<slug>/` numbering exists to avoid.
- **No shared axis is `unknown`** — *cannot establish*, never "belongs to this unit".
- `Attribution.__bool__` **raises**: `route` is a non-empty string for a refusal too, so
  `if unit.attribute(record):` would admit a foreign render. `.admitted` is the only reader.

Unit-side LUID, offline, from two claims that must agree (`agreed_luid`; two identities that
contradict each other are *less* evidence than none):

1. `source-provenance.json` — trusted when `matched_by == "luid"` **or** `match == "sha256"`.
   ⚠️ The old rule read `match` alone. `stamp_tableau_provenance.find_origin` documents the two as
   *"two independent axes: a LUID match with `name_only` means 'this is provably the same item on the
   site, and it has changed since we harvested it'"*. Reading the **revision** axis as the
   **identity** axis bought nothing: the fallback (artifact stem vs display name) is weaker on
   identity *and* carries no revision guarantee. On the reference estate **48 of 78** inputs are
   `matched_by: luid` + `match: name_only` — the common case, not an edge one.
2. the asset filename's `<luid>_` prefix (`harvest_estate_assets.py`'s own record of what it
   downloaded).

`Evidence.is_for` is **deleted**, not kept as a delegate — see "what review found", below.

### Second fix: evidence discovery prefers the package — in BOTH gates

`_default_dirs` stops walking up when a `package-manifest.json` sits beside the target (#451). A
package assembled inside a run directory saw its own `oracle/` **and** the run's flat capture two
levels up; every view matched twice and the gate refused the pair as *"2 records share this name once
normalized"* — every page went ready → unverifiable, silently, making packaging strictly worse than
not packaging. The walk-up is **unchanged** for a non-packaged unit and is pinned by a paired control
in the same run.

⚠️ **The brief scoped this to the entry gate; enumerating the closed surface found the identical
defect in the exit gate, and it is fixed here too.** `check_unit._oracle_dirs` walks up to
`target.parent` for the same reason, so a package at `_runs/<run>/<unit>/` beside the run capture
read both manifests. Measured on a synthetic package **before** fixing it: every page refused as
*"2 producer records are named 'Revenue'"* — **0 visual coverage**, silently. Fixing it locally a
second time is exactly the 66%-recurrence pattern
[`docs/review-throughput-postmortem.md`](docs/review-throughput-postmortem.md) measures, so the rule
lives **once**, in `bundle_corpus.is_self_contained` — the module whose own header says the
`check_*` gates *"should not keep separate copies of the same filesystem-discovery rules"*. That is
the one file outside the brief's four that this PR touches, and it is a **+30-line addition**, no
behaviour change to its existing helpers.

## Controls — all four, same run, on `_runs/407-dryrun-gates` (67 units, 313 usable of 360 views)

| control | before | after |
|---|---|---|
| **positive** — estate `pages_ready` | 42 | **167** |
| **positive** — `HR_Dashboard` | **0/7 ready**, 7 blind | **4/7 ready**; the 3 remaining blind are `page_status: dropped_unexplained` (a conversion gap, not an evidence gap) |
| **negative** — a foreign unit certified by HR's renders (`check_unit`) | **PASS, 7/7 visual + numeric** | **NOT_CHECKED, 0/7** |
| **negative** — refusals counted, not dropped | `foreign=0`, `unattributed=360` | `foreign=12851`, `unattributed=0`, `luid=295` |
| **fail-closed on unknown** | admitted (`admitted_evidence` 360 of 360) | refused, `admitted_evidence` 0, counted in `unattributed` |
| **no manufactured coverage** — `pages_blind` | 178 | **49, never 0** — 46 records have no render leg at all (a failed `/data` leg suppressed the image leg) |

Page-status cross-tab, which is what shows nothing was manufactured:

| | before | after |
|---|---|---|
| `emitted` + ready | 39 | 164 |
| `emitted` + blind | 158 | 29 |
| `emitted` + unverifiable | 0 | 4 |
| `dropped_unexplained` + blind | 20 | **20 (unchanged)** |
| `dropped_explained` + ready | 3 | 3 |

⚠️ The 4 new `unverifiable` are **fail-closed and correct**, not a regression: two workbooks in this
estate are the same content published twice, so their renders are **byte-identical**
(`sha256 a9f7a76d…` and `c5cde58b…` each appear under two view LUIDs) and cross-unit exclusivity
refuses them. They were invisible before only because those pages had no evidence at all.

## Closed surface — N = 21

Every consumer, phase and identity-loss join that can affect workbook attribution. Its absence is why
this defect existed in two places.

**Producers (6)** — `capture_tableau_oracle.py` (`views[].workbook_luid`/`workbook_name`) ·
`capture_tableau_reference.py` (`source_workbook_sha256`, **no workbook name at all**) ·
`stamp_tableau_provenance.py` (`origin.{workbook_luid,matched_by,match,same_name_count}`,
`input.sha256`) · `harvest_estate_assets.py` (asset filename `<luid>_<name>`) · engine handover slice
(`workbook.source_id`, `workbook.name`) · engine artifact stems (`<Name>.Report`,
`<Name>.SemanticModel`).

**Entry gate (7)** — `_provenance_luid` (unit LUID) · `resolve_source` (which asset is this unit's) ·
`_identify` (unit identity) · `_reference_states` (record sha) · `_oracle_record` (record luid+name) ·
`Evidence.attribution` (**the join**) · `_scope_evidence` (admit/refuse + census).

**Exit gate (5)** — `_reference_oracles` (record sha + `unit_local` location) ·
`_oracle_capture_oracles` (record luid+name) · `_unit_workbook_identities` (unit identity) ·
`_admissible_oracle_records` (**the join**) · `_lossy_workbook_rescue` (the name-only fallback).

**Location-based scope, which precedes every identity join (3)** —
`check_reference_readiness._default_dirs` (walks up 2) · `check_unit._oracle_dirs` (walks up 1) ·
`check_unit._reference_dirs` (never walks up — which is what makes a unit-local reference manifest
attributable by location). The first two now share `bundle_corpus.is_self_contained`.

**Independent second guard (1)** — `_enforce_exclusivity`: one render may not certify pages in two
units, keyed on the verified render digest. Unchanged here, and it is what catches the byte-identical
pair above.

### Residuals, named

- ⚠️ **`package_unit.py` (PR #451) implements the same rule a second time** — its `workbook_identity()`
  resolves a unit LUID from `source-provenance.json` cross-checked against the asset filename prefix,
  which is exactly `_provenance_luid`'s rule. A sibling agent owns that file right now, so converging
  it on `object_identity.WorkbookIdentity` is deliberately **not** in this PR. Filing a follow-up is
  the right move; two implementations of one identity rule is the shape this PR exists to remove.
- **`check_unit._bindable_workbooks` / `_unit_workbook_keys`** join a *handover slice* to a unit, not
  a render to a workbook. Same class, different artifact; already two-sided-uniqueness guarded and
  untouched.
- **`group_oracle_by_workbook.py`, `assess_estate.py`, `tableau_lineage.py`, `build_reconcile_items.py`**
  read `workbook_luid`/`workbook_name` for splitting, inventory and reconciliation. No gate verdict
  rides on them.
- **A unit whose LUID cannot be established** (no provenance entry, no harvest filename prefix) still
  joins on the exact display name, and in `check_unit` on the uniqueness-guarded lossy name. Both are
  disclosed (`evidence_attributed.name`; `known_gap_caveats`). This is the fail-closed-side residual:
  such a unit reports BLIND for a render whose display name differs from its stem.
- **`check_unit` reads the handover slice for the unit LUID**, so a raw `<bundle>/pbip/<Unit>` target
  (no `handover/` inside it) gets no LUID and falls to the name routes. A package or a migration unit
  has it. Not a regression — it had no LUID at all before.

## Can each test reach the production path?

21 targeted mutations, each applied to the real source file, each with a **control test that must
still pass** so a broken import cannot score as a kill. **21/21 killed** — and the first run was
15/18 (then 17/21 once the exit-gate fix was added), which is the evidence that the harness is not rubber-stamping:

- one mutation was **UNAPPLIED** (a stale pattern after `ruff format`) — fixed, then killed;
- one **SURVIVED** because my own mutation was a no-op (`luid = None or agreed_luid(...)`) — fixed,
  then killed;
- one **SURVIVED** genuinely: deleting the explicit `continue` after `unattributed += 1` changed no
  verdict, because the lossy rescue refuses an unestablished record too. Rather than accept a
  double-guarded line as "proved", the test now asserts the observable contract — an unattributed
  record is counted **once**, in `unattributed`, and **not** as `foreign` — which kills it.

Both committed mutation runners are green against the new code: `mutate_check_unit.py` **91/91
caught** with its negative control surviving, `mutation_reference_readiness.py` **69/69 matched (137
anchor/control checks)**.

## What review found that the brief did not — two corrections

1. **The brief says `check_reference_readiness._provenance_luid()` "deliberately returns no trusted
   LUID" because provenance records `match: "name_only"`. That is the symptom; the cause is one field
   narrower.** The record also carries `matched_by: "luid"`, and `find_origin` documents the two as
   independent axes. So the fix is *not* "trust `name_only`" — an origin with **no `matched_by`** and
   unconfirmed bytes may be a name collision (`find_origin` falls back to `name` and records
   `same_name_count`), and it is still refused. The existing test asserting that
   (`test_a_name_only_provenance_luid_is_not_trusted`) **still passes unchanged**; its docstring is
   re-aimed and a discriminating twin added.

2. **Deleting `Evidence.is_for` was forced by a measurement, not by taste.** It was first kept as a
   one-line delegate to `attribution()`. `mutation_reference_readiness.py` then reported two mutants
   as CAUGHT while patching a method the gate no longer called — i.e. two guards had become vacuous
   without any test noticing. Both mutants are re-aimed (one at `Evidence.attribution`, one at
   `object_identity._axis_equal`, so the lossy-name mutant now tests the join **both** gates use), and
   the boolean is gone. A second reader of one contract is where two answers diverge, which is the
   whole shape of #450.

One existing test was **reversed on purpose**:
`test_oracle_evidence_declaring_no_workbook_is_admitted_but_flagged` asserted the fail-open, on the
premise *"real oracle manifests carry no workbook field, so refusing them outright would reject every
capture in the estate."* The premise was false in the only way that mattered — a real manifest carries
`workbook_luid` and `workbook_name`; it carries no `workbook`, which is a defect in the **reader**. It
is now `test_a_record_whose_workbook_cannot_be_established_certifies_nothing`, with a positive twin
proving the fields a real capture writes ARE read.

`tests/test_shared_identity_pin.py` is re-taken deliberately: both PRs that shared
`object_identity.py` have merged, so `origin/master` is the shared truth (as
`verify_shared_identity_pin.PREFERRED_REFS` already said). The pin's job is now to make a shared-surface
edit a **reviewed** event; a new `PIN_PROVENANCE` records what moved and why, and a test asserts the
provenance names a reason rather than only a SHA.

## Gates — by exit code

| gate | exit |
|---|---|
| `pytest tests/` | **1** — `3 failed, 4064 passed, 6 skipped`; the 3 are the pre-existing `test_upstream_repro_pins.py` engine-drift failures, which import only `engine_source` and touch none of these files. CI (`checks` + `windows-bundle`) is **green**. |
| `ruff format --check .` | **0** |
| `ruff check .` | **0** |
| `pylint scripts` | **0** (⚠️ it printed `10.00/10` while exiting **24** before three real findings were fixed: two `too-many-locals`, two `line-too-long`) |
| `pylint conftest.py` | **0** |
| `sync_agent_conventions.py --check` | **0** |
| `check_identity_normalization.py` | **0** — the lossy comparison stays quarantined |

## Scope

`scripts/{object_identity,reference_evidence,check_reference_readiness,check_unit}.py` and their
tests, **plus a +30-line addition to `scripts/bundle_corpus.py`** (see the second fix above: putting
the package marker anywhere else would have duplicated one discovery rule across two gates, which
is the defect class this PR removes). `scripts/README.md`, `docs/agent-capability-wiring.md`, `docs/operator-runbook.md`,
`scripts/package_unit.py` and `tests/test_package_unit*.py` are untouched (PR #451 owns them).

### DIFF
--- a/scripts/bundle_corpus.py
+++ b/scripts/bundle_corpus.py
@@ -11,6 +11,34 @@ from __future__ import annotations
 
 from pathlib import Path
 
+#: A self-contained handover package writes this beside the unit (`scripts/package_unit.py`, #446).
+PACKAGE_MARKER = "package-manifest.json"
+
+
+def is_self_contained(target: Path) -> bool:
+    """Whether ``target`` declares that it carries its own evidence and must inherit none.
+
+    ⚠️ **This is what stops an evidence walk-up, in BOTH gates, and it has to be one rule.** Every
+    gate here looks for `reference/`/`oracle/` beside the target AND beside its ancestors, and
+    **unions** the hits - which is right for an un-packaged unit under `<bundle>/pbip/<Unit>/`, whose
+    capture lives further up. `package_unit.py` writes a unit-scoped `oracle/oracle-manifest.json`
+    holding THIS unit's views with rewritten paths, so a package assembled INSIDE a run directory
+    matches every view twice and both gates then refuse the pair as an ambiguity:
+
+    * `check_reference_readiness` reports *"2 records share this name once normalized"* and takes
+      every page from ready to **unverifiable** (issue #451);
+    * `check_unit` reports *"2 producer records are named X"* and reports **0 visual coverage** -
+      measured on a synthetic package, the same defect one gate along.
+
+    Both are silent, and both make packaging strictly WORSE than not packaging. The marker is the
+    package's own declaration that it is complete, so it is taken at its word.
+
+    ⚠️ Deliberately NOT "stop when the target has its own copy of this directory": a package that
+    OMITTED a render because it could not attribute it would then pick that render up from the
+    ancestor, undoing a fail-closed packaging decision at the consumer.
+    """
+    return (target / PACKAGE_MARKER).is_file()
+
 
 def shipping_reports(root: Path) -> list[Path]:
     """Return `.Report` folders that ship under ``root``.
--- a/scripts/bundle_corpus.py
+++ b/scripts/bundle_corpus.py
@@ -9,6 +9,7 @@ separate copies of the same filesystem-discovery rules. This module is the singl
 
 from __future__ import annotations
 
+from collections.abc import Sequence
 from pathlib import Path
 
 #: A self-contained handover package writes this beside the unit (`scripts/package_unit.py`, #446).
@@ -73,3 +74,52 @@ def shipping_models(root: Path, *, include_standalone: bool = False) -> list[Pat
         if standalone.is_dir():
             models.update(path.resolve() for path in standalone.rglob("*.SemanticModel") if path.is_dir())
     return sorted(models, key=str)
+
+
+#: How far above a target an evidence directory may live.
+#:
+#: THREE, from the canonical layout rather than from taste: a capture is written to
+#: ``_runs/<NNN>-<slug>/oracle/`` while an un-packaged unit sits at
+#: ``_runs/<NNN>-<slug>/bundle/pbip/<Unit>/`` - exactly three ancestors below it. Round-1 review of
+#: PR #454 measured the exit gate stopping at ONE, so that unit could not see the run's capture at
+#: all; stopping at two still misses it for the ordinary engine-bundle shape.
+#:
+#: Widening discovery is safe in the direction that matters: every record found this way still has to
+#: pass the identity join, so a foreign capture pulled in from a shared ancestor is refused and
+#: counted, and two indistinguishable records make the page UNVERIFIABLE. Discovery adds candidates;
+#: it never admits one.
+ANCESTOR_LEVELS = 3
+
+
+def evidence_dirs(target: Path, names: Sequence[str], *, also: Sequence[Path] = ()) -> list[Path]:
+    """Existing evidence directories for ``target``: beside it, then up to two ancestors.
+
+    WARNING: **The whole walk lives here, not just where it stops.** Round-1 review of PR #454:
+    centralising only the *stop* condition left the two gates disagreeing about the *search* -
+    `check_reference_readiness` looked two levels up while `check_unit` looked one, so a non-packaged
+    unit at `<bundle>/pbip/<Unit>/` could not inherit the run's flat capture at all in the exit gate.
+    A shared rule that covers half the behaviour is two rules wearing one name.
+
+    The ancestor portion is skipped for a self-contained package (:func:`is_self_contained`); the
+    target itself, and any ``also`` root a caller adds, are always searched. Results keep discovery
+    order, are de-duplicated by resolved path, and are returned unresolved so a caller still sees the
+    spelling it passed in.
+    """
+    roots = [target, *also]
+    if not is_self_contained(target):
+        ancestor = target
+        for _ in range(ANCESTOR_LEVELS):
+            ancestor = ancestor.parent
+            roots.append(ancestor)
+    found: list[Path] = []
+    seen: set[Path] = set()
+    for root in roots:
+        for name in names:
+            candidate = root / name
+            if not candidate.is_dir():
+                continue
+            resolved = candidate.resolve()
+            if resolved not in seen:
+                seen.add(resolved)
+                found.append(candidate)
+    return found
--- a/scripts/check_reference_readiness.py
+++ b/scripts/check_reference_readiness.py
@@ -169,6 +169,11 @@ class UnitResult:
     report_dir: str | None = None
     source: str | None = None
     pages: list[dict[str, Any]] = field(default_factory=list)
+    #: How every candidate render was attributed to (or refused for) this unit, keyed by
+    #: `object_identity`'s route vocabulary. A refusal that is not counted is indistinguishable from
+    #: a render that was never captured, which is exactly how issue #450 stayed invisible: a fix that
+    #: only makes `pages_ready` go up cannot be told apart from one that deleted the guard.
+    attribution: dict[str, int] = field(default_factory=dict)
 
 
 def engine_page_id(text: str) -> str:
@@ -488,32 +493,49 @@ def _handover(root: Path, unit: str) -> dict[str, Any] | None:
     return json_object(root / "handover" / f"{unit}.json")
 
 
-def _provenance_luid(root: Path, source_sha: str) -> str | None:
-    """The published workbook LUID for this source - ONLY when provenance is byte-confirmed.
+#: Provenance ``origin`` shapes that ESTABLISH which published workbook a local file is. See
+#: :func:`_provenance_luid` - deliberately distinct from ``origin.match``, which answers a revision
+#: question and was read as if it were this one (issue #450).
+IDENTITY_MATCHED_BY = frozenset({"luid"})
 
-    Round-2 finding 3: this returned a LUID without consulting `origin.match`.
-    `stamp_tableau_provenance.py:191-192` records `"sha256"` when local and server bytes agree and
-    `"name_only"` when they DIFFER, and its own docstring says figures will not reproduce in that
-    case - yet that LUID was making server oracle evidence ready. The repo's provenance is 26
-    `sha256` / 15 `name_only` / 6 unmatched, so trusting it blindly is the common case.
 
-    Both halves are required: the stamped input hash must be THIS file, and the server comparison
-    must have agreed. Anything else leaves the name as the only usable identity, which `is_for`
-    falls back to.
+def _provenance_luid(root: Path, source_sha: str, source: Path | None = None) -> str | None:
+    """The published workbook LUID for this source, from every offline route that ESTABLISHES one.
+
+    Two routes, and they must agree:
+
+    * ``source-provenance.json``'s ``origin.workbook_luid`` for the entry whose stamped input hash is
+      THIS file - trusted when ``matched_by == "luid"`` (the harvested filename's LUID was found on
+      the site, so it is provably that item) **or** when ``match == "sha256"`` (the bytes are
+      identical to the site copy, which confirms identity by content). A ``matched_by`` of ``name``
+      or ``sanitized_name`` with differing bytes is a name collision, not an identity.
+    * the asset filename's own ``<luid>_`` prefix, which is `harvest_estate_assets.py`'s record of
+      what it downloaded.
+
+    ⚠️ This used to require ``origin.match == "sha256"`` and nothing else, which read the REVISION
+    axis as if it were the IDENTITY axis. ``stamp_tableau_provenance.find_origin`` documents them as
+    "two independent axes"; on the reference estate 48 of 78 inputs are ``name_only`` while matched
+    ``by luid``, so the common case discarded a proven LUID and fell back to comparing the artifact
+    stem ``HR_Dashboard`` against the published name ``HR Dashboard``. That fallback is weaker on
+    identity AND carries no revision guarantee, so the trade bought nothing.
+
+    Disagreement between the two routes returns None: two identities that contradict each other are
+    less evidence than none, and the name axis then has to carry it.
     """
+    stamped = oid.harvest_luid(source.stem if source is not None else None)
     payload = json_object(root / "source-provenance.json")
     for record in (payload or {}).get("inputs") or []:
         if not isinstance(record, dict):
             continue
-        stamped = record.get("input") if isinstance(record.get("input"), dict) else {}
+        stamped_input = record.get("input") if isinstance(record.get("input"), dict) else {}
         origin = record.get("origin") if isinstance(record.get("origin"), dict) else {}
-        if stamped.get("sha256") != source_sha:
+        if stamped_input.get("sha256") != source_sha:
             continue
-        if origin.get("match") != "sha256":
-            return None
+        if origin.get("matched_by") not in IDENTITY_MATCHED_BY and origin.get("match") != "sha256":
+            return oid.agreed_luid(stamped)
         luid = origin.get("workbook_luid")
-        return str(luid) if isinstance(luid, str) and luid else None
-    return None
+        return oid.agreed_luid(stamped, luid if isinstance(luid, str) else None)
+    return oid.agreed_luid(stamped)
 
 
 def resolve_source(root: Path, unit: str, handover: dict[str, Any] | None, explicit: Path | None) -> Path | None:
@@ -550,10 +572,29 @@ def resolve_source(root: Path, unit: str, handover: dict[str, Any] | None, expli
     return None
 
 
+#: A self-contained handover package writes this beside the unit (`scripts/package_unit.py`, #446).
+#: Its presence is the marker that a target owns its evidence and must not inherit an ancestor's.
+PACKAGE_MARKER = "package-manifest.json"
+
+
 def _default_dirs(root: Path, name: str) -> list[Path]:
-    """Conventional evidence locations, mirroring `check_unit.py:830-839`."""
+    """Conventional evidence locations, mirroring `check_unit.py:830-839`.
+
+    The walk-up is a UNION, not a fallback, and that is deliberate: an un-packaged unit under
+    `<bundle>/pbip/<Unit>/` finds the run's flat capture two levels up, which is the ordinary case and
+    must keep working.
+
+    ⚠️ **A self-contained package stops the walk.** `package_unit.py` writes a unit-scoped
+    `oracle/oracle-manifest.json` holding THIS unit's views with rewritten paths; a package assembled
+    INSIDE a run directory therefore sees its own copy AND the run's flat capture two levels up, and
+    every view matches twice. The gate then refuses the pair as "2 records share this name once
+    normalized" and every page goes ready -> unverifiable - strictly worse than not packaging, and
+    silent. A `package-manifest.json` beside the target is that package's own declaration that it is
+    complete, so it is taken at its word and no ancestor is consulted.
+    """
+    bases = [root] if (root / PACKAGE_MARKER).is_file() else [root, root.parent, root.parent.parent]
     seen: list[Path] = []
-    for candidate in (root / name, root.parent / name, root.parent.parent / name):
+    for candidate in (base / name for base in bases):
         if candidate.is_dir() and candidate.resolve() not in {path.resolve() for path in seen}:
             seen.append(candidate)
     return seen
@@ -576,7 +617,7 @@ def _identify(root: Path, unit: str, source: Path) -> UnitIdentity | None:
     if digest is None:
         return None
     return UnitIdentity(
-        name=unit, source_path=source, source_sha256=digest, workbook_luid=_provenance_luid(root, digest)
+        name=unit, source_path=source, source_sha256=digest, workbook_luid=_provenance_luid(root, digest, source)
     )
 
 
@@ -665,7 +706,9 @@ def _datasource_only(unit: str, report_dir: Path, engine_report: dict[str, Any] 
     )
 
 
-def _readiness_result(unit: str, report_dir: Path, source: Path, rows: list[dict[str, Any]]) -> UnitResult:
+def _readiness_result(
+    unit: str, report_dir: Path, source: Path, rows: list[dict[str, Any]], attribution: dict[str, int]
+) -> UnitResult:
     """Fold per-page rows into the unit verdict."""
     findings = [row for row in rows if row["readiness"] != READY]
     return UnitResult(
@@ -675,6 +718,7 @@ def _readiness_result(unit: str, report_dir: Path, source: Path, rows: list[dict
         report_dir=str(report_dir),
         source=str(source),
         pages=rows,
+        attribution=attribution,
     )
 
 
@@ -725,9 +769,39 @@ def assess_unit(  # pylint: disable=too-many-arguments,too-many-positional-argum
     if isinstance(emitted, UnitResult):
         return emitted
 
-    scoped = [item for item in evidence if item.is_for(identity)]
-    rows = _page_rows(objects, emitted, drop_explanations(handover), scoped, require_validation_grade)
-    return _readiness_result(unit, report_dir, source, rows)
+    scoped = _scope_evidence(evidence, identity)
+    rows = _page_rows(objects, emitted, drop_explanations(handover), scoped.admitted, require_validation_grade)
+    return _readiness_result(unit, report_dir, source, rows, scoped.census)
+
+
+@dataclass(frozen=True)
+class ScopedEvidence:
+    """One unit's share of the candidate renders, WITH the census of what was refused.
+
+    The two travel together on purpose. A refusal that is dropped instead of counted is
+    indistinguishable from a render nobody ever captured, and that ambiguity is exactly how issue
+    #450's inert guard read as a working one for six review rounds.
+    """
+
+    admitted: list[Evidence]
+    census: dict[str, int]
+
+
+def _scope_evidence(evidence: list[Evidence], identity: UnitIdentity) -> ScopedEvidence:
+    """Attribute every candidate render to ``identity``, keeping the ones it may certify.
+
+    Every render is attributed, admitted or not, and the census records which axis decided - so the
+    negative control ("a different workbook's render is refused, AND counted") is assertable and a
+    change here cannot pass as a fix while actually deleting the guard.
+    """
+    census = dict.fromkeys((*oid.WB_ROUTES, oid.WB_FOREIGN, oid.WB_UNKNOWN), 0)
+    admitted = []
+    for item in evidence:
+        verdict = item.attribution(identity)
+        census[verdict.route] = census.get(verdict.route, 0) + 1
+        if verdict.admitted:
+            admitted.append(item)
+    return ScopedEvidence(admitted=admitted, census=census)
 
 
 def _units_without_reports(engine_report: dict[str, Any] | None, reports: list[Path]) -> list[UnitResult]:
@@ -858,6 +932,10 @@ def _merge(
         "evidence_records": len(evidence),
         "evidence_untyped": sum(1 for item in evidence if item.kind == KIND_UNKNOWN),
         "evidence_untyped_names": sorted({item.name for item in evidence if item.kind == KIND_UNKNOWN}),
+        # The route census, summed over units. Admissions are per-route so "admitted because a LUID
+        # matched" is never confused with "admitted because two display names happened to agree", and
+        # refusals are counted so a foreign render is visibly refused rather than silently absent.
+        "evidence_attributed": _census(units),
         "evidence_rejected": [
             {"name": item.name, "origin": item.origin, "path": item.path, "reason": item.reason} for item in rejected
         ],
@@ -873,10 +951,20 @@ def _merge(
                 "report": unit.report_dir,
                 "source": unit.source,
                 "pages": unit.pages,
+                "attribution": unit.attribution,
             }
             for unit in units
         ],
     }
+
+
+def _census(units: list[UnitResult]) -> dict[str, int]:
+    """Every unit's attribution census, summed. Keys are always present, so a zero is visible."""
+    total = dict.fromkeys((*oid.WB_ROUTES, oid.WB_FOREIGN, oid.WB_UNKNOWN), 0)
+    for unit in units:
+        for route, count in unit.attribution.items():
+            total[route] = total.get(route, 0) + count
+    return total
 
 
 GRADE_CEILING_NOTE = (
@@ -915,6 +1003,7 @@ def render(report: dict[str, Any], *, verbose: bool = False) -> str:
             f"  [UNTYPED EVIDENCE] {report['evidence_untyped']} render(s) carry no object type "
             f"({', '.join(report['evidence_untyped_names'][:4])}) - {MANUAL_KIND_HINT}"
         )
+    lines.extend(_render_attribution(report))
     if report["status"] == STATUS_CANNOT_ESTABLISH:
         lines.append(
             "  CANNOT_ESTABLISH is NOT a pass: this gate formed no opinion, so an agent starting "
@@ -925,6 +1014,24 @@ def render(report: dict[str, Any], *, verbose: bool = False) -> str:
     if not verbose and report["pages_ready"]:
         lines.append("  Run with --verbose to list the pages that ARE ready and their grades.")
     return "\n".join(lines)
+
+
+def _render_attribution(report: dict[str, Any]) -> list[str]:
+    """Say how evidence was attributed, INCLUDING the refusals.
+
+    Reported unconditionally when any candidate render was seen: a silent refusal is
+    indistinguishable from a render that was never captured, and that ambiguity is what let issue
+    #450's inert guard look like a working one for six review rounds.
+    """
+    census = report.get("evidence_attributed") or {}
+    if not sum(census.values()):
+        return []
+    admitted = ", ".join(f"{route}={census.get(route, 0)}" for route in oid.WB_ROUTES)
+    return [
+        f"  [ATTRIBUTION] admitted by {admitted}; refused {census.get(oid.WB_FOREIGN, 0)} as another "
+        f"workbook's and {census.get(oid.WB_UNKNOWN, 0)} with no shared identity axis "
+        "(counts are per unit-and-record, so one shared capture is weighed against every unit)."
+    ]
 
 
 def _render_page(page: dict[str, Any]) -> str:
--- a/scripts/check_reference_readiness.py
+++ b/scripts/check_reference_readiness.py
@@ -459,11 +459,12 @@ def _page_row(  # pylint: disable=too-many-arguments,too-many-positional-argumen
             }
         )
         return row
-    stale = _stale_note(obj, scoped)
-    if stale is not None:
-        # ⚠️ No `render_key`: a stale render certifies nothing, so it must not contest a legitimate
+    refusal = _refusal_note(obj, scoped)
+    if refusal is not None:
+        # ⚠️ No `render_key`: a refused render certifies nothing, so it must not contest a legitimate
         # render of the same bytes in another unit and drag that page down with it.
-        row.update({"evidence": "unverifiable", "matched_by": stale, "readiness": UNVERIFIABLE})
+        row.update({"evidence": "unverifiable", "matched_by": refusal[1], "readiness": UNVERIFIABLE})
+        row["revision"] = REVISION_UNCONFIRMED if refusal[0] == oid.WB_UNCONFIRMED else row.get("revision")
         return row
     if name_only:
         scopes = sorted({item.kind for item in name_only})
@@ -479,23 +480,49 @@ def _page_row(  # pylint: disable=too-many-arguments,too-many-positional-argumen
     return row
 
 
-def _stale_note(obj: SourceObject, scoped: ScopedEvidence) -> str | None:
-    """Why a render of this object exists but is of a DIFFERENT build of this workbook, or None.
+#: What to say when a refused render is the only one this page has, per route. Ordered strongest
+#: statement first: a record we could compare and reject says more than one we could not compare.
+REFUSAL_NOTES = {
+    oid.WB_FOREIGN: "a render of ANOTHER workbook carries this name - refused, not credited",
+    oid.WB_STALE: (
+        "STALE: the render is of the same workbook at a DIFFERENT build - the local source bytes "
+        "differ from a reproducible content digest of the site copy, so it cannot certify the "
+        "migrated revision"
+    ),
+    oid.WB_UNCONFIRMED: (
+        "REVISION NOT ESTABLISHED: the render is tied to this workbook by LUID, which says WHICH "
+        "workbook a picture is of and never WHICH BUILD, and this unit's provenance carries no "
+        "comparable revision key - re-stamp with scripts/stamp_tableau_provenance.py"
+    ),
+    oid.WB_NAME: (
+        "NAME ONLY: the render declares no machine identity at all, and a display name is not unique "
+        "across projects - it may inform discovery, never certification"
+    ),
+    oid.WB_UNKNOWN: "the render declares an identity this unit cannot answer - refused, not credited",
+}
 
-    Kept distinct from BLIND on purpose: "nobody captured this page" and "the capture is of another
-    revision" call for different operator actions - capture it, versus re-harvest or re-capture
-    against the migrated build.
+
+def _refusal_note(obj: SourceObject, scoped: ScopedEvidence) -> tuple[str, str] | None:
+    """``(route, note)`` when a refused render of THIS unit's workbook matches this object, else None.
+
+    ⚠️ Only :data:`WB_STALE` and :data:`WB_UNCONFIRMED` are consulted, and the restriction is
+    load-bearing. Those two are records whose LUID matched this unit - unambiguously a render of this
+    page, of this workbook, at a build we cannot vouch for - so reporting UNVERIFIABLE rather than
+    BLIND tells the operator something true and actionable.
+
+    A ``foreign``, ``name``-only or ``unknown`` record is NOT known to be a render of this page.
+    Letting one turn a page from BLIND to UNVERIFIABLE would credit another workbook's capture with
+    partial coverage here, which is the cross-workbook leak this whole change exists to close, and it
+    would quietly erode the blind count that the estate's 46-render floor is measured against.
     """
-    if not scoped.stale:
-        return None
-    match, _ = match_evidence(obj, scoped.stale)
-    if not isinstance(match, Evidence):
-        return None
-    return (
-        f"STALE: {match.origin}:{match.provider or 'unknown'} render is of the same workbook at a "
-        "DIFFERENT build - the local source bytes differ from a REPRODUCIBLE download of the site "
-        "copy, so this cannot certify the migrated revision"
-    )
+    for route in (oid.WB_STALE, oid.WB_UNCONFIRMED):
+        candidates = scoped.refused.get(route) or []
+        if not candidates:
+            continue
+        match, _ = match_evidence(obj, candidates)
+        if isinstance(match, Evidence):
+            return route, REFUSAL_NOTES[route]
+    return None
 
 
 def _engine_report(root: Path) -> dict[str, Any] | None:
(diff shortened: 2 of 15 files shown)


Guust-Franssens:
## Independent end-to-end measurement on cold-run-1 data (not this PR's fixtures)

I ran a controlled A/B against **real estate data** rather than the branch's own test fixtures, to
check whether this PR's LUID join actually changes the verdict an agent would receive on a packaged
unit. It does, decisively.

### Method

- **Input**: one unit (`HR_Dashboard`) packaged by `scripts/package_unit.py` from master `673e54f`
  (the #451 packaging that merged), out of the cold-run-1 fixture `_runs/407-dryrun-gates`
  (48 workbooks / 360 views, live Tableau Cloud pull).
- Packaging itself exited **0**: `OK HR_Dashboard [workbook] - 23 oracle object(s) via workbook_luid`.
- **The two gate runs consumed the byte-identical package directory.** The only variable is which
  copy of `check_reference_readiness.py` executed.
- Gate A = master `673e54f`. Gate B = this branch at **`3736c48`**.

⚠️ `3736c48` is **not** the branch tip — `ca4d395` is committed on top of it and there is further
uncommitted work in the author's worktree. So everything below is a **lower bound** on this PR's
behaviour, measured on a clean checkout with `git status --porcelain` empty.

### Result

| gate code | ready | blind | attribution reported |
|---|---:|---:|---|
| master `673e54f` | **0 / 7** | 7 | — none |
| this PR `3736c48` | **4 / 7** | 3 | yes (below) |

Master's verdict on a correctly packaged unit, verbatim:

```
- dashboard 'HR | Details'  -> page-HRDetails83013a67: BLIND - no reference render, ...
- dashboard 'HR | Summary'  -> page-HRSummarydecff6a2: BLIND - no reference render, ...
- worksheet 'Age'           -> page-ws-Aged40ea549:    BLIND - no reference render, ...
- worksheet 'test'          -> page-ws-testdd6499f7:   BLIND - no reference render, ...
```

Those four are **not** blind. The package physically contains, for exactly those objects:

```
oracle/dashboard/images/HR _ Details.svg   (793,795 B)
oracle/dashboard/images/HR _ Summary.svg   (912,014 B)
oracle/worksheet/images/Age.svg            ( 72,639 B)
oracle/worksheet/images/test.svg           (21,025,388 B)
```

So on master the evidence is present on disk and the gate cannot see it. **That is the whole of
#450, reproduced end to end on cold-estate data**, and it means an agent starting from a packaged
unit today would be told every page is unfalsifiable.

### What this PR emits instead — and why the extra lines matter as much as the coverage

```
[ATTRIBUTION] admitted by sha256=0, luid=23, name=0; refused 0 as another workbook's,
              0 as a different build and 0 with no identity this unit can answer
              (counts are per unit-and-record, so one shared capture is weighed against every unit).

[REVISION NOT ESTABLISHED] 4 ready page(s) rest on a render tied to this workbook by LUID or name,
              which says WHICH workbook a picture is of and never WHICH BUILD. Only a source-sha
              match, or `source-provenance.json` reporting `match: "sha256"`, proves the capture
              depicts the revision you are migrating.
```

Three things worth calling out, because a 0→4 coverage jump is *also* what deleting the guard looks
like and coverage alone cannot tell those apart:

1. **`name=0`.** The name-based admission route contributed nothing here — the 23 records were
   admitted by LUID. That is the direction blocker B-B asked for.
2. **The refusal counters are reported beside the coverage**, which is the pre-registered cold-run
   pass criterion #3. A zero refusal count on this unit is explainable (single-workbook package, all
   23 records genuinely belong to it), not a silent absence.
3. **The revision caveat is emitted rather than swallowed.** The four ready pages are *not* reported
   as revision-confirmed. That is blocker B-A behaving correctly on real data.

### The second gate also accepts the package with zero flags

`check_unit.py` from the same `3736c48`, same package, **no arguments beyond the path** — exit **4**:

```
UNIT CHECK (all scope): PRECONDITION_FAILED
  stopped after failed precondition: page-parity
  page pairing: 4 rebuilt PBIR page(s) for 7 expected Tableau page(s)  [PRECONDITION_FAILED]
      OMITTED worksheet 'Cities'     [declared-by-engine-unsigned]
      OMITTED worksheet 'Job Titles' [declared-by-engine-unsigned]
      OMITTED worksheet 'States'     [declared-by-engine-unsigned]
      suspected owner: report
      detail: ... the engine declared it produced no visual (manual attention required:
              mark class 'Bar' / shelf layout not supported -> no visual emitted) - that
              explains the omission but does not accept it; rebuild the page or sign a
              page-parity exemption
SUMMARY: blockers=1; compromises=0; compromises_not_evaluated=1; findings_by_owner=report=1;
         ladder=PRECONDITION_FAILED exit=4
```

Note the distinction it draws: *"that explains the omission but does not accept it."* An engine
explanation is not an exemption. That is the correct direction for a gate and I want it on the record
so a later round does not "simplify" it away.

### Consequence for the merge queue

This PR is the **cold-run gate**, and the measurement above is the reason: without it the packaging
that already merged in #451 hands an agent a unit whose every page is BLIND, on data where the
renders demonstrably exist. The remaining review blockers are about tightening admission further,
not about whether the join works — it does, on a real 48-workbook estate.

### What I did NOT verify

- Only **one** unit of 62 was packaged and gated. This is an existence proof, not an estate
  distribution — the estate-wide number is the cold run's job, not this comment's.
- `3736c48`, not the tip. Uncommitted work in the author's worktree is unmeasured here.
- The three `NO PAGE EMITTED` worksheets look like a **deterministic-engine** gap
  (*mark class 'Bar' / shelf layout not supported*), not a gap in this PR. One workbook is an
  anecdote; I have not established how often it happens estate-wide, so I am deliberately not
  filing it upstream yet.

Guust-Franssens:
# Blind review — BLOCKED (2 fail-open findings, both reproduced)

Independent blind review, given only issue #450 and the diff — never the author's reasoning.
Reviewed tree `b9ef38e`, confirmed clean. Elapsed 2192s.

## Blocking

### A. Ambiguous provenance selects the first matching SHA — verdict depends on JSON order

`scripts/reference_evidence.py:198-210`

Two provenance records with the **same source SHA** and **different workbook LUIDs**; only their
order was reversed:

```
AMBIGUOUS_FIRST   ={"status":"READY",   "pages_ready":1,"luid":1,   "exit":0}
AMBIGUOUS_REVERSED={"status":"FINDINGS","pages_ready":0,"foreign":1,"exit":1}
```

The loop returns on the first SHA match without checking for a second, conflicting match. Byte-identical
evidence flips between READY/exit 0 and refused purely on ordering. Ambiguous provenance must be an
explicit *cannot-establish*, never resolved by position.

### B. Contradictory machine identities are admitted as trustworthy

`scripts/object_identity.py:506-518`, `scripts/check_unit.py:1966-1969`, `scripts/reference_evidence.py:642-663`

Evidence whose SHA matched the unit but whose declared LUID belonged to **another workbook**:

```
CONTRADICTORY_ENTRY={"status":"READY","pages_ready":1,
  "attribution":{"sha256":1,"luid":0,"foreign":0},"page":"ready"}
CONTRADICTORY_EXIT ={"status":"PASS","visual_present":1,"numeric_present":1,
  "admitted_evidence":1,"foreign_workbook_evidence":[],"unattributed_evidence":0}
```

`WorkbookIdentity.attribute()` returns as soon as the SHA axis matches and never inspects the
contradicting LUID; the entry gate discards the manifest LUID before building `Evidence`.

> *"Conflicting machine identities are less evidence than none. They must be refused explicitly
> rather than resolved by axis priority."*

Note the asymmetry that the fix must preserve: an axis that is **absent** is not the same as an axis
that **actively contradicts**.

## Non-blocking

- `ruff check scripts tests` exit 1 — unused `OTHER_LUID`, `tests/test_reference_evidence.py:27`
- `pylint scripts` exit 8 — `object_identity.py:721` `R0911` (prints 10.00/10 and still exits non-zero)
- `tests/test_shared_identity_pin.py:79` — stale digest, expected `7642b526…` actual `6895a515…`
- `tests/mutate_check_unit.py` exit 1 — 5 stale mutation snippets, 1 obsolete test-node reference
- `mutation_reference_readiness.py` — uncaught name-normalization mutation; the anchor passed because a
  broad "it refused" assertion cannot tell *which* guard refused

## Mutation campaign

| mutation | result | right reason? |
|---|---|---|
| Reverse provenance iteration order | **NOT CAUGHT** (314 passed) | no — this is Finding A |
| Unanswerable LUID falls through to name admitting | caught (`blind` → `ready`) | yes |
| Ownerless `check_unit` evidence falls through | caught (`NOT_CHECKED` → `PASS`) | yes |
| Stop reading real `workbook_luid` | caught (coverage 1 → 0) | yes |
| Stop reading real `workbook_name` | caught on `name_only_evidence` | yes, diagnostic |
| Disable matching-LUID equality | 2 tests failed `foreign`/zero coverage | yes |
| Revision-key campaign | **12/12 killed with firing probes** | yes |
| Normalize workbook names | **NOT CAUGHT** | no — broad refusal assertion |

## Gates

| command | exit |
|---|---:|
| 4-file pytest | **0** (314 passed) |
| `ruff check scripts tests` | **1** |
| `ruff format --check` | **0** |
| shared-pin/provenance tests | **1** (1 failed, 32 passed) |
| `tests/mutate_revision_key.py` | **0** |
| `tests/mutation_reference_readiness.py` | **1** |
| `tests/mutate_check_unit.py` | **1** |

## Could not verify

- `C:\tfmig\_pkgmaster` held only `fabric/` — no migration specs or reference/oracle metadata — so the
  rumoured 4/7-vs-0/7 ready-count comparison **was not reproducible**. Both branch and master returned
  entry-gate exit **3** for `Admin_Insights_Starter`; `check_unit` exit **1**, oracle coverage
  `NOT_CHECKED`. Treating that earlier figure as unconfirmed rather than as fact.
- Local `master` had moved to `c8b0a58` (merge-base `0fae0cf`); both bases merge cleanly with the tip.

Guust-Franssens:
# Cold run 408 — estate-wide entry-gate baseline, and a retraction

## The retraction (read this first)

Earlier today I told the user, and posted to PR #454 as comment `5519914145`, that master scored
**0/7 reference-ready** on `HR_Dashboard` where PR #454's branch scored **4/7** — and concluded
"#454 is provably the cold-run gate".

**That measurement is not reproducible and must not be cited.**

The blind reviewer of #454 tried to reproduce it and could not: `C:\tfmig\_pkgmaster` "contained only
`fabric/`, without migration specs or reference/oracle metadata", and both branch and master returned
entry-gate **exit 3** for `Admin_Insights_Starter`. It recorded the figure as a rumour rather than a
fact — correctly.

Re-measured today on a *complete* package built by the current `package_unit.py` from the cold-run-408
bundle, master's gate reports **9/10 pages ready** for `Superstore`. So the earlier "0/7" was measuring
**an incomplete package**, not a defective gate.

⚠️ There is a second reason to distrust the old number: if that package genuinely carried no oracle
metadata, then the branch reporting 4 pages READY on it is the *fail-open* direction, not an
improvement. That possibility was never checked. It is one more reason the figure must not be reused.

Same defect class as the fabricated insertion count in `docs/review-throughput-postmortem.md` and the
declaration-vs-drop-count inflation recorded in the pass criteria: a plausible number, repeated because
nobody had a control.

## The real baseline

Measured with every package built from the **engine 2.353.0** bundle, gate code at **master `c8b0a58`**,
run per-unit and aggregated (pointing the gate at the packages *parent* yields
`CANNOT_ESTABLISH 0/0` — each child is a package, the parent is not).

| unit status | count |
|---|---:|
| FINDINGS | 39 |
| NOT_APPLICABLE | 19 |
| READY | 7 |
| CANNOT_ESTABLISH | 2 |
| **total** | **62** |

| page/evidence metric | master `c8b0a58` |
|---|---:|
| `pages_expected` | 220 |
| **`pages_ready`** | **28** |
| **`pages_blind`** | **192** |
| `pages_unverifiable` | 0 |
| `pages_insufficient_grade` | 0 |
| `pages_dropped_explained` | 3 |
| `pages_dropped_unexplained` | 20 |
| `evidence_records` | 283 |
| **`evidence_rejected`** | **73** |

## Against the pre-written pass criteria

* **Criterion 1** (`pages_ready` well above 42): master is at **28**, i.e. *below* the 42 figure from
  run 407. This is the number PR #454 must move.
* **Criterion 2** (`pages_blind` must NOT fall below ~46): 192 — far above the floor, so there is
  plenty of headroom before a drop would look like fail-open.
* **Criterion 3** (refusal counters beside coverage): satisfied — `evidence_rejected = 73` is non-zero
  and reported. Any post-#454 rise in `pages_ready` must be read together with this number.

## The A/B to run once #454 merges

Re-run the identical per-unit aggregation against the identical packages, changing only the gate code.
Nothing else may vary — same bundle, same oracle, same packages.

**Reading the result:**

* `pages_ready` rises **and** `evidence_rejected` stays materially non-zero → the join works.
* `pages_ready` rises **and** `evidence_rejected` collapses toward 0 → suspect fail-open, investigate
  before accepting.
* `pages_blind` falling below ~46 → fail-open, and a FAIL regardless of how good coverage looks.

Guust-Franssens:
# Blind review round 2 — BLOCKED (boundary moved, not removed)

Same reviewer as round 1, re-run against `e1d90b4`. Tree confirmed clean.

## Finding A — CLOSED

Round 1 (order decided the verdict):

```
FIRST:    READY,    pages_ready=1, exit=0
REVERSED: FINDINGS, pages_ready=0, exit=1
```

Round 2:

```
A_FIRST    {"status":"CANNOT_ESTABLISH","pages_ready":0,"exit":3}
A_REVERSED {"status":"CANNOT_ESTABLISH","pages_ready":0,"exit":3}
A_THREE    {"status":"CANNOT_ESTABLISH","pages_ready":0,"exit":3}
```

Positive controls hold: duplicate agreeing LUIDs -> `READY exit=0`; valid + whitespace claim -> `READY exit=0`; conflicting revision claims fold to the weakest (`FINDINGS`, unverifiable).

## Finding B — original closed, **boundary moved one layer earlier** (blocking)

The round-1 reproduction is fixed: `B_ENTRY_CONFLICT` -> `FINDINGS/blind/exit 1`, `B_EXIT_CONFLICT` -> `NOT_CHECKED`, evidence listed under `foreign_workbook_evidence`.

But `reference_evidence.py:688-702` and `check_unit.py:1966-1969` both select the **first non-blank LUID by scope precedence** rather than checking that all declared LUIDs agree. With a matching LUID at state/entry scope and a **conflicting** LUID at manifest scope (matching SHA retained):

```
B_ENTRY_MULTISCOPE_CONFLICT {"status":"READY","page":"ready","pages_ready":1,
                             "sha256":1,"conflicting-identity":0,"exit":0}
B_EXIT_MULTISCOPE           {"status":"PASS","visual_present":1,"numeric_present":1,
                             "admitted_evidence":1,"foreign_workbook_evidence":[]}
```

The conflicting LUID disappears **before** `WorkbookIdentity.attribute()` sees the record. No schema or documented contract says a state/entry LUID may supersede a contradictory manifest identity, so under this PR's own invariant all non-blank LUID claims must agree.

## Asymmetry verified at the identity level

```
SHA_ONLY            route=sha256,                admitted=true
LUID_ONLY           route=luid,                  admitted=true
SHA_MATCH_LUID_DIFF route=conflicting-identity,  admitted=false
SHA_DIFF_LUID_MATCH route=foreign,               admitted=false
UNIT_CANNOT_SHA     route=unknown,               admitted=false
```

Ordinary cases still usable: `B_ENTRY_AGREE READY`, `B_ENTRY_ABSENT READY`, `B_EXIT_AGREE PASS`, `B_EXIT_ABSENT PASS`. So the fix is not over-strict.

## Boundary neighbours probed

| scenario | result |
|---|---|
| two conflicting provenance LUIDs, either order | `CANNOT_ESTABLISH` exit 3 |
| three rows A, A, B | `CANNOT_ESTABLISH` exit 3 |
| duplicate agreeing LUIDs | `READY` exit 0 |
| valid + whitespace LUID | `READY` exit 0 |
| same LUID, conflicting revisions | `FINDINGS`, stale/unverifiable |
| matching SHA + one conflicting LUID | correctly refused |
| **matching SHA + conflicting LUIDs at different scopes** | **READY/PASS — blocking** |
| missing LUID axis / missing SHA axis | remain usable |

## Mutations (reviewer's own, not the author's)

Restore first-valid-claim -> failed on `forward["units"] == reversed_["units"]`. Delete contradiction check (entry) -> expected `blind`, got `ready`. Delete it (exit) -> expected `WB_CONFLICT`, got `sha256`. Discard manifest LUID -> expected `blind`, got `ready`. All right-reason kills. `mutate_revision_key` 12/12, `mutate_check_unit` 92/92 with a surviving negative control, `mutation_reference_readiness` 80/80 across 159 checks.

**Disclosed dead branch confirmed accurate:** with axis order sha256 -> luid, `_contradiction(decided=WB_LUID)` is unreachable via `attribute()`; disabling it left `test_a_matching_luid_does_not_admit_a_record_whose_sha_is_another_workbooks` passing. That test covers the safety property but not the branch. Honestly disclosed, non-blocking.

## Pin verified legitimate, not a silencing edit

```
ca4d395  7642b526...
b9ef38e  6895a515...
dbc588b  63c18e14...
e1d90b4  63c18e14...
```

`ca4d395..b9ef38e` changed `object_identity.py` +76/-31 without re-deriving the pin. The new digest matches both fix commits exactly.

## Gates — all exit 0

4-file suite (326 passed), `test_shared_identity_pin` (4 passed), ruff format, ruff check, pylint, `mutate_revision_key`, `mutate_check_unit`, `mutation_reference_readiness`.

R0911 suppressions judged defensible: function-local, justified, covering 7-8 distinct verdict exits.

## Non-blocking follow-ups

- Two equal-SHA records distinguishable by `input.file` still yield `CANNOT_ESTABLISH`; `input.file` is not used to disambiguate. Over-strict, safe direction.
- `verify_shared_identity_pin.py` exits 1 comparing intentional PR drift against `origin/master`; its "retake verbatim" guidance is stale for deliberately reviewed shared-surface changes.

Guust-Franssens:
# Closing evidence: real-estate A/B, identical packages, only the gate code differs

Run 408, 62 packages built from the engine-2.353.0 bundle. Per-unit aggregation; nothing varied but the gate code.

| metric | master `c8b0a58` | this PR `99c8192` |
|---|---:|---:|
| unit statuses | 39 FINDINGS / 19 N-A / 7 READY / 2 CANNOT_ESTABLISH | **identical** |
| `pages_expected` | 220 | 220 |
| **`pages_ready`** | 28 | **28 — unchanged** |
| `pages_blind` | 192 | **63** |
| `pages_unverifiable` | 0 | **129** |
| `pages_dropped_unexplained` | 20 | 20 |
| `evidence_records` | 283 | 283 |
| **`evidence_rejected`** | 73 | **73 — unchanged** |

**192 = 63 + 129.** This is a pure reclassification of the same non-ready set, not a coverage change.

## Why that is the right result

The pre-written pass criteria warned that a coverage rise is also what deleting the guard looks like, and required every coverage number to be read beside its refusal counter. Here:

- **`pages_ready` did not move** -> nothing new was admitted, so no fail-open.
- **`evidence_rejected` held at 73** -> refusals intact.
- **`pages_blind` 63 stays above the ~46 floor** -> criterion 2 satisfied.
- Every unit status is identical -> no unit changed verdict.

The change is diagnostic precision: 129 pages move from *"no reference render at all"* to *"evidence exists but cannot be trusted"*, which is a materially more actionable message for the agent that reads it.

## What this retracts

⚠️ This supersedes my earlier claim (comment 5519914145) that this PR takes the entry gate from **0/7 to 4/7** and is therefore the cold-run coverage gate. **It is not a coverage fix.** On a real 48-workbook estate it changes **no** admission decision. That earlier figure was measured against an incomplete package and is withdrawn; see comment 5522531064 for the full retraction.

The value of this PR is elsewhere, and is real: three **reproduced** fail-open identity holes are closed (SHA-match/LUID-mismatch; order-dependent provenance; multi-scope LUID collapse), plus two near-neighbours found during the round-3 fix (the entry gate never read an oracle manifest's own LUID; the exit gate never read state-scope LUIDs) - both measured admitting a foreign workbook.

That this estate does not exercise them is expected and worth stating plainly: **no current producer writes `workbook_luid` at more than one scope**, so every conflicting case today is an enriched or hand-edited manifest. These are guards against a shape that exists but is not yet common - which is the correct time to close them, not evidence they are unnecessary.

PR #455, opened 2026-09-02
Refuse an oracle merge across tenants or across source revisions (refs #423)

Guust-Franssens:
> **Stacked on [#431](../../pull/431)** — base branch is `fix/capture-leg-decoupling`, not `master`.
> This is forced, not chosen: `group_oracle_by_workbook.py` imports `tableau_oracle_manifest`, which
> is new in #431, and `tests/test_group_oracle_multi_batch.py` is new there too. Merge #431 first and
> GitHub retargets this automatically. **Review the diff against `fix/capture-leg-decoupling`.**

## Contract

**Invariant + direction.** A merged manifest may only combine legs that provably describe **the same
Tableau source** and **the same revision of the same view**.

| direction | consequence | blocks merge? |
|---|---|---|
| **fail-open** — legs combined across tenants or revisions | one customer's artifacts presented as another's reference evidence; a stale render credited as current, with a digest and timestamp asserting otherwise, while the manifest reports `failed=0, render_unestablished=0` | **YES** |
| fail-closed — a promotion refused | one leg not promoted; the gap stays visible in `merge_stale_candidates` and is re-capturable | no |

**Closed surface — every field a merge may take from a different batch. `N = 5`:**

| field | where it comes from | gated by |
|---|---|---|
| identity (`view_luid`, `view_name`, `workbook_*`, `updated_at`, `captured_at`, `view_type`, …) | the newest record that saw the view | source identity |
| each of the 4 legs — `data`, `image`, `svg`, `pdf` (`RENDER_LEGS`) | resolved independently, newest promotable first | source identity **and** revision |
| `requested_renders` / `reference_required` | **unioned** across batches, never taken from the newest alone | source identity |
| provenance block (`server`, `site`, `rest_api_version`, `render_capability`) | the newest batch | source identity — this is the field the cross-tenant merge falsified |
| `source_batch` per leg | the batch that actually produced the artifact | — |

Residual, stated: `reference_missing` is **recomputed** over the merged evidence rather than carried,
because the newest batch's verdict is about a different set of views.

**Independent oracle.** On-disk capture directories built by the test, read back through
`load_batches` → `merge_batches` → `render_unestablished` — the verdict function, not the merge, is
what reports the count. Provenance is asserted through `source_batch` and `merge_stale_candidates`,
which name the batch rather than restating the merge's own conclusion.

- **positive control:** `test_the_same_tenant_spelled_differently_is_still_one_tenant` — a trailing
  slash and a capitalised host must still merge. Without it, "refuse anything not byte-identical"
  would pass every negative test while making the feature unusable.
- **negative control:** `test_a_manifest_that_records_NO_source_cannot_be_merged_with_one_that_does`
  — and the same test then merges that batch **alone**, so the refusal is shown to be about ambiguity
  rather than about the field being absent.

**Round route.** Fix round for a blind review's blocker. A new class ⇒ simplify/split/descope; this
PR *is* the descope of #431.

---

## Blocker: `merge_batches` merged incompatible captures

It never established that batches shared a server/site or a source revision. It took identity from
the newest view, then selected each successful leg from **any** older batch. Both reproductions were
rebuilt before fixing.

### 1 — a stale render credited as current evidence

Old batch `image=ok`; newer batch, a revision later, `data=ok, image=transient`.

| | before | after |
|---|---|---|
| merged `updated_at` | `2026-08-19` (new revision) | unchanged — identity still comes from the newest record |
| `data` from | `new-revision` | `new-revision` |
| `image` from | ⚠️ **`old-revision`, status `ok`** | `new-revision`, status `transient` |
| summary | `data_ok=1, image_ok=1, failed=0, render_unestablished=0` | `data_ok=1, image_ok=0, render_unestablished=1` |

A picture of a workbook that has since changed is not weaker evidence than no picture — it is
evidence pointing the wrong way, and it arrives with a digest and a timestamp that say otherwise.

### 2 — cross-tenant evidence mixing

Two batches from **different servers and sites**, sharing only a workbook caption, were accepted into
one manifest that declared only tenant B's `server`/`site` while its views named tenant A **and**
tenant B as sources. Nothing in the output said so: `source_batch` names a directory, not a tenant.
Captions like "Sales Dashboard" collide across customers as a matter of course.

Now refused, naming both batches, both hosts **and both sites**:

```
these captures describe 2 different Tableau sources and cannot be merged into one manifest
(tenant-a: server=https://tenant-a.online.tableau.com site=alpha,
 tenant-b: server=https://tenant-b.online.tableau.com site=beta). Merging them would present one
tenant's artifacts as another's reference evidence, under a single top-level server/site that names
only one of them. Group each source separately.
```

`main()` maps it to **exit 2**, alongside the existing `DuplicateBatchLabel`.

## What the fix does

- **Source identity** normalized for case and a trailing slash only — one tenant spelled two ways is
  one tenant. An **absent** identity is its own value, never a wildcard: "this manifest does not say
  which server it came from" is not evidence that it came from the same one, and a wildcard would let
  precisely the batch we know least about merge with anything.
- **Revision gate.** A leg may only be taken from another batch when both records carry the same
  **non-empty** `updated_at`. Sameness must be *proved*, so a missing revision disqualifies a
  cross-batch promotion rather than permitting one — "we cannot tell" and "they match" are different
  answers. The newest candidate always keeps its own legs; only importing from an older batch is gated.
- **A refusal is recorded, not dropped.** `merge_stale_candidates` names view, leg, batch and both
  revisions; the leg carries an explicit `stale_revision` status preserving the older batch's own
  verdict, with **no `path`**, so nothing downstream copies it. It is never `ok`, so no counter
  credits it. Reported only when the current revision did not establish the leg itself — otherwise
  every leg of every twice-captured view would be listed and the entries that matter would be buried
  (caught while writing the tests: the `data` leg was appearing when the new batch had already
  established it).
- **The operator is warned on the console.** A manifest field nobody reads is not a report; the
  warning names the batch and the stale revision, and says to re-capture.

## A fixture claimed to be faithful and was not

`_view` carried the docstring *"shaped exactly as `capture_tableau_oracle` writes it"* while omitting
`updated_at` — which the real writer **does** stamp (`capture_tableau_oracle.py`,
`"updated_at": view.get("updatedAt")`). So every scenario described views of *unknown* revision — the
one shape that cannot support a cross-batch promotion — and the fixture structurally could not
distinguish "the same view captured twice" from "two versions of a view", which is the defect itself.

It now stamps a shared `REVISION` by default, so the six existing scenarios keep testing what they
were written for. `updated_at=None` stays reachable, and
`test_a_view_whose_revision_is_UNKNOWN_does_not_promote_across_batches` pins the fail-closed handling
of a server that omits `updatedAt` rather than leaving it to be discovered.

## Gates — by exit code

| gate | exit |
|---|---|
| `pytest tests/` | 1 — **3 known pre-existing engine-drift failures only** (`test_upstream_repro_pins.py`), 3913 passed |
| `pylint scripts` / `pylint conftest.py` | **0** ⚠️ it printed `10.00/10` while exiting 8 before the fix |
| `python tests/mutation_capture_legs.py` | **0** — 69 mutations, 65 caught, 2 cosmetic controls survived, 2 absent-anchor controls invalid, as declared |

Five new mutations, each killed by its own anchor: no refusal at all; an absent identity treated as a
wildcard; raw string comparison (which would refuse one tenant as two); the site ignored; the
revision gate removed; an empty revision counting as a match; a stale leg dropped instead of marked.


Guust-Franssens:
## Blind round-1 review: **BLOCKED**

Reviewed at `e351b7b` (CI confirmed green for that **exact** SHA via `check-runs`, not the pre-rebase ancestor `badc8c9c` — the force-push did not auto-trigger, so the run was dispatched manually as `33706728543`).

### What is confirmed working — do not regress it
Three-batch `original -> retry -> retry2` selection is correct: `retry2` wins when only it succeeds, an older batch wins when the newest failed, and partial success stays **leg-granular** (image and data resolve to different batches independently). Targeted suite 60 passed. Mutation harness exit 0, 70 caught / 2 cosmetic controls survived; the argument-order parity test was checked for self-referentiality and is sound. Revision comparison uses Tableau REST `updated_at`, **not** raw `.twbx` digests — so the archive-repack false-drift problem measured on the sibling PR does **not** apply here.

### Blocking findings

**1. Two UNKNOWN sources are treated as the same tenant** (`group_oracle_by_workbook.py:255-279`). `_source_identity()` maps missing identity to `("", "")` and `_refuse_incompatible_sources()` checks only set cardinality, so two anonymous manifests merge although sameness was never established — `refusal_fired: false`, `promoted_image_source_batch: "anonymous-b"`. This contradicts the PR's own cross-tenant refusal claim. "Cannot establish" needs to be its own blocking state rather than a value that compares equal to itself.

**2. A refused old-revision artifact stays physically in `reference/`** (`:540-585,716-735`). The revision guard correctly refuses it and the manifest correctly records `image_status: transient` with no image path — and the old-revision PNG is still present in `reference/images/`. A consumer reading the directory rather than the manifest gets refused evidence.

**3. Source-side normalization collisions merge different workbooks** (`:177-199,752-757`). Destination collisions are caught; two distinct *source* workbooks normalizing to one key are not. Measured: 2 distinct source workbooks, 1 normalized key, 1 destination, 2 image files, 1 manifest view, **1 unreported image file**. ⚠️ Not hypothetical — the reference estate has **48 distinct workbook LUIDs but only 46 normalized name keys**.

### AC#3 not yet met
`:208`, `:677-690,808` read only directories passed via repeated `--oracle`, not every batch on disk. With `retry2` present but unlisted, it is silently unused (`retry2_exists_on_disk: true`, reported batches exclude it). In the reviewer's words: *"the boundary moved from one batch to every argument the operator remembered."* That is a moved failure boundary, not a removed one.

### Follow-up (fail-closed, non-blocking)
Cross-revision refusal warns and persists a stale count but returns exit 0 (`:843-857,885-897`); the "reflected in exit code" behaviour is absent.

`Fixes #423` deliberately **not** proposed — AC#3 is the criterion this PR exists to deliver and it is not yet satisfied.

Routed to a fresh agent for fixes; re-review will go back to this reviewer, which already has the reproductions built.

Guust-Franssens:
Parking as draft to cut the in-flight batch down.

Two things established this morning:

**1. The reported size was an artifact, not real work.** This PR's base branch was `fix/capture-leg-decoupling`, which had already been squash-merged into master as `dab3acb`. GitHub was therefore diffing against a dead base and reporting **+25,876 / 68 files / CONFLICTING**. Retargeted the base to `master`: the real change is **+3,068 / 4 files / CLEAN**. No code changed — only the base pointer.

Worth remembering generally: *a stacked PR whose base gets squash-merged reports garbage sizes and a false conflict.*

**2. The scope question that survives that correction is real.** `scripts/group_oracle_by_workbook.py` goes **304 -> 1,264 lines** (4.2x) with +1,895 lines of test/mutation machinery, across 16 commits. That is the "unbounded claim fixed one site at a time" shape `AGENTS.md` warns about, and it is larger than #423 AC#3 on its own justifies.

Also recovered here: the agent working this branch was killed by a host crash at ~06:00 with uncommitted work. Salvaged as `3f61688` after verifying `tests/test_group_oracle_by_workbook.py` -> 23 passed.

**Not a cold-run blocker**, so it waits behind #454. When it resumes, the question to answer first is whether this should be *split or descoped* rather than given another review round.

Guust-Franssens:
## Blocker before this can leave draft: it breaks a pre-existing test it never touched

CI is red, and **not** for the standing engine-drift reason. `origin/master` is green across its last
five runs, so the zero-failure baseline is real and this failure belongs to this branch.

```
FAILED tests/test_group_oracle_by_workbook.py::test_a_missing_manifest_names_the_file_it_wanted
  group_oracle_by_workbook.UnlistedBatchOnDisk: 48 capture batch(es) sit beside the ones given
  and were not passed: /tmp/pytest-of-runner/pytest-0/test_a_clean_capture_exits_zer0,
  /tmp/pytest-of-runner/pytest-0/test_the_manifest_censuses_vie0, ...
1 failed, 4559 passed, 98 skipped in 275.50s
```
(run 33720872472)

`tests/test_group_oracle_by_workbook.py` is **not in this PR's diff** — the four changed files are
`scripts/README.md`, `scripts/group_oracle_by_workbook.py`, `tests/mutation_capture_legs.py`,
`tests/test_group_oracle_multi_batch.py`. `UnlistedBatchOnDisk` does not exist on `origin/master`; it
is new at `group_oracle_by_workbook.py:205`. So the new guard changed the behaviour of an existing
passing test.

## The guard itself looks right — do not weaken it

`_refuse_unlisted_siblings()` (`:302-334`) is narrow exactly as its docstring claims: it walks only
the immediate parents of the directories actually named, and only counts a sibling that satisfies
`is_capture_batch()`. Those 48 directories really are capture batches — they are the tmpdirs of
`test_capture_tableau_oracle*.py`, every one of which legitimately writes an oracle manifest. The
guard found real batches beside the named one and refused, which is its whole purpose under #423.

## So the fix is test isolation, not a guard change

The failing test places its batch **directly in pytest's shared `tmp_path` root**, where every other
test's tmpdir is a sibling. Give it a private parent (`tmp_path / "run" / "batch"`) so the sibling set
is its own, and the guard has nothing unrelated to see.

Worth stating why this matters beyond the red tick: the failure is **order-dependent**. It only fires
once enough oracle-capture tests have run first and left batches in the shared root, so it will
present as an intermittent, unrelated-looking flake on other PRs rather than as a defect in this one.

## Two things to confirm while fixing

1. **Is any other pre-existing test in the same position?** The same shared-root pattern will bite any
   test that names a batch, so grep for batch fixtures rooted straight at `tmp_path` rather than
   fixing only the one CI happened to reach first.
2. **`--oracle-root` uses a different, stricter rule** (`:290-293`: an unclassifiable directory is
   blocking). Confirm the isolation fix does not simply move the same test onto that path.

Not fixing it here — flagging so it is done deliberately. Everything else in this PR (the cross-tenant
refusal and the revision gate) reads as the right shape, and the `_view` fixture correction that
stamps `updated_at` is a genuine find.

Guust-Franssens:
@copilot The current exact-tip blind review reproduced three fail-open provenance holes. This is the final scope-frozen correction: fix only these three central invariant failures, with direct tests and no new framework.

1. Canonical run-layout discovery
- The supported layout is `_runs/<run>/oracle`.
- Listing `_runs/001/oracle` and `_runs/002/oracle` must discover or refuse an omitted `_runs/003/oracle`; immediate-sibling-only scanning is insufficient.
- `--oracle-root _runs` must recognize the bounded immediate-grandchild shape `<run>/oracle` rather than rejecting every run directory as unclassified.
- Do not add arbitrary recursive filesystem discovery. Define one explicit boundary: a capture root itself, immediate capture children, or immediate run children whose `oracle/` child is a valid capture. Anything else under the declared boundary remains blocking unless explicitly excluded.

2. Artifact containment
- One shared helper must validate every `entry.path` before both promotion and copy.
- Reject absolute paths, UNC paths, rooted paths, `..` traversal, and resolved symlink/junction escapes.
- Resolve the candidate and batch root and require the source to remain beneath the resolved batch root. A manifest from tenant A must not reference bytes outside tenant A's batch while retaining tenant A provenance.
- The direct reproduction `path='../outside.csv'` must be non-promotable and must never copy.

3. Timestamp semantics
- Present `captured_at` and `updated_at` values must be parsed and normalized as timestamps once at manifest load/validation.
- Blank, whitespace, or invalid strings are unassessable/malformed, never proof that revisions match and never lexicographic freshness keys.
- Absent/null fields may retain the existing documented unestablished behavior.
- Two whitespace `updated_at` values must not authorize combining an old image with new data.

Hard scope limits:
- Change only `scripts/group_oracle_by_workbook.py`, `tests/test_group_oracle_multi_batch.py`, and, if needed for the canonical-layout control, `tests/test_group_oracle_by_workbook.py`.
- No new module, mutation runner, digest, census, generic path framework, unrelated refactor, or documentation expansion.
- Preserve same-source/same-revision per-leg retry selection and existing explicit `--exclude` behavior.
- Do not address any new non-safety finding in this PR; report it as residual instead.

Add direct positive and negative controls for all three reproductions. Run focused tests, ruff format/check, and pylint by exit code. Commit and push to this PR branch. This will receive a targeted re-review of these three findings only.

Guust-Franssens:
@copilot Targeted re-review closed artifact containment and timestamp validation. Two defects remain in the same canonical-layout discovery class. Make only this final small correction.

1. Explicit canonical cousins
- Given listed captures `_runs/001/oracle` and `_runs/002/oracle`, an existing `_runs/003/oracle` must be discovered or cause `UnlistedBatchOnDisk`; the first two must never be accepted alone.
- Keep discovery bounded to the already-supported `<scope>/<run>/oracle` shape. Do not add generic recursive scanning.
- Prefer deriving the shared canonical scope and comparing the bounded discovered set with listed + excluded captures, reusing existing discovery logic rather than adding another independent walker.

2. Exclusion regression
- Under `--oracle-root _runs`, `--exclude _runs/003/oracle` must exclude the discovered batch itself.
- Preserve support for excluding `_runs/003` if that is already intentional, but the CLI's batch-path form must work.
- When recognizing `child / "oracle"`, compare the exclusion set with both the run child and the resolved discovered batch path before adding it.

Direct controls required:
- listed 001/002 with unlisted 003 -> refuse and name 003;
- `--oracle-root _runs --exclude _runs/003/oracle` -> discover only 001/002;
- existing direct-sibling layout and explicit parent exclusion remain unchanged.

Hard scope: change only `scripts/group_oracle_by_workbook.py` and `tests/test_group_oracle_multi_batch.py`; no new helper module, recursion, mutation framework, docs, timestamp/path changes, or unrelated cleanup. Run focused tests, ruff, and pylint; commit and push. The same reviewer will confirm only these two reproductions.

Guust-Franssens:
@copilot Production is approved; make only this mechanical test correction.

`tests/test_group_oracle_multi_batch.py:1476` expects the Unix-only text `003/oracle`, but the production exception correctly uses the native Windows path `003\oracle`. The exact targeted test is red on Windows before mutation.

Fix the assertion so it is separator-independent. Prefer comparing the exception text with `str(the_expected_capture_path.resolve())` or normalized `Path` components; a `[\\/]` regex is acceptable if the existing fixture does not expose the path object cleanly.

Hard scope:
- Change only `tests/test_group_oracle_multi_batch.py`.
- Do not change production code, discovery behavior, messages, docs, or any other test.
- Run the focused cousin-listing/exclusion tests and the full multi-batch test module, then ruff/pylint as applicable.

Commit and push. The same reviewer will only confirm the Windows test baseline and mutation reachability.

Issue #458, opened 2026-09-03
promote_unit.py: the phase 2 to phase 3 step is manual, and it is the hop where a wrong byPath ships a report with no model


Guust-Franssens:
## What

Build `scripts/promote_unit.py` — the missing phase 2 → phase 3 step.

Today the pipeline is:

| phase | location | tool |
|---|---|---|
| 1 collect + convert | `_runs/<NNN>-<slug>/{assessment,assets,bundle,oracle}` | `assess_estate.py`, `harvest_estate_assets.py`, `run_estate.py`, `capture_tableau_oracle.py` |
| 2 package for the agent | `_runs/<NNN>-<slug>/packages/<Unit>/` | `package_unit.py` (#446 / #451) |
| 3 ship | `migrations/{workbooks,datasources}/<slug>/fabric/` | **nothing — manual copy** |

Phase 3 is a hand-executed copy, and it is the hop with the highest cost of getting it wrong.

## Why it matters

`powerbi-report-gotchas` §3 documents two cases and only one is a plain copy:

* **model per workbook** — `<Name>.Report/` and `<Name>.SemanticModel/` are already siblings, so copy the *contents* of `<bundle>/pbip/<wb>/` (not the folder — it is named for the workbook while the model inside is named for the datasource, so copying the folder nests them wrongly).
* **shared/published datasource** — the model lands once under `datasources/<ds-slug>/fabric/` and the report under `workbooks/<slug>/fabric/`, so they are no longer siblings and `definition.pbir` must be rewritten to `../../../../datasources/<ds-slug>/fabric/<Name>.SemanticModel` — four levels up from inside `<Name>.Report/`.

A wrong `byPath` opens as **a report with no model**, which reads like a binding defect and sends whoever hits it debugging the wrong layer. `powerbi-report-author validate` **cannot catch it** — measured: a `.Report` whose `byPath` named a `.SemanticModel` that exists nowhere in the bundle returned `errorCount: 0`. Shape, not target.

There is a second, independent failure already observed on a 46-asset estate: a report folder that had **passed a sign-off** contained only Desktop-local settings — no pages, no visuals, no model. Every folder that was supposed to exist did exist. A folder count is not a content check.

## Acceptance criteria

1. `python scripts/promote_unit.py --package <_runs/NNN/packages/Unit> --slug <slug>` copies phase 2 → `migrations/{workbooks,datasources}/<slug>/fabric/`.
[...]

Issue #415, opened 2026-08-31
Document the remaining parallel-test contention modes and supported invocation


Guust-Franssens:
## Completed

PR #529 now serializes the TMDL oracle's first build with a bounded, fd-scoped OS advisory lock. Concurrent workers no longer launch overlapping `dotnet build` operations into the same output directory.

The separate availability question where waiters share a deadline across failed build generations is tracked in #539 and is not part of this issue.

## Remaining scope

Parallel-test guidance still describes several different conditions in different places:

1. **Timing-bounded tests** under xdist: a measured rare failure that passes serially; widening the bound is not automatically a fix.
2. **Live GUI/UI-Automation contention**: slow harvest can correctly degrade to `DIALOG_UNREADABLE`; this is a non-clean contention result, not an ordinary flaky pass/fail.
3. **`test_check_unit...unattributable_connection_fixture` sighting**: one reported full-parallel failure, but controlled serial and xdist-only repetitions were green and another full baseline did not reproduce it.
4. **Supported command/baseline ambiguity**: agents repeatedly run different xdist marker/filter combinations and compare only counts, then treat a different failure set as the same baseline.

## Bounded acceptance

- Put one concise supported parallel-test command and its reason-keyed expected skips/deselections in `docs/parallel-test-loop.md`.
- Separate true product failures, expected contention degradation, timing-sensitive diagnostics, and unsupported invocation artifacts.
- Record exact node IDs/reasons, never only a failure/skip count.
- For the unattributable-connection sighting, either produce a deterministic reproduction and fix its cause or mark it unconfirmed with the attempts that failed to reproduce.
- Link the TMDL-oracle build lock to #529 as completed rather than re-explaining its implementation here.

Do not add retries, widen timing thresholds, or create a generalized flake framework without a controlled reproduction that distinguishes the proposed cause.

Guust-Franssens:
## A concrete mechanism for one of the three — `tools/tmdl_oracle` has an unlocked first build

Found while merging `origin/master` into another branch, and it is **verified in the source**, not inferred from a flaky run.

`scripts/tmdl_oracle.py:114` is check-then-act with **no lock**:

```python
def ensure_built(dotnet: str) -> Path:
    """Build tools/tmdl_oracle once, returning the assembly path."""
    if not _sources_newer_than_build():
        return DLL
    ...  # subprocess: dotnet build tools/tmdl_oracle -c Release
```

[...]

PR #459, opened 2026-09-03
Document the three-phase migration pipeline, and name the packages/ location

Guust-Franssens:
## What this is

The toolkit has a clear three-phase shape and **no document explained it** — `docs/operator-runbook.md`
carries the individual commands but never says why there are three separate locations or what moves
between them. This adds the explanation, and makes one small code change so the doc describes reality
rather than an aspiration.

- **`docs/migration-phases.md`** (new) — phase by phase: what it produces, which tool, where it lands,
  what the gate is, what the known gaps are. ASCII flow diagram, the `packages/` vs `deliverables/`
  distinction, and the phase 2 → 3 `byPath` hazard.
- **`README.md`** — a general-overview `## The three-phase pipeline` section that links into the doc
  and the runbook, plus a nav-bar entry.
- **`scripts/work_dirs.py`** — `packages` joins `CANONICAL_SUBDIRS` (after `oracle`, before
  `deliverables`) with a matching `RunPaths.packages` property, so `package_unit.py --out` has a
  canonical home rather than the ad-hoc `_packages` the runbook used.
- **`tests/test_work_dirs.py`**, **`docs/operator-runbook.md`**, **`AGENTS.md`**, **`docs/INDEX.md`** —
  follow-through.

`package_unit.py` itself is unchanged.

## What I verified

Everything measured against `_runs/408-…`, a real 48-workbook cold run, on 2026-09-03.

| claim in the doc | how |
|---|---|
| 48 workbooks, 0 listing errors | `assessment/estate_survey.json` `summary` |
| 360 views = 60 dashboards / 300 worksheets | `oracle/oracle-manifest.json` `view_count` / `view_types` |
| 67 assets, 0 parse failures on either tier | `assets/parse-sweep.md` |
| 62 pbip units / 48 report baselines / **18** model baselines / 46 handover slices | directory counts under `bundle/` |
| `deliverables/` is empty and nothing writes to it | 3 of 9 `_runs/` dirs have one, all 0 items; the other 6 have no such subdir. Only `scripts/` matches for the word are prose in `check_migration_progress.py` / `run_estate.py` |
| phase 3 artifacts are **not** blanket-gitignored | `git check-ignore -v` on six representative paths; `git ls-files migrations` = 3 READMEs |
| issue #458 exists and is about the missing promote step | `gh issue view 458` |

**The `--out` constraint was measured, not reasoned:** `python scripts\package_unit.py --bundle <real
bundle> --out <run>\packages` exits **2** — *"sits beside evidence the gates also scan … Choose an
`--out` outside the capture tree"* — while `--out <run>\packages\<batch>` exits **0**, and
`check_reference_readiness.py` on the resulting package (no flags) reports **0 unverifiable**, i.e. no
shadowing.

**The new tests were mutation-tested.** Removing `"packages"` from `CANONICAL_SUBDIRS` fails 3 tests;
swapping its order with `"oracle"` fails 1. The nav-index gate was mutation-tested too: deleting the
`docs/migration-phases.md` row makes `check_navigation_index.py` exit 1 naming the file.

## Gates (by exit code)

| command | exit |
|---|---|
| `ruff format --check conftest.py scripts tests .github/skills` | 0 |
| `ruff check conftest.py scripts tests .github/skills` | 0 |
| `pylint scripts` | 0 |
| `set_data_folder.py --check` | 0 |
| `pytest -q tests/test_work_dirs.py` | 0 (25 passed) |
| `sync_agent_conventions.py --check` | 0 (AGENTS.md edit is outside the synced block) |
| `pytest -q -n auto --dist loadfile -m "not (serial or timing)"` | 1 — 4548 passed, 6 failed, all known/unrelated (see below) |

The 6: 3 in `tests/test_upstream_repro_pins.py` (known-expected), 2 in `tests/test_gui_marker_gate.py`
(xdist-only, issue #415), and `test_check_datamodel.py::test_biff8_gate_reaches_the_cli_exit_code`,
which **passes in isolation** (exit 0) and is an xdist encoding flake — my diff touches no file it
reads.

## What I did NOT verify

- **The `packages/<batch>` depth is a design choice, not the only possible fix.** Making the guard
  aware of the run layout, or letting `package_unit.py` accept the bare `packages/` and create the
  batch level itself, would work too. I did not change `package_unit.py`, so `--out` stays free-form
  and nothing *enforces* the nested form beyond the guard and the docs.
- **No re-run of `run_estate.py` / `assess_estate.py` / the oracle capture.** All phase-1 figures are
  read from artifacts an earlier run left on disk, not reproduced.
- **The `deliverables/` finding is machine-local.** It is "empty in every run on this machine", not
  "empty everywhere"; and "no script writes to it" is a grep of `scripts/`, not of every possible
  caller.
- **Phase 3's `byPath` rules are quoted from `.github/skills/powerbi-report-gotchas/SKILL.md` §3**
  (🟢 verified there 2026-08-11). I did not re-execute a shared-datasource ship or open the result in
  Power BI Desktop.
- The `packages/` addition is inert for existing runs — `allocate_run` only creates subdirs at
  allocation time, so runs allocated before this change have no `packages/` and nothing backfills it.
  I did not add a migration helper (that is #234 follow-up).

The phase 2 → phase 3 step remains a manual copy; that gap is tracked as #458 and this PR only
documents it.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -722,6 +722,11 @@ _runs/<NNN>-<slug>/
     assets/             harvest_estate_assets.py-shaped downloads
     bundle/              run_estate.py-shaped conversion output
     oracle/               capture_tableau_oracle.py-shaped reference capture
+    packages/              package_unit.py's per-unit, agent-facing handover packages — ⚠️ `--out`
+                            names a subdirectory INSIDE this one (`packages/<batch>/<Unit>/`), never
+                            `packages/` itself: `conflicting_evidence_dirs` refuses an `--out` whose
+                            parent holds `oracle/`, and a run root always does (measured: bare exits
+                            2, one level deeper exits 0). See `docs/migration-phases.md`
     deliverables/          operator-facing outputs meant for the CUSTOMER, never for git —
                             the `ses-prep/` near-miss (issue #322): a `connections.json`/`.md`
                             naming 17 real customer servers landed unprefixed, unignored, at the
@@ -736,7 +741,7 @@ ignored, which proves nothing; see `harvest_estate_assets.py`'s own guard for th
 
 **`scripts/work_dirs.py`** is the single source of truth for these paths — `sanitize_unit_key`,
 `allocate_run` (atomic `mkdir`-exclusive, retry on collision, never a read-then-write race),
-`RunPaths` (the six subdirs above as properties), and `list_runs`. It resolves the repo root from
+`RunPaths` (the seven subdirs above as properties), and `list_runs`. It resolves the repo root from
 its **own file location**, never from `Path.cwd()` — an empty stray `fabric/` was once written at
 the repo root by a script that resolved a relative path against whatever CWD an agent happened to
 invoke it from, which is exactly the failure a single importable resolver removes.
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 &nbsp;![Migrations](https://img.shields.io/badge/migrations-16-2ea44f)
 &nbsp;![Parser tests](https://img.shields.io/badge/parser_tests-48%2F48-2ea44f)
 
-**[TL;DR](#tldr)** &nbsp;·&nbsp; **[Quickstart](#quickstart)** &nbsp;·&nbsp; **[What you get](#what-you-get)** &nbsp;·&nbsp; **[Repo layout](#repo-layout)** &nbsp;·&nbsp; **[How it works](#how-it-works)** &nbsp;·&nbsp; **[Prerequisites](#prerequisites)** &nbsp;·&nbsp; **[Capabilities &amp; limits](docs/capabilities-and-limitations.md)**
+**[TL;DR](#tldr)** &nbsp;·&nbsp; **[Quickstart](#quickstart)** &nbsp;·&nbsp; **[What you get](#what-you-get)** &nbsp;·&nbsp; **[Repo layout](#repo-layout)** &nbsp;·&nbsp; **[Three phases](#the-three-phase-pipeline)** &nbsp;·&nbsp; **[How it works](#how-it-works)** &nbsp;·&nbsp; **[Prerequisites](#prerequisites)** &nbsp;·&nbsp; **[Capabilities &amp; limits](docs/capabilities-and-limitations.md)**
 
 </div>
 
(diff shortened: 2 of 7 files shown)


Guust-Franssens:
## Review round 2 — all three blocking findings fixed at `0a01953`

Every finding was reproduced before fixing. All three were real.

### BLOCKING 1 — phase 3 promoted from the wrong location

Reproduced the mechanism independently: `scripts/package_unit.py:847` is
`shutil.copytree(source, dest / "fabric", dirs_exist_ok=True)`; `package_unit.py:785` writes a README
telling the agent *"`fabric/` | the engine WORKING COPY - edit here"*; `AGENTS.md:773` says
`<bundle>/pbip/` is where agents edit. Also confirmed the trees are **byte-identical immediately after
packaging** (`git diff --no-index` exit 0, no output) — so they can only diverge by someone editing
one, which is exactly the reviewer's sentinel result.

`docs/migration-phases.md` phase 3 now opens with **"First decide WHERE you are promoting from — the
repo currently gives two answers"**: a table citing both sources, the copytree line, the sentinel
evidence, and the interim instruction to promote from wherever the edits are and verify both ways.
The `byPath` / shared-datasource mechanics are unchanged and now phrased source-neutrally
("the promoted unit folder", not "`<bundle>/pbip/<wb>/`"). #460 is added to the flow diagram, the
phase-2 section, the gaps table and the README bullet — as a plain reference, no closing keyword.

The verification commands were **run**, not written:

| command | result |
|---|---|
| `git diff --no-index --stat <bundle>\reports\Superstore.Report <bundle>\pbip\Superstore\Superstore.Report` | `22 files changed, 1807 insertions(+), 194 deletions(-)`, exit 1 |
| `git diff --no-index --stat <bundle>\pbip\Superstore <pkg>\Superstore\fabric` | exit 0, no output |

Both readings are documented, including that exit 0 there means "not diverged yet, the choice does
not matter", so the reader is not left to guess.

### BLOCKING 2 — `_runs` is not "safe to delete"

`README.md` now reads: *"Gitignored by construction (`/_*`), but **not disposable**: only its
`scratch/` subdir is, and a whole run becomes safe to delete only after its units are promoted and
verified."* `docs/migration-phases.md` gained a matching ⚠️ paragraph ("Gitignored is not the same as
disposable"), quoting `work_dirs.py`'s own "the only subdir a future `--prune` may ever delete".
Note the "safe to delete" wording was pre-existing at `README.md:181`; my first commit widened its
scope to "the first two phases", which made it worse. Fixed, not defended.

### BLOCKING 3 — a documented command does not run

Confirmed: `grep` in PowerShell → *"The term 'grep' is not recognized"*. Replaced with
`Select-String -Path … -Pattern '^EMPTIED_VISUAL'`, and **run in a child `pwsh -NoProfile` process to
get a real process exit code**: **0** with a match and **0** without one. The prose now also states
that this runbook's commands are PowerShell. That is the only remaining `grep` in
`docs/operator-runbook.md` (it is the explanatory clause); the five in four other docs are
pre-existing and outside this PR.

### Non-blocking

1. **Superlatives dropped.** "the riskiest hop" / "the highest-cost mistake in the pipeline" →
   "a high-risk hop", followed by the two failure modes I can evidence (`byPath` target unchecked by
   `validate`; unsettled promotion source).
2. **`blind` is a finding, not a pass** — said explicitly in the README section: it exits non-zero,
   and it means a fidelity bug on that page would be structurally unfalsifiable.
3. **Shared-datasource split clarified** — the model lands **once** under
   `migrations/datasources/<ds-slug>/fabric/` and is shared; the report goes to its own
   `migrations/workbooks/<slug>/fabric/`; they do **not** both sit under the workbook's `fabric/`.

### Untouched, as instructed

The gitignore table, the three tracked files under `migrations/`, the `--out` nesting constraint, the
19-harvested-vs-11-required distinction, the run-408 figures and the `work_dirs.py` tests are
unchanged. `AGENTS.md:773` is inside the synced `<!-- BEGIN:shared-conventions -->` block — changing
it would resolve #460 and rewrite five persona files, so it is left alone
(`sync_agent_conventions.py --check` exits 0).

**Engine version:** I cite no engine version anywhere in the diff (checked by regex for `2.3xx`), so
the 2.339.0 → 2.353.0 upgrade needs no edit. The structural figures I do cite (62/48/18/46) are
reported unchanged.

### Gates, by exit code

| command | exit |
|---|---|
| `ruff format --check conftest.py scripts tests .github/skills` | 0 |
| `ruff check conftest.py scripts tests .github/skills` | 0 |
| `pylint scripts` | 0 |
| `set_data_folder.py --check` | 0 |
| `sync_agent_conventions.py --check` | 0 |
| `pytest -q tests/test_work_dirs.py` | 0 (25 passed) |

Plus a relative-link check over both changed docs: 0 broken.

### What I still cannot verify

The `#460` interim instruction is a **procedure**, not a guard — nothing enforces it, and a reader who
skips the two `git diff` commands can still promote the wrong tree. That is inherent until #460 is
decided and a promote tool exists (#458). I also did not open any promoted deliverable in Power BI
Desktop.

Guust-Franssens:
## Round 3 — both blocking items fixed at `5bd9ed8`

Both reproduced before fixing. Both were errors in the interim instruction itself.

### BLOCKING 1 — `--stat` proves divergence, not provenance

Reproduced the falsification: a **deletion-only** edit inside the package (`_Measures.tmdl`, 9 lines
removed) →

```
1 file changed, 9 deletions(-)
DIFF_EXIT=1
```

No extra insertions, yet the package is the edited side. The doc now states:

- a non-empty `git diff --no-index` means the two trees have **diverged, and nothing more**;
- **insertion count specifically does not identify the edited side** — deletion-only produces none,
  replacement yields equal insertions and deletions, and both trees may have been edited;
- provenance comes from the **full diff** (drop `--stat`), the `_build/generated-edit-declarations/`
  records where they exist, and the operator's knowledge of where the agent was told to work;
- **if you cannot establish it, stop and do not promote.**

Kept as a correction inside the existing block; no new subsection.

### BLOCKING 2 — the after-promotion check was wrong for a shared datasource

Reproduced on a **correctly** split promotion (report → `workbooks/<slug>/fabric/`, model →
`datasources/<ds-slug>/fabric/`):

| comparison | exit | result |
|---|---|---|
| whole source unit vs workbook destination (the old documented form) | **1** | `20 files changed, 1272 deletions(-)` — flags a difference *because the model correctly split off* |
| `<source>\<Name>.Report` vs `migrations\workbooks\<slug>\fabric\<Name>.Report` | **0** | report matches |
| `<source>\<Name>.SemanticModel` vs `migrations\datasources\<ds-slug>\fabric\<Name>.SemanticModel` | **0** | model matches |
| whole source unit vs `fabric/` for a **model-per-workbook** promotion | **0** | whole-tree form is correct here |

The command block is now labelled per case: one whole-tree comparison for model-per-workbook, two
separate comparisons for a shared datasource, plus an explicit ⚠️ that the whole-tree form must never
be used on a shared datasource because it calls a correct promotion wrong. The commands now match the
prose that was already correct.

### Gates, by exit code

| command | exit |
|---|---|
| `ruff format --check conftest.py scripts tests .github/skills` | 0 |
| `ruff check conftest.py scripts tests .github/skills` | 0 |
| `pylint scripts` | 0 |
| `set_data_folder.py --check` | 0 |
| `sync_agent_conventions.py --check` | 0 |
| `pytest -q tests/test_work_dirs.py` | 0 (25 passed) |
| relative-link check over both changed docs | `MISSING_LINKS=0` |

Diff for this round: **1 file, +24/−7** — `docs/migration-phases.md` only. Nothing else touched; the
items confirmed closed (retention, `Select-String`, #460 handling, superlatives, `blind`, the
shared-datasource prose) are unchanged.

### Nothing in the round-2 review looks wrong to me

Both findings reproduced with the reviewer's own numbers on the first attempt. One thing I deliberately
did **not** add, in line with "do not expand the doc further": the provenance routing is prose, not a
command — `_build/generated-edit-declarations/` is a convention that may be absent, which is why the
doc says "where they exist" and terminates in *stop and do not promote* rather than in a check that
would be vacuous when the records are missing.

Issue #460, opened 2026-09-03
Two documented edit locations: AGENTS.md says <bundle>/pbip/, package_unit.py says <package>/fabric/ - and the package is a copytree, so promotion silently loses agent work


Guust-Franssens:
## The defect

Two shipped documents name **two different locations** as the place an agent edits, and one is a **physical copy** of the other. Whichever an agent picks, promoting from the other silently discards its work.

| source | says the agent edits |
|---|---|
| `AGENTS.md:773` (synced into every persona) | `<bundle>/pbip/` — *"agents edit **here**"* |
| `scripts/package_unit.py:785` (the package's own `README.md`) | `<package>/fabric/` — *"the engine WORKING COPY - **edit here**"* |

`package_unit.py:847` is `shutil.copytree(source, dest / "fabric", dirs_exist_ok=True)` — a real copy, not a link or junction. So the two locations diverge the moment either is edited.

## Reproduction

Found by a blind review of PR #459 using a sentinel edit, then independently confirmed:

```
PACKAGE_EDIT_EXISTS=True
BUNDLE_EDIT_EXISTS=False
```

An edit made where `package_unit.py` tells the agent to make it is **invisible** in `<bundle>/pbip/`. Promotion guidance that copies from `<bundle>/pbip/<wb>/` therefore ships unchanged engine output and loses every agent-authored TMDL/PBIR change — silently, with no gate firing.

## Why this is not just a docs fix

The bundle is not merely a convention; machinery keys on it:

* `run_estate.py` records generated-artifact hashes in `input_manifest.json`;
* `check_migration_progress.py --tamper` compares against those hashes and needs a matching `_build/generated-edit-declarations/*.json`;
* `scripts/declare_generated_edit.py` runs a `_build/` script against the **bundle** and records before/after hashes.

If the package becomes the sanctioned edit location, none of that observes the edits. If the bundle stays the edit location, then `package_unit.py`'s "no flags, self-contained" promise (#446) is undercut, because the agent must reach back into the bundle anyway.

`check_migration_progress.py` also measures *deliverable output over a window* to distinguish a working agent from a spinning one — pointed at the bundle, an agent correctly editing the package would read as producing **zero** output.

## Options (not decided)

1. **Bundle is canonical.** Package is read-only context; correct `package_unit.py`'s README. Keeps tamper/declaration/progress intact; weakens the self-contained promise.
[...]

Issue #461, opened 2026-09-03
Packaged models carry an ABSOLUTE machine path into the bundle's data/ - 17 of 62 units, so the package cannot refresh, is not portable, and would poison phase 3


Guust-Franssens:
## The defect

A packaged unit's semantic model can contain an **absolute, machine-local path pointing OUTSIDE the package**, into the originating bundle's gitignored `data/` directory.

Verbatim from `HR_Dashboard`'s packaged model (`fabric/HumanResources.SemanticModel/definition/tables/HumanResources.csv.tmdl`):

```
partition 'HumanResources#csv' = m
    mode: import
    source =
        let
            Source = Csv.Document(File.Contents(
              "C:\...\_runs\408-coldrun2-fabric-migration-lab\bundle\data\HumanResources\federated_1uy6vyf0tya8w3119f4041\Extract_Extract.csv"
            ), [Delimiter=",", Encoding=1252, QuoteStyle=QuoteStyle.Csv]),
```

## Why it matters

`package_unit.py` exists (#446) to emit **one self-contained folder** an agent can work from with no flags. For these units it is not self-contained:

1. **The package cannot refresh on its own.** There is no `.pbi/cache.abf`, so Power BI Desktop opens the model **empty** until refreshed - and the refresh needs a file the package does not carry.
2. **The package is not portable.** The path is absolute and machine-local. Copy the package to another machine, or to a customer, and the model cannot load.
3. **It poisons phase 3.** Promoting this to `migrations/{workbooks,datasources}/<slug>/fabric/` ships a deliverable whose model points into `_runs/` scratch - a gitignored, machine-local, prunable directory. That is precisely the *"structurally present and functionally empty"* failure `powerbi-report-gotchas` SS3 already records from a 46-asset estate.

## Measured scope

Estate run 408, engine 2.353.0, 62 packaged units:

| | |
|---|---:|
| `File.Contents` references with an absolute machine path | **22** |
| packaged units affected | **17 of 62 (27%)** |
| total bytes those paths point at | **11.2 MB** |
| `File.Contents` references that are relative/other | 1 |

`HR_Dashboard`'s referenced CSV is 1.1 MB. The whole estate's payload is **11.2 MB** - small enough that carrying it is not the problem.

[...]

PR #462, opened 2026-09-03
Fixes #458: promote_unit.py - the phase 2 to phase 3 ship step

### LINKED ISSUE #458: promote_unit.py: the phase 2 to phase 3 step is manual, and it is the hop where a wrong byPath ships a report with no model

Guust-Franssens:
Fixes #458.

Phase 3 of the pipeline had **no tool** — the phase 2 → phase 3 hop was a manual copy, and
`docs/migration-phases.md` names it as the hop with the highest cost of getting wrong. This adds
`scripts/promote_unit.py`.

```
python scripts/promote_unit.py --package <_runs/NNN-slug/packages/<batch>/<Unit>> --slug <slug>
    [--datasource-slug <ds-slug>] [--bundle <bundle>] [--dry-run] [--force] [--json <file>]
    [--migrations-root <dir>]
```

| exit | meaning |
|---|---|
| 0 | PROMOTED, or a clean `--dry-run` plan |
| 1 | REFUSED_BY_GATE — `check_unit.py` exited non-zero and `--force` was not given |
| 2 | **CANNOT_ASSESS** — the package cannot be read or its shape is ambiguous |
| 3 | REFUSED_CONTENT — structurally present, functionally empty |
| 4 | PROMOTION_FAILED — the copy, the `byPath` rewrite, or a post-copy verification failed |
| 64 | usage error |

2 exists so an unassessable package can never collapse into the clean bucket, and 64 (rather than
argparse's default 2) so "you typed the flag wrong" stays distinguishable from "I could not tell".

## The guards, and the evidence behind each

**The gate is re-run here and refusal is by exit code.** Deliberate duplication: re-running costs
under a second (measured 0.63 s on a real package) and this is where a defect stops being a working
copy and becomes a deliverable. `--force` overrides, and both the override and the observed exit code
are written into the promotion record, so an unchecked promotion can never look checked afterwards. A
gate **timeout** is a refusal, not a pass.

**`byPath` is resolved on the FILESYSTEM.** Negative control, measured against
`powerbi-report-author` **0.1.4** on a copy of the committed `examples/shipping-kpis` deliverable
with the byPath repointed at `../NoSuchModel.SemanticModel`:

```
VALIDATE_EXIT=0
result=succeeded errorCount=0 warningCount=0
PROMOTE_CHECK_CAUGHT: ShippingKPIs.Report: byPath '../NoSuchModel.SemanticModel' does not resolve
  to a semantic model on disk (a wrong byPath opens as a report with NO MODEL and validates clean)
```

Same file, same moment: the validator says `succeeded` / `errorCount: 0` / exit **0** for a report
that cannot open, and this check refuses it. It checks reference *shape*, not *target*.

**Content, not existence** — at the source (refuse before shipping) and again at the destination
(the last hop is the one nobody re-checks). Real pages carrying real visuals; real TMDL `table`
declarations, not merely `.tmdl` files. Measured on a real promotion:
`{"HR_Dashboard.Report": {"pages": 7, "pages_with_visuals": 7, "visuals": 102},
"HumanResources.SemanticModel": {"table_files": 3, "tables": 3}}`.

**Both shapes**, per `powerbi-report-gotchas` §3: model-per-workbook copies the *contents* of
`fabric/` (the folder is named for the WORKBOOK, the model inside for the DATASOURCE, so copying the
folder nests them wrongly); `--datasource-slug` lands the model once under
`migrations/datasources/<ds>/fabric/` and rewrites the reference to
`../../../../datasources/<ds>/fabric/<Name>.SemanticModel`, then resolves it.

**Drift report** (`--bundle`, #460's silent-loss case). On the real end-to-end package it found the
divergence immediately — 6 files only in the package (three agent-added pages and their visuals) plus
a differing `pages.json`. It is **divergence only**, never provenance, and never fatal; with no
`--bundle` it records `not_checked`, which is deliberately distinct from "no drift".

## Verification

Positive controls for both shapes, on synthetic packages in tests **and** on the real reference
estate (`_runs/408-…`): the agent-modified `e2e-HR_Dashboard` package refuses at the gate (exit 1),
promotes under `--force` (139 files), re-runs idempotently, and the `Superstore` package splits into
a workbook report plus a shared `sample-superstore` datasource whose rewritten byPath resolves.

**Mutation table: 26 mutations, 26 caught, control clean.** One mutation initially SURVIVED — deleting
the "no pages" guard still refused, because the "no visuals" guard fired instead: the right verdict
for the wrong reason. Fixed by asserting the finding names the missing pages.

| gate | exit |
|---|---|
| `ruff format --check conftest.py scripts tests .github/skills` | 0 |
| `ruff check conftest.py scripts tests .github/skills` | 0 |
| `pylint scripts` | 0 |
| `set_data_folder.py --check` | 0 |
| `pytest -q tests/test_promote_unit.py` | 0 (32 passed) |

No promoted artifact is committed — every test promotes into its own `tmp_path` via
`--migrations-root`, and `git ls-files migrations` is still 3 (the README navigation files).

## Deliberately not done

`docs/migration-phases.md` still says phase 3 has no tool and that #458 is open. It is owned by
another agent's in-flight branch, so it is left for a follow-up rather than edited into a collision.
`docs/agent-capability-wiring.md` **was** edited, because `check_agent_capabilities.py` exits 1 for an
unindexed agent-facing script — that was a CI gate my change tripped, not a discretionary edit.


Guust-Franssens:
## Update — issue #461 refusal added

`promote_unit.py` now refuses a model that **reads data from outside the deliverable**, with its own
exit code (**5**, `REFUSED_EXTERNAL_DATA_PATH`) and its own named reason. Scanned at the source (so a
refusal ships nothing) and again on the **shipped** model after the copy. `--force` overrides and is
recorded in the promotion record, plus a loud stderr warning.

Detection on the real artifact, no stubs:

```
found 1
  definition/tables/HumanResources.csv.tmdl -> <absolute-path-redacted>\Extract_Extract.csv
```

Findings are **redacted to the leaf** in every artifact — an absolute path embeds a real username and
this repo is public — while the operator gets the full path on stderr, where it is a terminal line
rather than something `set_data_folder.py --check` has to catch later.

### Three things measured rather than assumed — two of them corrected the design

**1. Containment is decided LEXICALLY, never by touching the filesystem.** The first version called
`Path.resolve()`. On a UNC literal naming a host that does not exist, that blocks on SMB name
resolution: measured here, it took the test module from **30 s to 52 minutes** and starved the
`check_unit.py` subprocess into its 600 s timeout. `os.path.normpath` + `is_relative_to` is
millisecond-fast, and it also stops the verdict depending on what happens to be mounted.

**2. A POSIX-absolute literal is ambiguous where a drive-absolute one is not.** Taken at face value it
produced **9 POSIX-only hits on the reference estate, 8 of them false positives** — a Databricks
`expression HttpPath = "/sql/1.0/warehouses/<id>"` (a live connection, in 3 units) and a bare `"/"`
inside a `TableauFormula` annotation. Refusing those would block every live-Databricks promotion. A
POSIX candidate must now also carry a file suffix; that leaves the **1 genuine** hit, a macOS
`/Users/<someone>/…/Global Superstore.xlsx` that a Windows-only rule would have shipped. Named
residual: a POSIX *folder* parameter is missed; its Windows equivalent is not.

**3. A `"://"` URL guard was dead code.** It killed no mutation — a URL is absolute in neither
flavour — so it was deleted rather than kept as protection-shaped decoration. The URL positive
control remains, and now fails only if the detector becomes broad enough to need one.

### ⚠️ The brief's own measurement is low, and I think the brief is wrong here

The brief reports **22 references across 17 units (27%)**. Re-measuring run 408's 62 packaged units:

| scope | references | units |
|---|---|---|
| `File.Contents(...)` only | **23** | **17** — matches the brief |
| all absolute external references | **32** | **26 (42%)** |

The extra 9 arrive through **`Excel.Workbook`** (`.xlsx`) and a **`SourceFolder` M parameter**, not
`File.Contents`, and they are the same defect. So the byte figure in the brief (11.2 MB) is what my
per-reference sum reproduces exactly, but the count appears to have been scoped to `File.Contents`.
**The defect is ~1.5× more widespread than reported** — worth knowing for the packaging half.

### Verification

* **Positive controls** (a refuser passes every negative test): a relative path promotes; an
  absolute path under the deliverable's own `data/` promotes (this is `set_data_folder.py`'s
  convention, and breaking it would have been a silent regression); a URL promotes; a Databricks
  `HttpPath` promotes; a bare `"/"` promotes.
* **Negative controls**: an absolute path outside → exit 5, nothing shipped; a UNC share → exit 5;
  a macOS POSIX data file → exit 5; an absolute path injected only into the **shipped** copy → exit 4.
* **36/36 mutations caught**, control clean (was 34/34 before this addition; the two new ones are
  the POSIX-suffix rule in both directions).
* Gates re-run, all by exit code: ruff format 0 · ruff check 0 · pylint 0 · navigation index 0 ·
  agent capabilities 0 · `set_data_folder --check` 0 · pytest 0 (**44 passed**).

`set_data_folder --check` did fire once during this work, on my own illustrative `/Users/…` strings
in a docstring and a fixture — the gate working correctly. Fixed with its documented `<placeholder>`
form rather than by weakening the pattern.

**Scope:** this is the promotion half only. Packaging is being fixed in parallel and this gate stays
valuable either way — it is the last check before an artifact reaches a customer, and it must not
trust that upstream got it right. The PR grew by ~1 function plus 10 tests; still one script, one
test module.

Guust-Franssens:
# Blind review — BLOCKED (4 fail-open paths, all reproduced)

Reviewer given only issue #458 + the diff, never the author's reasoning. Reviewed clean HEAD `93a7730` against base `16643a3`; worktree left clean.

⚠️ **The 44-test suite passes despite all four.** There are no tests for traversal slugs, report-only fake targets, missing `pages.json`, empty visual documents, or concatenated absolute paths. That is the finding behind the findings.

## 1. Slug traversal escapes the migration root — and can DELETE outside it

`promote_unit.py:469-511,683-712`. Only whitespace is rejected; path components and absolute paths are accepted.

```
traversal slug: exit=0
PROMOTE: PROMOTED Wb -> ..\..\escaped
record .../traversal/escaped/promotion-record.json
traversal_target_exists=True
```

`execute_plan()` removes an existing destination before copying, so a crafted or mistyped slug can delete a directory outside the migration root. Highest severity of the four.

## 2. `byPath` can resolve to something that is not a semantic model

`promote_unit.py:253-272,483-489,548-577`. `verify_bypath()` requires only an existing directory containing some `definition/`.

```
report-only fake model: exit=0
PROMOTE: PROMOTED package -> wb (8 files)
fake_has_model_tmdl=False
```

Reproduced under a PASS-gate control and under `--force`. The intended dangling-reference control is correct and should stay — for contrast, same moment:

```
powerbi-report-author: result=succeeded, errorCount=0, exit=0
promote verify_bypath: PromotionFailed: ... does not resolve to a semantic model on disk
```

So the missing-target case is covered; criterion 4 is satisfiable by an unrelated directory.

## 3. The report content guard is still an existence check

`promote_unit.py:282-304`. It never reads `pages.json` and never parses `page.json`/`visual.json`; any file at `visuals/*/visual.json` counts.

```
missing pages.json: exit=0     PROMOTE: PROMOTED Wb -> wb
empty visual JSON:  exit=0     PROMOTE: PROMOTED Wb -> wb
```

Both promotion records then **claimed** `{"pages": 2, "pages_with_visuals": 2, "visuals": 4}`. A record asserting content that was never verified is worse than no record. Directly violates criterion 5.

## 4. Concatenated literals bypass the #461 external-path scan

`promote_unit.py:322-398`. Each quoted literal is classified independently, so this evaluates to `C:\secret\data.csv` and passes:

```m
File.Contents("C:" & "\secret\data.csv")
```

```
literal_classification=[('C:', False), ('\\secret\\data.csv', False)]
exit=0
external_data_paths={'source': [], 'shipped': [], 'forced': False}
```

Promoted **without `--force`** under a PASS gate — the exact outcome #461 forbids.

## Non-blocking follow-ups

- `verify_bypath()` still calls `Path.resolve()` on the untrusted PBIR literal (`:570`). A nonexistent-host probe returned in 0.487 s so the multi-minute stall did not reproduce, but the unsafe call is present and was already removed elsewhere for this reason.
- **Partial copies survive failure**: an injected `copy2` failure exits 4 leaving `report_already_written=True, record_written=False`. Stage and swap atomically, or clean up.
- Deleting both manifests plus `--force` promotes with `engine_version=null` — clarify whether criterion 6 permits it.
- A relative `..\..\outside\data.csv` classifies `False`; Power Query's runtime base was not established, so recorded as a residual rather than a blocker.

## Mutations — all caught for the right behavioural reason

`assert 0 == 1` (ignore non-zero gate) · `assert 4 == 0` (shared byPath off by one) · `DID NOT RAISE PromotionFailed` (disable filesystem byPath check) · `assert 0 == 3` (accept zero visuals) · `assert 0 == 3` (accept zero table declarations) · `assert 0 == 5` (disable external-path detector) · dry-run created the deliverable · idempotency `assert 4 == 0` · `assert 0 == 2` (collapse CANNOT_ASSESS) · `assert None == '2.353.0'` (drop engine version).

## Gates — all exit 0

`ruff format --check` · `ruff check` · `pylint scripts` · `check_navigation_index` · `check_agent_capabilities` · `set_data_folder --check` · `pytest tests/test_promote_unit.py` (44 passed).

## Could not verify

The agent-modified package path in the brief was wrong (it is under `_runs/408-.../scratch/`, not `C:\tfmig\scratch`) — my error, not the reviewer's. No Power BI Desktop render test. Worst-case UNC latency not reproduced.

Guust-Franssens:
## Round 3 — all four BLOCKING findings fixed, with regression tests

Every reproduction was confirmed against the pre-fix code and re-run against the fixed code. The
reviewer's central point stands and was the most useful thing in the review: **the 44-test suite
passed despite all four, because there were no tests for any of these shapes.** There are now 99
tests and 52 mutations.

### Per finding — what changed

| # | fix | file:line |
|---|---|---|
| 1 | `slug_problem()` rejects separators, `..`, absolute paths, drive letters, reserved device names and control chars → **exit 64**. `_assert_contained()` refuses any destination not lexically inside the migrations root → **exit 2**. `execute_plan()` re-checks before touching the filesystem. **Both** `--slug` and `--datasource-slug` are validated. | `promote_unit.py:344-377`, `:700-712`, `:944-950` |
| 2 | `verify_bypath()` now requires the `.SemanticModel` suffix, containment in the migrations root, and the **same** `check_model_content()` criterion 5 applies to a shipped model — reused, not re-implemented. | `promote_unit.py:760-800` |
| 3 | Every PBIR document is parsed: `pages.json` must declare a non-empty `pageOrder`, a page must name itself, a visual must declare `visual.visualType` (or a `visualGroup`). Unreadable input is a new `ContentCheck.unassessable` bucket → **exit 2**. | `promote_unit.py:270-320`, `:380-425`, `:1092` |
| 4 | The scanner classifies the **joined** value of every `&` run, not only each literal. | `promote_unit.py:440-460` |

**Reproductions, re-run against the fixed code:**

```
BLOCKING 1: exit=64  --slug must be a single path component (no '/' or '\')
            victim intact = True   escaped target exists = False
BLOCKING 2: exit=4   byPath '../Fake.SemanticModel' resolves to a folder that is not a working
                     semantic model: no definition/model.tmdl; no definition/tables/ directory
BLOCKING 3a: exit=2  pages.json is missing - page order is unreadable        shipped = False
BLOCKING 3b: exit=2  v1/visual.json is not readable JSON: Expecting value    shipped = False
BLOCKING 4:  scanner values = ['C:', '\...\secret\data.csv', 'C:\...\secret\data.csv']
             joined value is a local path = True   external_data_paths found = 1   shipped = False
```

**On finding 4, the choice you asked me to name:** I implemented **literal-concatenation joining**,
not the CANNOT_ASSESS fallback. Reason: an *identifier* concat (`SourceFolder & "\x.csv"`) is not a
new hole, because the parameter is itself `expression SourceFolder = "<abs>"` in `expressions.tmdl`,
which this scan reads like any other file — measured, that is how **9 of the 32** estate findings
surfaced, and there is a test pinning it. Refusing every unresolvable partition would refuse exactly
the legitimate packages the #461 packaging fix is about to produce. The residual is a path built from
a genuinely runtime value, which is not a fixed machine path.

**On `--force`:** confirmed in code and now pinned by `test_force_cannot_bypass_the_mandatory_content_checks`
— content runs *before* the gate in `assess()`, and `--force` guards only the gate and the #461 scan.

### Non-blocking follow-ups — all four addressed

* `verify_bypath()` is now lexical (`os.path.normpath`); no `resolve()` on the untrusted literal.
* **Rollback**: a failed copy *or* a failed verification now restores the previous deliverable
  (destinations are moved aside, not deleted). Two tests: nothing half-shipped, and a re-promotion
  that fails verification leaves the *previous* deliverable intact.
* Missing receipt now records `engine_version_source: "UNAVAILABLE: no engine-output-receipt.json"`,
  distinct from `"UNAVAILABLE: receipt declares no engine.version"`. Permitted, never silent.
* Relative `..\..\outside\data.csv` remains a **recorded residual** — Power Query's runtime base is
  still unestablished, so guessing would be worse than naming it.

### Mutation table — 52/52 caught, control clean

The three that initially **SURVIVED** are worth reporting, because all three were the trap you
warned about (a *different* guard firing), and fixing them changed the tests, not the code:

| survived mutation | why it survived | fix |
|---|---|---|
| `elif not target.is_dir(): → elif False:` | the reused model-content branch caught it instead | retargeted the mutation at the whole `_refuse_bad_model_target` call |
| page `name` check → `if False:` | the fixture was unparseable JSON, caught by the parse guard | added `test_a_page_that_declares_no_name_cannot_assess` (parseable, unnamed) |
| join-everything mutation | fixture had no fragments that could form a path | rewrote the positive control to hold `"C:"` and `"\Windows\Temp\evil.csv"` **unconcatenated** — a naive whole-file join now manufactures a path and the test fails |

Representative kills (all on behavioural assertions):

```
B1 slug: accept any slug                       E Failed: DID NOT RAISE SystemExit
B1 containment: no guard inside execute_plan   E Failed: DID NOT RAISE CannotAssess
B2 byPath: accept any directory, as before     E AssertionError: assert 0 == 4
B2 byPath: skip the reused model content check E Failed: DID NOT RAISE PromotionFailed
B3 content: unparseable → permissive stub      E AssertionError: assert 0 == 2
B3 content: count files not parsed documents   E AssertionError: an empty file
B4 concat: judge each literal alone            E AssertionError: assert 0 == 5
B4 concat: over-eager joining                  E AssertionError: assert 5 == 0
rollback: leave the half-shipped copy behind   E AssertionError: assert not True
```

Two **positive-control guard** mutations are in the table deliberately (`slug_problem` returning
"always bad"; the #461 scan flagging everything) — a refuser passes every negative test, so the
positive controls must themselves be killable.

### Positive controls

* Both shapes on the **real** reference estate: `e2e-HR_Dashboard` → exit 0, 139 files, 7 pages /
  102 visuals parsed, `bypath_verified.target = HumanResources.SemanticModel`,
  `engine_version 2.353.0 (engine-output-receipt.json)`; `Superstore --datasource-slug
  sample-superstore` → exit 0, model under `datasources/`, byPath
  `../../../../datasources/sample-superstore/fabric/Superstore.SemanticModel` verified; idempotent
  re-run → exit 0.
* An absolute path under the deliverable's own `data/` still promotes; Databricks `HttpPath` and a
  bare `"/"` still promote; a relative path still promotes; ordinary slugs (`wb`, `customer-acme`,
  `a.b.c`, `Superstore_2024`) all accepted.

### Gates, by exit code

| gate | exit |
|---|---|
| `ruff format --check conftest.py scripts tests .github/skills` | 0 |
| `ruff check conftest.py scripts tests .github/skills` | 0 |
| `pylint scripts` | 0 |
| `set_data_folder.py --check` | 0 |
| `pytest -q tests/test_promote_unit.py` | 0 — **99 passed** |
| mutation harness (throwaway, not committed) | 0 — **52/52** |

### Two things in the review I'd push back on, with evidence

1. **`../escaped` does not escape the migrations root** — it normalises to `<root>/escaped`, the
   wrong *tier* but still contained. My first containment test asserted a refusal there and failed
   honestly. That is correct layering, not a gap: `slug_problem` refuses any slug with a separator,
   and containment guards the root. Both layers are now tested separately, and the boundary is
   recorded in `test_containment_is_anchored_at_the_ROOT_not_at_the_workbooks_tier` rather than left
   implicit. The genuinely escaping `../../escaped` **is** refused.
2. **The brief's `_runs` path correction was right and mine was wrong** — noted, and the real
   artifact under `_runs/408-…/scratch/e2e-HR_Dashboard` is what the positive controls above ran on.

One knock-on worth flagging: the module hit pylint's repo-wide `max-module-lines = 1200`. I trimmed
prose rather than raise the budget or split the file — no fact was dropped, but the docstrings are
denser than they were.

PR #463, opened 2026-09-03
feat(package): self-contained packages (#461), the PACKAGE is canonical (#460), plus six cold-run defects

### LINKED ISSUE #460: Two documented edit locations: AGENTS.md says <bundle>/pbip/, package_unit.py says <package>/fabric/ - and the package is a copytree, so promotion silently loses agent work

Guust-Franssens:
## Summary

One design decision (#460) and six package defects, all from the 2026-09-03 cold run in which an agent was given **nothing but a `HR_Dashboard` package folder** and took `check_unit.py` from exit 4 (`PRECONDITION_FAILED`, 4/7 pages) to exit 1 (`FINDINGS`, page pairing **PASS 7/7**). Its report is the work list.

## PART 1 - the PACKAGE is canonical (Fixes #460)

`AGENTS.md:778` said agents edit `<bundle>/pbip/`; the generated package README said `<package>/fabric/`. `package_unit._copy_fabric` is a `shutil.copytree`, so the two are byte-identical at packaging time and diverge only by being edited - and promoting from the wrong one silently discards every agent edit.

Decision implemented: **when you work from a package the package is canonical**; the bundle stays canonical when you work from a bundle directly. Deliberately cheap - declared-edit tooling (`declare_generated_edit.py`, `check_migration_progress.py --tamper`) is **scoped to bundle work** rather than taught to follow packages, because the cold-run agent reported its package edits were "undeclarable by construction" already. The engine baseline is **not** shipped into the package.

`AGENTS.md`'s working-copy row is inside the synced block, so `sync_agent_conventions.py` re-ran and all five personas carry it. Funded by deleting a duplicated *"keep `<bundle>/reports/` pristine"* sentence (the row above and the `byPath` note both already carried it): net **+86 chars**, largest persona **29,655 -> 29,741** of a 30,000 cap.

Acceptance criterion 2 - the guard that stops this recurring - is a **cross-document derivation, not two copies of a string**: it reads which row the generated README marks `**edit here**`, takes that row's own path, and requires `AGENTS.md`'s working-copy row to name it under `<package>/`. Change either side alone and it fails.

## PART 2 - the six package defects

| # | fix |
|---|---|
| 1 | `migration-spec.schema.json` is **shipped** into every package, byte-identical to `docs/migration-spec.schema.json`. The agent invented `{id, category, objects, detail, owner, status}` and lost all six `limitations_encountered` entries; the real item is exactly `item`/`issue`/`severity`/`stage` under `additionalProperties: false`. |
| 2 | `oracle/*/data/*.csv` is promoted to lead as the **numeric oracle**; the PNG-vs-SVG paragraph (with its estate-specific `122 <text> elements`) is cut to one line, keeping the zero-text caveat. |
| 3 | The page-pairing contract is stated in one sentence - **verified against `check_unit.actual_pages`, `page_expectation` and `_page_visual_count`**, not taken on trust. |
| 4 | Both documented gate commands took a unit **name**. Measured: `check_reference_readiness.py HR_Dashboard` exits **2** (`is not a directory`) - a usage error, not a verdict. Now the folder path; the new guard **runs both command lines the README prints**, with the bare name as its negative control. |
| 5 | A short `UNFIXABLE FROM THIS PACKAGE` section names the provenance ceiling: `origin.match: "name_only"`, the remedy needing Server credentials **and** the three `origin` fields `scope.dropped_fields` strips, and the plain consequence - `check_reference_readiness.py` can never exit 0 from the package alone. |
| 6 | `validate_spec.py` no longer rewrites the file it validates. Repair is opt-in behind `--repair`; `--check` is kept as an accepted alias of the new default so CI and the three personas keep working. |

**Generated README: 3,350 -> 3,844 bytes on disk, 38 -> 42 lines** while absorbing five new obligations and losing a ~1,000-byte paragraph.

## Verification

`python tests/mutate_package_unit.py` -> **32/32 caught, negative control SURVIVED, exit 0**, including eight new mutations covering every claim added here. `validate_spec` was mutation-tested by hand, both halves, each failing on its behavioural assertion.

All gates green by exit code: `ruff format --check` 0, `ruff check` 0, `pylint scripts` 0, `sync_agent_conventions.py --check` 0, `check_navigation_index.py` 0, `check_agent_capabilities.py` 0, `set_data_folder.py --check` 0, targeted `pytest` 0 (250 passed).

## Where the brief was wrong

Item 5's *"reference readiness can never reach exit 0 from the package alone"* is true as shipped but **too strong as a mechanism claim**. Measured: copying the package's own `assets/*.twbx` into a `source/` subdirectory and running `capture_tableau_reference.py` admits 23 records **by sha256**, confirms their revision with no Server credentials, and moves two of the four unverifiable pages to `ready`. The two **dashboard** pages stay unverifiable for a different reason entirely - `UNVERIFIABLE - name only; scope worksheet cannot satisfy a dashboard page`, because the embedded-thumbnail provider only produces worksheet-scoped records. So the shipped wording states the consequence for the package as delivered and does not claim the stronger mechanism.

Two related observations, **not fixed here** (out of scope, no issue filed yet): `capture_tableau_reference._source_workbook` scans only `source/` and `in/`, never a package's `assets/`; and where both an oracle LUID render and a sha-confirmed local reference exist for one object, the LUID one is what the verdict reports.


Guust-Franssens:
**Correction landed in d345c62** - PART 2 item 3's sentence was accurate about pairing and incomplete about consequence.

Re-verified against scripts/check_unit.py, not against the summary:

- `_page_visual_count` (:1213) counts `visuals/**/visual.json` and says why: *"without it renaming an empty page to an expected page's title certified it as rebuilt"*.
- a zero-visual page that is not the engine's crash-guard placeholder is `blank` (`_zero_visual_pages`, :1745), and `blank` is one of the **four** conditions forcing `STATUS_PRECONDITION_FAILED` (:1527). So it **FAILS**, rather than merely going uncredited - which is what "at least one visual" alone reads as.
- the expected set is every dashboard **plus every ORPHAN worksheet** (`_spec_pages`, :853-854); `expected_pages`' docstring measures the cost of the narrower reading - 19 of 43 workbooks in a real 2.339.0 estate have zero dashboards.

**Third reading, on the ids** - I dropped that clause rather than restating it, because there are two distinct ids and they are not interchangeable:

| id | what it is |
|---|---|
| `migration-spec.json` `dash.hr_details` / `ws.age` | what `_spec_pages` uses for the orphan test and its duplicate-id refusal |
| `page-HRDetails83013a67`, what `check_reference_readiness.py` PRINTS | the predicted **PBIR page directory id**, recomputed from the object NAME by `engine_page_id` (:186) reproducing the engine's sanitiser |

So *"page ids come from migration-spec.json and are printed by check_reference_readiness.py"* conflates them, and a spec id cannot disambiguate a rendered page in any case: `_claim` returns `AMBIGUOUS` on the **name**, so nothing the report author does to the PBIR page resolves it.

Generated README **3,802 -> 3,794 bytes**, still 42 lines. Mutation campaign now **34/34 caught, negative control SURVIVED**; the two new semantic mutations (soften `blank and FAILS` to "simply not credited"; narrow the expected set to "every dashboard in the workbook") each fail `test_the_readme_states_the_page_pairing_contract` on its own assertion. All gates 0.

Guust-Franssens:
## PART 0 landed in 7ff223d — `Fixes #461`

Re-measured across all 67 packaged units of run 408 and it reconciles exactly with your report: **23** absolute `File.Contents` refs, **17** units, **11.19 MB** per-ref (17 distinct files, 4.69 MB actual), **1** ref whose file is absent from this machine (the `/Users/bs/...` macOS path baked into `global_superstores_db`'s workbook).

### ⚠️ Your step 2 does not work, and I did not do it

**Power Query rejects a relative `File.Contents` argument outright.** A relative rewrite would produce a model that refreshes **nowhere** — strictly worse than one that refreshes on a single machine. The documented workaround is a folder parameter, and **this repo already ships exactly that**: sixteen committed, Desktop-verified models under `examples/*/fabric/*.SemanticModel/definition/expressions.tmdl` use `expression SourceFolder = "..." meta [IsParameterQuery=true, ...]`, and `scripts/set_data_folder.py` already localizes, sanitizes and CI-gates the name `DataFolder`.

So: sources copied to `<package>/data/<parent>/<file>`, partitions repointed to `File.Contents(DataFolder & "<relative>")`, `DataFolder` declared in `expressions.tmdl` naming the package's own `data/`. **No new script**, and a unit promoted to `migrations/<slug>/fabric/` is covered by the existing privacy gate for free.

### Your two cautions

**(a) size** — largest source observed is **1.33 MB**, so nothing needs a ceiling today. There is one anyway at 256 MB, and exceeding it is a **loud recorded omission** (manifest + package note + `handover.md`), never a silent skip. Both directions are tested from one fixture by moving the ceiling, not the data.

**(b) no existing gate sees it** — confirmed, so the new tests are the only guard. **41/41 mutations caught, negative control SURVIVED.** The campaign earned its keep twice: *"drop the data/ row"* **SURVIVED** at first because the shared fixture's model imported nothing so the map guard never saw `data/`, and again because that guard tested a **bare substring** and the README says `oracle/*/data/*.csv` elsewhere. Both fixed — the fixture now imports rows, and the guard matches a backticked table key.

### Two traps I hit, each now pinned by its own test

- the parameter must name the **final** package, not the `.<unit>.staging` dir `replace_dir` renames away — a staging value looks right in the file and resolves to nothing;
- `<parent>/<file>` is readable but **not unique** (every engine extract is `.../<table>/federated_<hash>/Extract_Extract.csv`), so a collision digests the full original path. Overwriting would silently repoint one partition at another table's **rows**, which no structural check can see.

### Verified on the real `HR_Dashboard`

1.13 MB extract shipped; partition reads it through the parameter; `check_datamodel.py` **exit 0** with the AMO `TmdlSerializer` accepting the model (2 M expressions, 7 TMDL documents, 0 problems); **zero** absolute paths resolving outside the package — the one that remains is `DataFolder`, naming the package's own `data/`. `check_reference_readiness.py` (1) and `check_unit.py` (4) unchanged from before the fix, so no regression.

Gates all **0**: ruff format/check, pylint — which exited **24** on `too-many-lines` + `too-many-arguments` while printing `10.00/10`, fixed rather than silenced — sync `--check`, nav index, agent capabilities, privacy `--check`, pytest 256 passed.

Generated README **3,794 → 4,038 bytes** for one new shipped artifact.

**Items 2 and 6 were already landed before your message**, so I kept them rather than dropping them — there was nothing to remove and no room to reclaim.

Guust-Franssens:
## Broadened to the general shape — `0f7ac74`

Not materially larger, so no split. Re-measured independently across all 67 packaged units and it reconciles **exactly** with your table: 22 `File.Contents` / **31** Windows+UNC / **32** with the one macOS POSIX file. The 9 my narrow scan missed are one shape, and it is every datasource-only unit:

```
expression SourceFolder = "<bundle>\pbip\<Unit>\<Unit>.Data" meta [IsParameterQuery=...]
... Source = Excel.Workbook(File.Contents(#"SourceFolder" & "\Sample - Superstore.xlsx"), null, true)
```

The scan is now the general question — *any* absolute path literal in *any* packaged `.tmdl` — with **two repairs**, because the shapes need different ones:

| shape | repair |
|---|---|
| `File.Contents("<file>")` | ship to `data/<parent>/<file>`, repoint at a new `DataFolder` parameter |
| folder parameter | keep the parameter, move only its **value**, copy the folder's files alongside |

The folder case reuses the existing parameter rather than replacing it, and `_moved_folder_value` preserves the original's **trailing-separator convention** — the partitions' concatenation was written against it, so adding or dropping one separator silently breaks every path built from it.

## Your three findings, all adopted, each with a test

1. **Lexical containment.** `normpath` + `is_relative_to`; `Path.resolve()` is never called on one of these literals. I extended it one step: `_classify_source` also refuses a UNC source **unprobed**, because `is_file()` blocks on SMB name resolution exactly as `resolve()` does — that was still a live hang in my first version. The test is **time-boxed at 5 s**, since the whole failure mode is that it doesn't come back.
2. **POSIX needs a file suffix.** Reproduced your 9-hits/8-false split precisely: Databricks `HttpPath = "/sql/1.0/warehouses/<id>"` in 3 units, bare `"/"` in a `TableauFormula` annotation in 2 more. Suffix rule keeps the one genuine `.xlsx`, drops all 8. Both directions asserted.
3. **Under the package is legitimate.** Rule is *"absolute AND not under the destination"*. There is a mutation that drops the containment half specifically, because that version condemns its own output — it's caught.

**Redaction adopted**: findings carry the leaf only. Your convention proved itself mid-build — the repo's privacy gate (`set_data_folder.py --check`, exit 1) failed on a test fixture of mine that quoted the real `/Users/<name>/…` path out of the customer workbook. Fixed, not suppressed.

## Estate-wide result: 32 external literals → 1

| | |
|---|---:|
| units now carrying their own rows | **25** |
| bytes shipped | **9.98 MB** |
| literals resolving inside their package | **25** |
| literals still external | **1** |

The remainder is the macOS `.xlsx` whose bytes are not on this machine at all — physically unshippable, recorded once as a loud omission (leaf-only). Your 11.2 MB reproduces as the per-reference figure; 9.98 MB is what actually lands, since a package copies each distinct source once.

**All 62 packaged models still load**: `check_datamodel.py` exit 0, AMO `TmdlSerializer` accepting **272 M expressions across 549 TMDL documents, 0 problems**.

## Verification

- Mutation campaign **45/45 caught, negative control SURVIVED**. It also caught **two of its own anchors going stale** under this refactor — reported as `ANCHOR-SNIPPET x0`, never as a pass, which is the refusal it was built for.
- Gates all **0** by exit code: ruff format/check, pylint (which exited **8** on `too-many-locals` while printing 10.00/10 — fixed by extracting `_ship_file`/`_ship_folder`/`_moved_folder_value`, not silenced), sync `--check`, nav index, agent capabilities, privacy `--check`, pytest 260 passed.
- README **4,038 → 4,025 bytes**, 43 lines.

`promote_unit.py` untouched — this is the repair half; your exit 5 refusal is the last gate, and they are not alternatives.

Guust-Franssens:
## Blind review round 1 — BLOCKED

Reviewed at tip `0f7ac74` with an empty `git status --porcelain`. Diff: 14 files, +1,176/−103. Reviewer had the requirements and the diff only, never the author's reasoning.

### Verified successes (worth recording, so the fix round does not regress them)

- Reproduced the original inventory exactly: **32 external literals across 26 of 62 packages**.
- All 67 copied fixture units packaged; 62 working models; 25 shipped data; **33 parameterized partition references resolved**; every shipped file non-empty and SHA-256-matching its source.
- `check_datamodel.py` handed all 62 repaired models to AMO/TMDL: **exit 0, 0 parser errors**.
- Injected assembly failure preserved the previous package byte-for-byte and removed staging.
- UNC classification returned in 0.000098 s without probing SMB.
- Generated parameters name the final directory, never `.staging`.

### Blocking (7)

| # | Defect | Location |
|---|---|---|
| 1 | Cross-shape destination collision **corrupts rows** — a folder partition reads another table's rows; exit 0, AMO clean | `package_unit.py:921-943,1120-1128,1164-1169` |
| 2 | Folder localization ships **unrelated private files** — an unreferenced secret was copied, manifested, shipped, exit 0 | `:1092-1106,1120-1128` |
| 3 | Documented relocation guidance produces **broken paths**, and ignores `PackageDataFolder` | `:845,1109-1117`; `set_data_folder.py:24,64-69` |
| 4 | Packaging turns a **loadable model unloadable** by appending a duplicate `PackageDataFolder`; exit 0 then AMO exit 1 | `:1003-1006,1009-1031` |
| 5 | Suffixless POSIX folder paths **silently missed** — no folder shipped, **no omission recorded**, exit 0 | `:946-959` |
| 6 | Repackaging **silently destroys canonical package edits** — the exact location this PR declares canonical | `:1273-1285,1294-1323` |
| 7 | Canonical-location docs still say the decision is **unsettled** and route agents to the stale advice (Requirement B) | `README.md:212`; `docs/migration-phases.md:150,206,210,227,355`; `docs/INDEX.md:75` |

### The mutation harness gave false credit

It reported **45/45 caught, exit 0**. Controlled inspection found otherwise:

- `copy2` → `pass` raised `FileNotFoundError` in production **before the assertion ran** — scored CAUGHT.
- Zero-byte destination instead of a copy — **SURVIVED**.
- Removing the general post-rewrite scan — **SURVIVED** (87 passed, 1 skipped).
- Declared-edit tooling scoped package-only — **SURVIVED**.

Being fixed by porting the `PROBES` / `probe_is_trivial` pattern from `tests/mutation_harness_running_total_axis.py`, which proves each mutation actually executes its intended behaviour and that the failure names the specific assertion.

### Gates

All targeted gates exit 0 (ruff, pylint, sync, navigation, `set_data_folder --check`, changed tests 117 passed). Full `pytest -q` exits 1 on **six pre-existing** engine-pin failures in unchanged files — the engine has moved to 2.353.0 while the pins expect older. Not branch-caused.

Fix round dispatched.

Guust-Franssens:
## Blind review at `c41994bb` — REQUEST CHANGES, 6 blocking

Independent review, given only issues #460/#461 and the diff. CI is green on this exact SHA and every local gate exits 0 — none of these are visible to either.

⚠️ **Scope note:** my brief called this a small change. It is **18 files, +2,934/-151**, spanning package assembly, path handling, binding, `validate_spec`, generated personas, docs and tests. The reviewer corrected me; that correction stands.

### B1 — #460 is not actually closed

Packages are declared canonical, but progress/tamper/declaration still inspect the **bundle**. An edit to a canonical package produced:

```
exit=0   state=CLEAN
3 generated artifact(s) are pristine against their engine-run hashes
```

Progress mode: `exit=2 state=SILENT deliverable_count_seen=0`. And `docs/migration-phases.md:147-151, 208-228, 355` still says the edit location is undecided.

### B2 — a report pointing at a missing model packages as self-contained

```
package_exit=0   manifest_model=null
byPath=../../Shared/Shared.SemanticModel
target_exists_inside_package=false
package_self_contained_flag=true
```

This is a supported shape (shared datasource fixtures exist), so it is not exotic.

### B3 — a declared SHA-256 mismatch is ignored

`input_manifest.staged_input_path` is interpreted with **host** path rules, so a POSIX literal became `C:\...` on Windows and the wrong bytes shipped:

```
expected SHA=5d65d756...1c02bab
shipped  SHA=54a6036a...b1efbc      package_exit=0
```

The manifest digest exists precisely to catch this and is not enforced. Zero test coverage.

### B4 — a customer host path survives into shipped TMDL

`File.Contents("/Users/review-canary/private-data")` is classified `unclassified`, reported, and **left verbatim**:

```
package_exit=4   self_contained=false   literal_still_in_shipped_tmdl=true
leaking file = fabric\Book.SemanticModel\definition\tables\Imported0.tmdl
```

Exit 4 is correct, but the leaking package is still written. Refuse to write, or contain the literal.

### B5 — malformed and missing inputs collapse into success

| input | observed |
|---|---|
| truncated `report.json` | **exit 0**, `OK Book [unclassified]`, no notes |
| truncated handover | **exit 0**, handover absent |
| missing source asset | **exit 0**, `OK Book` |
| zero units | **exit 0**, `packaged 0/0` |
| empty known unit | exit 1 ✅ |
| truncated existing manifest | exit 3 ✅ |

The repo's dominant defect class. "Cannot assess" must be its own blocking state.

### B6 — path traversal writes outside `--out`

Unit names from `report.json` are not constrained to one path component:

```
workbook name = "..\escaped-package"
requested_out = ...\packages\out
written       = ...\packages\escaped-package
written_is_inside_out = false
```

A source-controlled name escaping the output root is a security issue, not a tidiness one.

### Non-blocking

- **N1** — the committed mutation campaign is **stale**: exit 1, 40/45 caught, five anchors no longer exist. Proof-maintenance gap, not a product defect.
- **N2** — after binding, `package-manifest.json` still reads `state: unbound` and its digest immediately reports `expressions.tmdl` as edited. Misleading but fail-closed.

### Confirmed working — do not churn

Move the package, then delete the bundle, assets, original extract and oracle: **shipped rows still resolve beneath the moved package.** That is the core #461 claim and it holds. Missing/absent/truncated oracle all correctly reach exit 1 FINDINGS. A locked source gives exit 1 `PermissionError` with no final package.

Only one test was removed and it was replaced by a stronger one — no weakening found.

### Gates

pytest **0** (103 passed, 1 skipped) · ruff format **0** · ruff check **0** · pylint **0** · mutation campaign **1** (stale anchors) · CI green on the reviewed SHA.

### Note for whoever picks this up

This branch is **12 commits behind master** and does not contain `scripts/promote_unit.py` (merged in #462), so phase-2→phase-3 promotion could not be exercised. Merge master first.

**These findings propagate:** #487 has merged this branch as its base, so it inherits all six.

Guust-Franssens:
## Round-2 verification — APPROVE (verified by the maintainer, directly)

The dispatched round-2 reviewer returned an **empty response** after 416s, so I verified the load-bearing claims myself rather than merge on the author's summary.

**B6 (path traversal) — genuinely fixed, and by two independent guards.** I disabled `unit_name_problem()` entirely:

```
mutation: unit_name_problem() -> return None
result:   1 failed, 7 passed
          FAILED test_the_unit_name_guard_names_the_problem_rather_than_sanitizing_it
```

The 7-shape parametrized traversal test **still passed** with that guard gone — because `assert_package_destination()`'s resolved-containment tripwire catches it independently. So the escape is blocked twice over.

⚠️ Non-blocking observation worth a follow-up: that means the parametrized test **cannot tell which guard fired**. This repo has hit that exact vacuity before — fail-closed-heavy code where any guard satisfies a broad "it refused" assertion. Asserting *which* guard refused would make it precise.

**Verified independently:**

| claim | result |
|---|---|
| no test deleted or weakened | `git diff --numstat 8a4e881..HEAD -- tests/` = **647 insertions, 0 deletions** (175 + 472) |
| targeted suite | **135 passed, 1 skipped, rc=0** |
| CI on the exact tip `88c48de2` | **success**, both jobs |
| B5 regression tests | zero-units and cannot-assess-outranks both present and passing |

**Carried forward as non-blocking, not fixed here:**

- **N1** — the committed mutation campaign still exits 1 (40/45, five stale anchors). The author proved these were already dead at the base commit `8a4e881`; pre-existing proof-maintenance gap.
- **N2** — `package-manifest.json` still reads `state: unbound` after binding. Fail-closed.
- The traversal-test guard-attribution point above.

**Author's own disclosures, accepted as honest scope limits:** nothing was run against a real estate bundle (synthetic fixtures + committed `minimal.twb`); Linux was untested locally, with CI as the first POSIX execution; the 67-unit run-408 numbers in docstrings are inherited, not re-measured.

Merging: two customer-blocking PRs (#484 path budget, #487 batch resilience) declare this branch as their base and cannot ship until it lands.

PR #475, opened 2026-09-03
fix(oracle): never advise raising the client REST pin above a server that cannot move

### LINKED ISSUE #474: Oracle capture tells a customer to raise TABLEAU_REST_API_VERSION above their server's ceiling - a remedy that cannot work, in the log AND in the evidence file
A customer on an on-prem Tableau Server queried their own site and reported, verbatim:

> Confirmed — queried `/api/3.21/serverinfo` ... productVersion: **2025.3.3**, restApiVersion (server ceiling): **3.27**
> Yes, this server does not support SVG. Per `docs/reference-capture.md`, `?format=svg` needs REST 3.29+ (Cloud June 2026 / Server 2026.2). This server only advertises 3.27, so SVG requests would get HTTP 400 "SVG export requires API version 3.29 or later". The next-best rung is PDF (needs only 2.8+, universal) — so if we ever run `--reference-best`, it should resolve to `pdf`, not `svg`.

They are correct, and `--reference-best`'s ladder already handles them correctly. **What we do not handle is what we TELL them when SVG fails.**

## 1. We emit a remedy that cannot work, and never check the one number that decides it

`scripts/tableau_oracle_manifest.py` (`_log_blocked_and_stale`):

```
N view(s) could not produce SVG: this site's REST API version is below 3.29.
Set TABLEAU_REST_API_VERSION=3.29 in .env and re-run; the PNG and PDF captures
are unaffected (they reach back to API 2.5 and 2.8 respectively).
```

For this customer that is **arithmetically impossible**: raising a *client* preference cannot make a 3.27 server export SVG. It is also the loudest, most actionable-looking line in the run.

The function **already receives `capability_report`** and uses it on the very next line, so the advertised ceiling is in scope and is being ignored.

## 2. The same false remedy is written into the evidence file, per view

`scripts/capture_tableau_oracle.py` (`_capture_render`) stamps

```python
record["remedy"] = f"set TABLEAU_REST_API_VERSION={SVG_MIN_API_VERSION} or later in .env"
```

— a flat constant with no access to the advertised ceiling — and it is serialised into `oracle-manifest.json`. The status label `unsupported_api_version` is also ambiguous: it does not distinguish *"our client pin is too low"* (fixable by the customer) from *"this server can never do it"* (not fixable at all).

## 3. The third state is real, not hypothetical

`capability_report` is only non-`None` when `--reference-best` is passed, so a plain `--svg` run has **no advertised ceiling at all**. That is a genuine *cannot assess* state, and this repo's own rule is that unassessable input must never collapse into a confident bucket.

## 4. The assessment, where this belongs, never asks

`scripts/assess_estate.py` *sends* a REST API version on every call but never probes `/serverinfo` and never reports the server's ceiling or product version. The assessment is the FIRST thing an operator runs against a new site, and the render ceiling is a property of the *site* — so an operator should learn *"this site tops out at PDF"* there, rather than discovering it much later as a capture-time warning that then misadvises them.

## What good looks like

Every place we say why SVG failed resolves to exactly one of three states, and never emits a bare confident knob outside the first:

- **advertised ≥ 3.29** — raise `TABLEAU_REST_API_VERSION`; only claim the tier *works* where the ladder's floor re-probe proved it.
- **advertised < 3.29** — SVG is unavailable **at any client setting**; route to PDF (API 2.8, vector with embedded fonts).
- **ceiling not established** — say so, and give the conditional rather than a confident instruction.

⚠️ Wording constraint: do **not** claim as measured that pinning above a server's ceiling breaks other calls. Tableau documents an unsupported-version error, but nobody here has measured it. State only what is provable — that it cannot enable SVG.

`scripts/tableau_render_capability.py` already models all of this carefully (`server_info`, the three-valued `supports`, `ApiVersions`, `API_RELEASE`, and `_add_pin_warnings`' refusal to infer support from an advertised number). The careful module and the careless one disagree; the fix is to make the careless path as careful as the ladder already is.

Guust-Franssens:
Fixes #474

A customer on an on-prem Tableau Server (`productVersion 2025.3.3` / `restApiVersion 3.27`) had their SVG legs refused, and the loudest, most actionable-looking line in the run told them:

> Set `TABLEAU_REST_API_VERSION=3.29` in `.env` and re-run

A client preference cannot lift a server's ceiling. **That is arithmetic, not judgement** — the advice was false for them, and the function that printed it already had the advertised ceiling in its own arguments and never looked at it.

---

## The three-state contract, and which branch each message implements

Every place we now say why SVG failed resolves through **one** classifier — `tableau_render_capability.svg_gate_advice` — to exactly one of three states. They are the three values of `supports(advertised, "3.29")`, so the partition is total **by construction** rather than by a chain of `if`s somebody has to keep exhaustive.

| `cause` | condition | what it says |
|---|---|---|
| `server_meets_floor` | advertised **≥ 3.29** | raise `TABLEAU_REST_API_VERSION`. Where the ladder's floor re-probe **proved** the tier, it says so; otherwise it says the advertised number is a *claim*, not proof |
| `server_below_floor` | advertised **< 3.29** (the customer) | SVG is unavailable **at any client setting**; raising the pin above the ceiling is **not** a fix. Routes to PDF (API 2.8, vector, embedded fonts) |
| `ceiling_not_established` | no `/serverinfo` answer | the **conditional**, never a confident instruction, plus how to establish the ceiling |

Three surfaces, all fed by that one function so they cannot disagree:

| surface | before | after |
|---|---|---|
| `tableau_oracle_manifest._log_blocked_and_stale` (console) | always the `.env` remedy | the branch above, prefixed with its `cause` |
| `capture_tableau_oracle._capture_render` (per-view `remedy` in `oracle-manifest.json`) | a flat constant | `ceiling_not_established` — honest, because the ceiling is a property of the **site** and is not in scope at that call — then upgraded to A or B by `_stamp_svg_gate` before serialisation |
| `assess_estate` (`report.md`, `assessment.json`, console) | never asked | the three numbers reconciled, with *"Best rung expected: PDF"* |

### Two things this deliberately does NOT claim

- **That pinning above a server's ceiling breaks other calls.** Plausible — Tableau documents an unsupported-REST-version error — and **unmeasured here**, so it is not stated at all, not even hedged. The only assertion is the provable one: it cannot enable SVG. `test_nothing_claims_that_over_pinning_breaks_other_calls` pins that.
- **That SVG *works* on a server whose advertised number clears the floor.** Only `proved_by_reprobe` licenses that, matching the discipline `_add_pin_warnings` already keeps.

The surviving half of the old sentence is kept but rescoped: *"PNG and PDF do not depend on the SVG floor at all — they reach back to API 2.5 and 2.8 respectively"* is a statement about **those routes' floors**, explicitly not a prediction about what changing the pin would do. It can no longer be read as reassurance about a state the reader reaches by following the advice above it.

---

## One status or two? (kept one; the `cause` carries the distinction)

I kept `unsupported_api_version` as a single status and added a sibling `cause` field.

**Why.** At the moment the status is stamped (`_capture_render`) the distinction is *not knowable* — the advertised ceiling is a run-level fact, and in state C it is unknowable at all. A split status would therefore need a **third** value meaning "gated, but we cannot say by what", which is a `cause` field wearing a status field's clothes. Keeping one status also leaves every existing consumer correct: `_partition`'s failed/blocked sets, `_VIEW_HEALTH_FAILURES` (which deliberately excludes it so salvage legs continue), `render_unestablished`, the exit-code logic in `write_manifest`, the mutation harness at `tests/mutation_capture_legs.py:146`, and the `_A_STATUS_LITERAL` certification in the redaction gate.

## Deviation from the brief, stated plainly

The brief calls state A *"client pin is the cause"*. I named it **`server_meets_floor`** — for the **evidence**, not a cause — because when the pin **also** clears the floor, "raise the pin" is itself a remedy that cannot work: the same defect one case over. That sub-branch says so instead (`raising the pin further will not help`) and is gated with its own negative control. Still three states; the label just stops asserting something the measurement does not support.

---

## The three questions

**Can the test reach the production code path?** Yes, and deliberately at three depths.
- `tests/test_svg_server_ceiling.py::test_a_plain_svg_run_probes_serverinfo_so_the_cause_is_established` drives the **real `oracle.main()`** — real `capture_view`, real `write_manifest`, real manifest written to disk — with only the transport and view-listing doubled. It kills three separate mutations: dropping the probe, probing when no SVG was asked for, and dropping `server_info` from the `write_manifest` call.
- `tests/test_assess_estate_resilience.py::test_the_render_ceiling_reaches_both_report_md_and_assessment_json` drives the **real `assess_estate.main()`** over the existing loopback-shaped `FakeTableau`. It exists **because a unit-only version of it did not kill anything**: `collect()` never probing, and `render_report` dropping the section, both SURVIVED the first mutation run. That is the honest record of this PR's one blind spot and how it was closed.
- The rest are unit tests over `svg_gate_advice` / `svg_gate` / `server_ceiling`.

**Does the guard cover the whole surface its name claims?** The surface is *"every place we tell a human why SVG failed"*. Enumerated: the console warning, the per-view `remedy` in the manifest, the capture-time record, the CLI `--svg` help text, the assessment report, `assessment.json`, the assessment console line — plus `docs/reference-capture.md` and `scripts/README.md`, which both carried the now-false claim in prose. `test_the_console_and_the_manifest_carry_the_SAME_remedy` asserts the two machine-readable surfaces are byte-identical rather than merely similar.

Named residuals, in scope but not closed:
- `capture_tableau_reference.py`'s `server_rest` provider is still a `NotImplementedError` stub (#194) and says nothing about versions, so there was nothing to correct there.
- The probe is skipped when no SVG is requested. Free is not weightless, and no other rung has a floor a customer can miss; `test_a_run_that_asks_for_no_svg_does_not_pay_for_the_probe` pins that as intended, not accidental.

**Does the evidence prove what the verdict claims?** The verdict is "each of these three states is reachable, distinguishable, and cannot silently collapse into another". Evidence: **20 source-level mutations, 20 caught**, each mutation a literal edit to a real script, each kill named by the specific test. Critically, the harness **reported 2 SURVIVED on its first run** — which is what demonstrates it can report a hole, rather than scoring an import error as a kill (the failure mode this repo has measured). Every assertion names the `cause` literal *and* a guard-specific substring, because a broad "it refused" assertion is satisfied by whichever guard runs first; and each state has a negative control that must **not** produce it.

---

## Redaction

`/serverinfo` is unauthenticated, so the version strings cannot reflect one of our credentials — but "it cannot leak *here*" is an argument that has to be re-made every time a call site moves, and this repo has watched that argument stop being true four times. So both version strings go through `tableau_env.redacted_note` **inside** `svg_gate_advice`, on the untransformed value, before any f-string touches them; `svg_gate_advice`'s `gate` parameter is a declared `TAINT_SEEDS` entry so the module boundary re-seeds it; and `test_a_reflected_credential_arriving_as_a_SERVER_VERSION_never_reaches_the_svg_advice` plants a live-token-shaped value **as** a server version and asserts nothing survives.

## Fixture fidelity (worth flagging)

`tests/test_assess_estate_resilience.py`'s `_Response` double lacked the `headers` attribute a real `HTTPResponse` has. The assessment's probe goes through `tableau_http._request`, which reads it — so without the fix the probe failed with `AttributeError` and **the run still passed**, because a broken probe and a fail-soft probe are indistinguishable from the outside. `FakeTableau` now serves an unauthenticated `/serverinfo` (XML, like the real endpoint) and takes `serverinfo=None` for a site that refuses it.

---

## Gates (by exit code, never by printed text)

| gate | result |
|---|---|
| `ruff format --check conftest.py scripts tests .github/skills` | exit 0 |
| `ruff check conftest.py scripts tests .github/skills` | exit 0 |
| `pylint scripts` | exit 0, 10.00/10 |
| `pylint conftest.py` | exit 0 |
| `python scripts/check_navigation_index.py` | exit 0 |
| affected suites (10 modules, 784 tests) | exit 0 |
| mutation harness, 20 mutations | 20 caught |

Two `too-many-locals` regressions this introduced were fixed rather than waived: `main()` gained a helper (`_advertised_ceiling`), and `write_manifest` carries a one-line targeted disable naming the 16th local and why it is bound once.

**Full suite: 4690 passed, 8 failed — all 8 pre-existing on `origin/master`**, verified by running them in a detached `origin/master` worktree: 6 in `test_upstream_repro_pins.py` / `test_issue_424_chart_type_pin.py` (the installed engine has moved 2.260.0 → 2.353.0, which those pins exist to detect) and 2 in `test_gui_marker_gate.py`, which reproduce at `origin/master` under `-n auto --dist loadfile` and concern `.github/skills/pbip-model-refresh/tests/`. I did not touch any of them.

---

## What I could NOT verify

- **Anything against the customer's server.** No access. Their numbers (`2025.3.3` / `3.27`) are used as fixture constants exactly as they reported them; every claim about *their* run is a claim about what our code now prints given those numbers, not about their site.
- **What a 3.29 request actually does to a 3.27 server.** Unmeasured, which is precisely why nothing in the code or docs says it. Only the arithmetic is asserted.
- **That a real on-prem Server below 2026.2 refuses SVG the way the fixture says.** The refusal text is the one already committed in `test_capture_tableau_oracle_svg.py`, measured on Cloud at 3.21/3.24/3.28. The on-prem version floors remain **inferred from Tableau's documentation** — `docs/reference-capture.md` already labels that table as inferred, and this PR does not upgrade it.
- **That `/serverinfo` answers unauthenticated on every on-prem topology.** Measured on Cloud (200 at 2.4/3.15/3.21/3.29); a reverse proxy in front of an on-prem Server could refuse it. That is exactly why state C exists and why the assessment fails soft, and `FakeTableau(serverinfo=None)` covers it — but the shape is a hypothesis, not a measurement.
- **The `svg_gate_advice` `_meets_floor_remedy` sub-branch in the wild.** A server that clears the floor while still refusing an at-or-above-floor request is an anomaly I have never observed; the branch exists because the alternative is emitting a knob that is already turned.


Guust-Franssens:
## Blind review (issue + diff only, no author reasoning) — 3 findings, all Medium

Reviewed tip `3ab3874`. Ruff exit 0; requested suite **111 passed** exit 0; other changed test modules **421 passed** exit 0. Targeted in-memory mutations were executed and killed for the three classifier states, manifest-only stamping, plain-SVG probing, no-SVG probe suppression, bare-serverinfo gate assembly, and the assessment JSON/report/console/unknown-state wiring.

### 1. An exact-floor server is still advised above its ceiling (`tableau_render_capability.py:437`)

A server advertising exactly REST 3.29 is told to set the client to **"3.29 or later"**. Every "later" value exceeds that server''s advertised ceiling, so the remedy still contains the impossible configuration this PR exists to remove. Reproduced with `advertised="3.29"`, `configured="3.21"`:

> set it to 3.29 or later in .env and re-run

The text reaches both the persisted remedy and the console via `tableau_oracle_manifest.py:410` and `:484`.

⚠️ **The covering test is vacuous, which matters as much as the defect.** Mutating only the exact-floor arm to "set it to 3.29." **executed** (`HITS=1`) and **both** the exact-floor and above-floor tests still passed — the exact-floor test asserts only on the *cause*, and a separate test explicitly expects the unbounded wording.

### 2. Malformed and unsuccessful `/serverinfo` responses become confident verdicts (`:289`, `assess_estate.py:1229`)

`api_tuple()` extracts arbitrary digit sequences rather than validating an API-version grammar, and `server_info()` parses `<restApiVersion>` **regardless of HTTP status**. `server_ceiling()` then treats any nonempty value as established:

| input | result |
|---|---|
| `not-a-version` | `(0,)`, `server_below_floor`, `established=True`, every rung unavailable |
| `garbage-999` | `(999,)`, `server_meets_floor`, `established=True`, best rung **SVG** |
| `3.x` | `(3,)`, `established=True`, best rung PDF |
| HTTP **500** body with `<restApiVersion>3.30</restApiVersion>` | parsed as REST 3.30 |
| HTTP **404** body with `<restApiVersion>3.27</restApiVersion>` | parsed as REST 3.27 |

Tests cover missing and empty versions, but not nonempty-malformed or non-200-carrying-the-element.

Note the reviewer explicitly declined to call a numeric-but-unpublished version (`9.99`) defective: `API_RELEASE` is a release-name lookup, **not** the capability boundary. Requiring membership in it would break real future servers.

### 3. The required pylint gate fails (`assess_estate.py:1246`)

Independently reproduced — **exit 12** while printing `9.98/10`:

```
assess_estate.py:1246:18: W0621: Redefining name 'tier' from outer scope (line 569)
tableau_render_capability.py:224:0: R0902: Too many instance attributes (8/7)
```

`R0902` comes from the new `RenderTier.ceiling` attribute; `W0621` from the new `_rung_verdict()`. My own fault for not catching this — I ran ruff and pytest on the recovered work before pushing, but not pylint.

### Not verified

Real Tableau Server behaviour when configured above its advertised ceiling (no live server); customer-specific behaviour against the reported server; the full repository suite (requested suites plus other changed modules only).

---

Routed to a fresh agent. Finding 1 must fix the remedy **and** make the test able to observe the remedy text; finding 2 must classify unknown rather than confident, without regressing future numeric versions.

Guust-Franssens:
## Designed mutation run for the #475 review fixes

The blind reviewer's finding was as much about a **vacuous test** as about the defect: the mutation
executed (`HITS=1`) and both floor tests stayed green. So each new behaviour here was broken
deliberately, with the mutated path instrumented to count executions — a non-zero pytest exit alone
is not a kill, since a mutation that never runs exits non-zero too.

Tests run per mutation: `tests/test_svg_server_ceiling.py`, `tests/test_assess_estate_resilience.py`
(+ the redaction suite where relevant). Sources restored from original bytes; tree verified clean
afterwards.

| # | behaviour broken | hits | verdict | first tests red |
|---|---|---|---|---|
| M1 | `api_tuple` refuses non-versions (reverted to the old arbitrary-digit-run extraction) | **41** | CAUGHT | `test_api_tuple_refuses_anything_that_is_not_an_api_version[*]`, `…_is_UNKNOWN_in_the_assessment_too[*]`, end-to-end `…NON_VERSION_leaves_the_ceiling_unknown…` (28 total) |
| M2 | `server_info` trusts a version only from a **200** (status gate removed) | **5** | CAUGHT | `test_a_version_carried_by_an_UNSUCCESSFUL_serverinfo_is_not_a_ceiling[500-3.30]` / `[404-3.27]`, `…_ERROR_PAGE_never_becomes_the_sites_ceiling_end_to_end` |
| M3 | `supports()` returns the **unknown** third state (→ `False`, a confident "below floor") | **63** | CAUGHT | every unknown-state test in both files (27 total) |
| M4 | `server_ceiling` re-validates **independently** of `server_info` (defence in depth) | **13** | CAUGHT | `test_a_ceiling_that_is_not_a_version_is_UNKNOWN_in_the_assessment_too[*]` — and *only* those three, which is the point: it is the one test that stubs `server_info`, so it is the only one with power over this layer |
| M5 | the offending text is kept as `invalid_rest_api_version` (the probe diagnostic) | **3** | CAUGHT | `…yields_no_ceiling_but_keeps_the_evidence`, `…goes_through_the_redaction_chokepoint`, end-to-end |
| M6 | that diagnostic goes through the redaction chokepoint (raw value instead) | **2** | CAUGHT | `test_the_offending_version_text_goes_through_the_redaction_chokepoint` |
| M7 | finding 1: the remedy names the floor **exactly** (reintroduced "3.29 or later") | n/a¹ | CAUGHT | `test_an_exact_floor_server_is_NEVER_advised_above_its_own_advertised_ceiling`, `test_no_remedy_names_a_REST_VERSION_the_advertised_ceiling_cannot_serve[3.29\|3.30\|9.99]`, `test_exactly_at_the_floor_counts_as_meeting_it` (7 total) |

¹ A string-literal mutation has no branch to instrument. Execution is proved instead by the seven red
tests, which fail *on the mutated text itself* — including the invariant test, which parses the
version tokens out of the remedy.

**7/7 caught.** M7 is the reviewer's own experiment run in the opposite direction: reintroducing the
unbounded wording now turns the exact-floor test red, which it could not do before, because that test
asserted only the `cause`.

### The harness can report a hole

A perfect first result is a reason for suspicion, not celebration — this repo has already had a
mutation harness score 22/22 by counting an import error as a kill. Two negative controls:

| # | mutation | result |
|---|---|---|
| C1 | `CLI_FETCH_TIMEOUT_SEC = 120 → 121` (a constant these suites deliberately do not cover) | **SURVIVED**, exit 0, hits 0 |
| C2 | a word in a docstring | **SURVIVED**, exit 0, hits 0 |

So the 7/7 above is a measurement, not a mechanism that scores everything green.

### Gates (by exit code, never the printed score)

| gate | exit |
|---|---|
| `ruff format --check` + `ruff check` (5 changed files) | **0** |
| `pylint scripts/assess_estate.py scripts/tableau_render_capability.py scripts/tableau_oracle_manifest.py` | **0** (10.00/10; was **12** at 9.98/10) |
| `pytest test_svg_server_ceiling.py test_assess_estate_resilience.py test_diagnostic_redaction.py -q` | **0**, 538 passed |

Full suite measured once at `9b43069`'s tree: **7 failed / 4486 passed** — the six pre-existing
engine pins (`test_issue_424_chart_type_pin.py`, `test_upstream_repro_pins.py`) plus one caused by
this change and fixed in `a9f7c7b`; expected state is now exactly six.

Guust-Franssens:
## Round-2 High finding closed: `/serverinfo` product and build fields

**Reproduced independently before fixing** — production-shaped 200 with a synthetic session token in
**both** `productVersion` and its `build` attribute, redactor supplied:

| case | established | json_leak | report_leak | console_leak |
|---|---|---|---|---|
| `PRODUCT_BUILD_TOKEN_200` | True | **True** | **True** | **True** |
| `REST_TOKEN_200` (control) | False | False | False | False |
| `TOKEN_NON200` (control) | False | False | False | False |

After the fix, `PRODUCT_BUILD_TOKEN_200` reads `established=True  json_leak=False  report_leak=False
console_leak=False` — **established stays True**: redaction is not refusal, so the ceiling is still
established and the rung table still prints. The two controls confirm this was a separate,
pre-existing route: the `restApiVersion` and non-200 paths closed earlier in this PR stayed closed.

Fixed **at the parse boundary** inside `server_info()`, not at the three consumer sites — a
consumer-side fix protects the consumers that exist today and no others, which is how this route
stayed open while the neighbouring one was closed twice. `None` is preserved rather than collapsing
onto `redacted_note`'s `""`, because "the element was absent" and "the server sent an empty string"
are different facts.

### The certification rationale, and why the category is deleted

`_SERVERINFO` certified these fields via **`UNAUTHENTICATED-SOURCE:`** — *"that request carries no
credential, so the response cannot reflect one"*. That is an argument about the **request**; the
hazard is the **speaker**. The same server, or any intermediary, observed the PAT sign-in earlier in
the same run.

The justification is now `REDACTED-UPSTREAM:` (free-form fields) plus `SHAPE-VERIFIED:` (a *valid*
`restApiVersion`, proved numeric by `api_tuple` — a property of the value, never of the request), and
**`UNAUTHENTICATED-SOURCE:` is removed from the `CATEGORIES` vocabulary** rather than left unused, so
the same unsound reasoning cannot be reached for again. It had exactly one holder: this leak.

### Designed mutation run

| # | behaviour broken | hits | verdict | red |
|---|---|---|---|---|
| S1 | `product_version` redacted at the boundary (raw returned) | **19** | CAUGHT | end-to-end `…SERVERINFO_VERSION_FIELDS_reaches_no_surface`; unit `…redacted_at_the_parse_boundary` |
| S2 | `build` redacted at the boundary (raw returned) | **19** | CAUGHT | same two, failing on the *other* field |
| S3 | both reverted — i.e. the exact pre-fix code | **19** | CAUGHT | same two |
| S4 | an **absent** field stays `None` (returns `""`) | **38** | CAUGHT | `…stays_None_rather_than_becoming_an_empty_string` |
| C1 | *control:* `CLI_FETCH_TIMEOUT_SEC 120→121` | 0 | **SURVIVED** | — |
| C2 | *control:* redaction kept, output bound widened 40→4000 | 0 | **SURVIVED** | — |

S1 and S2 each move **exactly one** side of the two-field assertion and are each caught on their own
field. C2 is a deliberate, honest survivor: the output bound is not the security boundary — redaction
has already run before truncation — so nothing pins the number, and no test pretends to.

### Tests added

Three surfaces, driven through the **real `main()`**, planting the run's **own configured PAT
secret** rather than a synthetic stand-in, so the test is about a reachable route:
`assessment.json`, `report.md`, and the assessment console. Each asserts `[REDACTED]` is **present**
as well as the secret absent — a fix that made the fields *vanish* cannot pass — plus a negative
control that an ordinary `2025.3.3` / `20253.25.0904.1234` survives byte for byte.

The pre-existing security test covered only `svg_gate_advice()`, which does its own later redaction
and was therefore structurally incapable of observing this leak.

### Gates (exit codes)

| gate | exit |
|---|---|
| `ruff format` + `ruff check --fix` | **0** |
| `pylint scripts/assess_estate.py scripts/tableau_render_capability.py scripts/tableau_oracle_manifest.py` | **0** (10.00/10) |
| `pytest test_svg_server_ceiling.py test_assess_estate_resilience.py test_diagnostic_redaction.py -q` | **0**, 542 passed |
| `pytest` on the adjacent consumers of these fields (`test_capture_tableau_oracle{,_svg}.py`, `test_tableau_render_capability.py`, `test_stamp_tableau_provenance.py`, `test_tableau_env.py`) | **0**, 309 passed |
| docs gates (`test_repo_layout`, `test_no_conflict_markers`, `test_agent_capabilities`) | **0**, 58 passed |

Full suite not re-run this round, per instruction; expected state remains the six pre-existing engine
pins.

Guust-Franssens:
## Security confirmation (round 3, narrow) — verdict: **NO, one route remains. Blocks.**

Reviewed tip `ffec48eadde865d24e2ec83ff938513fe0542336`. Gates re-verified: ruff **0**, pylint **0**, targeted pytest **542 passed, 0**.

### Confirmed CLOSED by the last commit
- Free-form `productVersion` and `build` are protected **at the parse boundary**, so a new downstream consumer is protected without doing anything itself.
- Invalid token-shaped `restApiVersion` is refused.
- Non-200 bodies stay safe.
- **Truncation ordering verified**: a long token crossing the 40-character bound was redacted *first* — the bounded result held `[REDA` and no raw token. C2''s survival is therefore honest.
- **S3 spot-check**: reverting to the exact pre-fix code executed across **19** `/serverinfo` calls and failed the expected two security tests.
- Deleting `UNAUTHENTICATED-SOURCE:` from the vocabulary is **agreed correct** — whether the current request carries credentials says nothing about what the server or an intermediary previously observed.

### Still open — numeric-shaped credentials bypass redaction (`tableau_render_capability.py:583`)

```
NUMERIC_REST_TOKEN_200   info_leak=True json_leak=True report_leak=True console_leak=True
                         established=True  advertised=3.27
```

The code returns `rest_api_version` untransformed, and says why:

> *"only because it is SHAPE-VERIFIED: ... a numeric API-version grammar, **which no credential can satisfy**."*

**That is unenforced.** `assess_estate.py:309` assigns `self.token = creds["token"]` with no shape validation at all, so a server or intermediary can issue a token that is literally `3.27` and reflect it back as `restApiVersion`. It satisfies the grammar, takes the untransformed branch, and reaches `assessment.json`, `report.md` and the console. **The supplied redactor knows the value is a credential; this branch deliberately bypasses it.**

`established=True` is behaviourally correct, but it is currently kept true **by carrying the raw credential downstream** as `advertised_api_version`.

### Why this is the same defect as the one just fixed

`UNAUTHENTICATED-SOURCE:` was deleted because "the request was unauthenticated" is an unenforced claim about safety. **`SHAPE-VERIFIED` is that same argument one layer down** — an unenforced claim that credential shapes and version shapes are disjoint. A future exemption of this kind must rest on a genuinely disjoint, *enforced* shape.

Consequently the fix must **not** be a tighter grammar: narrowing the overlap leaves the assumption intact and the next token shape walks through it.

### Fix being applied

Compare the raw value internally to derive the ceiling, rung verdicts and `established`; carry a **redacted display value** as the only thing serialized, rendered or logged. Derived capability results travel separately from displayable text — so the rung table still prints and the credential never leaves the parse boundary.

Plus: correct the `SHAPE-VERIFIED` comment so it no longer asserts something untrue, and pin `NUMERIC_REST_TOKEN_200` as a permanent test on all three surfaces.

Guust-Franssens:
## Round-3 High closed: a numeric-shaped credential is no longer published as the ceiling

**Confirmed independently before fixing** (the console column read `False` on my first run because I
omitted `logging.basicConfig(level=INFO)` and the line is emitted at INFO — a measurement error in my
harness, corrected before drawing any conclusion):

| case | info | json | report | console | established | advertised |
|---|---|---|---|---|---|---|
| `PRODUCT_BUILD_TOKEN_200` | False | False | False | False | True | `3.27` |
| `REST_TOKEN_200` | False | False | False | False | False | `None` |
| `TOKEN_NON200` | False | False | False | False | False | `None` |
| **`NUMERIC_REST_TOKEN_200`** | **True** | **True** | **True** | **True** | True | **`3.27`** |

After: `NUMERIC_REST_TOKEN_200 → info=False json=False report=False console=False established=True
advertised='[REDACTED]'`, with `best=pdf` and rungs `(unavailable, available, available)` — the
capability is fully intact.

⚠️ Substring checks are worthless for `3.27` (it is a plausible substring of a floor, a release name
or a date), so every assertion names the **field** or the **rendered phrase** — `advertised_api_version`,
`"advertised_api_version": "3.27"`, ``REST `3.27` ``, `advertises REST 3.27`.

### Not fixed by tightening the grammar

Agreed with the reviewer, and stated in the code so nobody "improves" it back: a tighter grammar
narrows the overlap between credential shapes and version shapes while leaving the **assumption** in
place, and the next token shape walks through it. `SHAPE-VERIFIED` here was the same class of claim as
the `UNAUTHENTICATED-SOURCE` category retired last round — an assumption standing where an enforced
property was needed.

**Derived capability now travels separately from displayable text.** The raw value is compared inside
`server_info` and never leaves it. What leaves:

| field | what it is |
|---|---|
| `rest_api_version` | the **redacted display** string — byte-identical to the raw in every run where the redactor finds nothing |
| `rest_api_version_reflected` | the redactor rewrote it, i.e. the server quoted back a credential we hold |
| `ceiling_established`, `rung_support` | the **derived capability** from the raw value: one boolean per *published* floor |

The **release name is suppressed with the number** — `API_RELEASE` is a bijection, so printing
*"Tableau 2025.3"* hands back the digits redaction removed. `report.md` and the console both **say**
the number was suppressed and why; a bare marker where a version belongs reads as a bug in this tool,
and a server echoing a credential is worth escalating.

⚠️ **Residual, stated not hidden:** three booleans across documented floors narrow a suppressed
version to an *interval* (`svg: False, pdf: True` ⇒ `[2.8, 3.29)`). Unavoidable for any honest report —
the operator's question *is* which rungs the site reaches. It is recorded in `rung_support`'s
docstring and in its `DERIVED-IRREVERSIBLY` certification.

**The certification gate caught all three new dict-values on the first run** — exactly its job — and
each is certified with what is actually enforced: `NOT-A-STRING` (the boolean), `REDACTED-UPSTREAM`
(the display string), `DERIVED-IRREVERSIBLY` (`rung_support`, with the residual).

### Designed mutation run

| # | behaviour broken | hits | verdict | red |
|---|---|---|---|---|
| N1 | version leaves `server_info` redacted (republish the raw — the round-2 code) | **24** | CAUGHT | both `NUMERIC…` tests, `…never_carries_the_version_STRING`, **and both certification gates** |
| N2 | `reflected` reports what the redactor did (hardwired `False`) | **24** | CAUGHT | both `NUMERIC…` tests |
| N3 | `rung_support` derived from the RAW ceiling (derive from display text instead) | **24** | CAUGHT | both `NUMERIC…`, `…never_carries…`, stale-certification gate |
| N4 | the assessment consumes the derived capability (re-derive from the display string) | **48** | CAUGHT | end-to-end + `…suppresses_the_RELEASE_NAME…` |
| N5 | the release name is suppressed with the number | **48** | CAUGHT | end-to-end + `…suppresses_the_RELEASE_NAME…` |
| N6 | the report **says** the number was suppressed | **21**¹ | CAUGHT | end-to-end + `…suppresses_the_RELEASE_NAME…` |
| C1 | *control:* `CLI_FETCH_TIMEOUT_SEC 120→121` | 0 | **SURVIVED** | — |
| C2 | *control:* `SERVERINFO_TIMEOUT_SEC 30→31` | 0 | **SURVIVED** | — |

¹ N6 first reported `hits=0` because I put the marker in a **comment** on a dead-branch disable — it
measured nothing. Re-run with the marker in executable position: **21 hits**, same two tests red.
Reporting this rather than the convenient first number, since a hit count that cannot have executed is
exactly the false kill this harness exists to avoid.

### Gates (exit codes)

`ruff check` **0** · `pylint` (3 modules) **0**, 10.00/10 · `pytest test_svg_server_ceiling.py
test_assess_estate_resilience.py test_diagnostic_redaction.py -q` **0**, 550 passed · adjacent
consumers (oracle ×2, render capability, provenance, env) **0**, 309 passed · docs gates **0**, 58
passed. Full suite not re-run, per instruction.

### One observation, not a change

`SHAPE-VERIFIED` survives for the view **LUID**, anchored to a full UUID regex by `artifact_stem`.
That exemption is *enforced* (the regex runs), which is the property this round established as the
bar — but it still assumes a session token cannot be UUID-shaped, and nothing validates a token's
shape. I have **not** touched it: no measurement, and it is outside this PR. Flagging it as a
candidate for the same scrutiny rather than expanding scope silently.

PR #477, opened 2026-09-03
fix(work_dirs): detect a renumbered _runs/ directory, and state the prohibition where an agent reads it

### LINKED ISSUE #470: A renumbered _runs/ directory is undetectable, and work_dirs.py never says not to
Customer feedback from the SES engagement. Their agent renumbered **14 `_runs/` directories** in one reorg. Per this repo's own convention that is destructive - and right now it is also **undetectable**.

### 1. The prohibition is absent from the artifact an agent actually reads

The rule is documented, twice:

- `docs/migration-phases.md:57-61` - *"**The number is the identity** - never renamed, never reused, because generated bundle output embeds absolute self-paths."*
- `docs/operator-runbook.md:1252-1255` - same rule.

But `scripts/work_dirs.py` - the module their agent read - never states it. Its module docstring and `allocate_run`'s docstring say only *"the number is the identity"* and *"the slug is decoration only (never parsed back)"*. Both are statements about **naming semantics**. Neither says renaming an existing run is destructive, nor why.

Their agent, in its own words: *"I ran the whole per-workbook `_runs/` reorg (14 dirs renumbered) based on pattern-matching `work_dirs.py`'s docstring, without first checking `docs/operator-runbook.md` / `docs/migration-phases.md`."*

A rule only a separate document carries is a rule the hand on the tool does not see.

### 2. A renumbered run is silently indistinguishable from an intact one

`allocate_run` writes `run`, `unit_key`, `created`, `status` into `run.json`. It does **not** record the path it allocated.

`list_runs` then does `data.setdefault("root", str(child))` - it derives `root` from wherever the directory is found *now*. So a run that has been renumbered, renamed or moved reads back as perfectly healthy, and every status query over it silently reports the new location as if it were the original.

This is the repo's recurring defect class: unassessable state collapsing into the clean bucket. There is no signal to collapse *from* - the manifest holds no self-path to compare against.

## Proposed fix

**Detection (the substantive half).** Record the allocated path in `run.json` at allocation time, and surface a mismatch between recorded and actual location as its own distinct state - not clean, not silently backfilled.

Three states, and the third is mandatory:
- recorded path == actual location -> intact
- recorded path != actual location -> **moved/renumbered**, reported as such
- no recorded path (every run allocated before this change, e.g. the existing `_runs/408-*`) -> **unverifiable**, never "intact"

Backwards compatibility is the crux: existing manifests have no such key, and they must not be reported as either corrupt or clean.

**Prevention (cheap).** Put the real prohibition, with its reason, in `work_dirs.py`'s own docstrings, and point at `docs/INDEX.md` plus the two documents that define run granularity. Their agent also inferred one run *per workbook*; `docs/migration-phases.md:57` says one *per pipeline run*.

## Also in scope - three Windows/PowerShell misreads that cost them real time

Their agent self-reported these, all avoidable:

- PowerShell surfaces normal stderr progress (`git clone`, any Python `print`-to-stderr) as `NativeCommandError`, so a **succeeding** command looks like it failed. They double-checked the filesystem to be sure the engine clone had worked.
- Long Windows paths broke plain `Remove-Item`. `docs/windows-path-limits.md:290-292` already documents the fix (`\\?\` prefix plus a chmod handler for read-only git objects) - they found it by trial and error instead.
- `Start-Process -ArgumentList` silently mangled a multi-word `--workbook` name, wasting an oracle-capture launch, until they switched to a properly quoted script file.
- An SSL probe using `requests`+certifi failed against the live server while the toolkit's own scripts succeeded, because those use `urllib.request` and the Windows certificate store. They had to reverse-engineer that from our source.

Make these discoverable from `docs/INDEX.md`. Prefer extending an existing document over adding new ones.

Guust-Franssens:
Fixes #470

A customer's agent renumbered **14 `_runs/` directories** in one reorg. Two things failed at once: the prohibition was absent from the module the agent was reading, and afterwards there was **no way to tell**. This fixes the second properly and the first cheaply.

---

## Two premise corrections, up front

The issue's evidence is accurate about the defect, and slightly off in two places. Both are load-bearing, so they are stated rather than quietly worked around.

**1. `work_dirs.py` did already carry the rule — in the one docstring nobody reads.** The issue says the module never states it. `RunPaths` (line 155 on master) says: *"Never rename `root` after allocation (issue #234, rule 1): generated bundle output embeds absolute self-paths, so moving a run after the fact breaks refresh."* That does not weaken the issue — it sharpens it. The rule was in a **dataclass** docstring, while an agent reading "how do I make/name a run" reads the **module header** and **`allocate_run`**, which said only *"the number is the identity"* and *"the slug is decoration only"*. Placement, not absence, was the defect. So the prohibition now appears in all three, and a test asserts it stays in the two that were missing it.

**2. `run` + `unit_key` already imply the directory name — legacy runs are not information-free.** `allocate_run` names the directory `f"{run:0{max(3,len(str(run)))}d}-{unit_key}"`, and both fields are already in every manifest. A pre-change run *can* therefore be compared against its own directory name. This matters because it is exactly the trap the issue warns about: a derived answer that looks like a recorded one.

The resolution keeps both properties. `unverifiable` is a hard state that inference can never promote, **and** the derived comparison is still reported — clearly subordinate, on `unverifiable` runs only, labelled `INFERENCE` in its own message. Discarding it would have thrown away the only thing that can flag an *already*-renumbered run; letting it decide the state would have rebuilt the defect. Verified by mutation: promoting a matching derived name to `intact` turns `test_a_legacy_manifest_with_no_recorded_name_is_unverifiable_never_intact` red.

---

## The contract

| state | condition | surfaces as |
|---|---|---|
| `intact` | recorded name == actual directory name | `list_runs()[i]["location_check"]["state"]`; `--verify` prints `INTACT`; **exit 0** |
| `moved` | recorded name != actual name | same, `MOVED`, **naming both directories**; **exit 1** |
| `unverifiable` | no usable recorded name | same, `UNVERIFIABLE` + an explicit *"'unverifiable' is NOT 'intact'"* line; **exit 3** |

Exit codes deliberately mirror `check_reference_readiness.py` — 0 clean / 1 findings / 3 `CANNOT_ESTABLISH`, **and neither 1 nor 3 is a pass**. `moved` outranks `unverifiable` because an established finding beats an unanswered question; a caller gating on `-eq 0` catches both either way.

Three surfaces, one verdict function:

- `check_run_location(manifest, run_dir)` — pure, no disk, no mutation, never raises.
- `list_runs()` — attaches `location_check` to every run. Still *"pure inspection, never mutates, never raises"*; it does not repair and does not raise.
- `python scripts/work_dirs.py --verify [--json]` — new CLI mode. Allocates nothing (asserted).

**No auto-repair.** Rewriting the recorded name to match the current location would erase the only evidence that anything moved.

**`allocate_run` is unchanged where it matters.** Still `mkdir`-exclusive with retry, never read-the-max-then-write; the existing true-race test still passes. The new key is written in the **same reserved block** as `run`/`unit_key`/`created`/`status`, i.e. after `extra_manifest` is merged, so a caller cannot plant a false self-path and forge an `intact` verdict.

## Why the directory NAME, and not a path

`allocated_dir_name` holds `"408-coldrun2-fabric-migration-lab"` — not an absolute path, not a path relative to the repo.

- **A renumber/rename changes it. Every legitimate relocation does not.** Cloning the repo, moving the checkout, or working in a `git worktree` all change the absolute path while the run keeps its identity. Recording an absolute path would report *every run in a fresh clone* as `moved` — and a check that fires on a legitimate clone is one people learn to ignore, which is strictly worse than no check. `test_a_relocated_repository_is_still_intact_not_moved` copies a whole repo root elsewhere and asserts `intact`, after first asserting the absolute paths really did change (otherwise the test proves nothing).
- **Relative-to-`_runs/` would be identical here.** `list_runs` scans `_runs/` at depth 1 only, so the relative path *is* the directory name for anything it can observe. Recording both would be redundant.
- **The honest limit:** copying a run into a *different* repo's `_runs/` under the same name reports `intact`. That is a real blind spot of name-based comparison, accepted deliberately — it is rare, and closing it costs the false positive on every clone. Recorded in "not verified" below rather than hidden.

## The three questions

**Can the test reach the production code path?** Yes, and for the CLI it is proven rather than assumed. `test_verify_cli_*` shell out to `sys.executable scripts/work_dirs.py --verify --repo-root <tmp>` in a subprocess, so argparse, `main()`, the printer and the exit code are all real. The state tests consume `list_runs()`, which is what a caller actually calls, not the helper underneath it. The docstring test reads `work_dirs.__doc__` / `allocate_run.__doc__` from the imported module, so deleting the warning from the source turns it red (mutation-verified).

**Does the guard cover the whole surface its name claims?** Scoped honestly: it claims *"is this run still in the directory it was allocated in"*, and that is all it checks. It does **not** validate the bundles underneath, does not detect a run moved out of `_runs/` entirely (`list_runs` cannot see it), and does not detect the cross-repo copy above. Every run `list_runs` returns gets a `location_check` — there is no path through it that returns a run without one (`test_list_runs_attaches_a_location_check_to_every_run`), and an unusable recorded value (int, list, dict, empty string) degrades to `unverifiable`, never to `intact`.

**Does the evidence prove what the verdict claims?** For `intact`/`moved`, yes — the verdict rests on a value written by `allocate_run` at allocation time and compared to the filesystem, with `derived_name_check` pinned to `not_consulted` so it is provable no inference contributed (asserted). For `unverifiable`, the verdict is *"cannot be established"*, and that is exactly what the evidence supports: the manifest has no record. The derived hint never upgrades it.

## Verification — all by exit code

| gate | result |
|---|---|
| `ruff format` + `ruff check --fix` on both changed `.py` | exit 0, "All checks passed!" |
| `pylint scripts` (the CI invocation) | **exit 0**, 10.00/10 |
| `pylint tests/test_work_dirs.py` | exit 16, 12 findings — **all 12 pre-existing**, measured against `git show origin/master:tests/test_work_dirs.py` (exit 16, 12 findings, 8.95/10 → now 9.57/10). Zero findings added; CI does not lint `tests/` |
| `pytest tests/test_work_dirs.py` | **exit 0**, 58 passed |
| `pytest tests/test_scripts_help.py` | **exit 0**, 211 passed (see the F003 incident below) |
| `pytest -q` (tier-2 full suite) | **6 failed, 4713 passed, 10 skipped** in 1184 s — all 6 failures reproduce on `origin/master` (see below) |
| `python scripts/check_navigation_index.py` | **exit 0** |
| `python scripts/check_agent_capabilities.py` | **exit 0** |
| `python scripts/work_dirs.py --help` under `PYTHONIOENCODING=cp1252` | **exit 0** |

**The full suite caught a real regression in this branch, and one self-inflicted false alarm.** Both are recorded because the first is exactly the kind of thing a green targeted run hides:

- ❌ **Real, mine, now fixed (commit 3):** I put a `⚠️` in the **module docstring**, which is argparse's `description`. Windows defaults stdout to cp1252, so `python scripts/work_dirs.py --help` exited 1 with `UnicodeEncodeError` before printing anything — caught by this repo's own F003 guard, `test_help_exits_zero_on_a_cp1252_stdout[work_dirs.py]`. The glyph now lives in a comment directly beneath the docstring, which also states the constraint so the next editor does not re-add one. Verified after: module-docstring non-ASCII characters `[]`, `--help` exit 0 under cp1252.
- ⚠️ **False alarm, mine, not a code defect:** `test_help_exits_zero_on_a_cp1252_stdout[check_navigation_index.py]` also failed — because my own untracked scratch `pr-body.md` was sitting at the repo root, and `check_navigation_index.py` treats every untracked-but-not-ignored `.md` as an indexable candidate. That test invokes the script with `--help`, which it ignores, so it really ran the check and really failed. Scratch moved into a gitignored `_pr/`; the gate is exit 0 and the test passes. Worth recording: **a scratch `.md` at the repo root breaks the navigation gate**, and the failure surfaces under an unrelated test's name.
- ✅ **Pre-existing, not mine:** the remaining 6 failures (`test_issue_424_chart_type_pin.py` ×3, `test_upstream_repro_pins.py` ×3) reproduce identically on **`origin/master`** in a separate checkout — they pin an engine version (`- 2.260.0 / + 2.353.0`) that has moved on. Not touched.

**The live-fixture proof (the backwards-compatibility claim, not a synthetic fixture):**

```
$ python scripts/work_dirs.py --verify --repo-root <this machine's repo>
_runs/ location check: ...\_runs
  UNVERIFIABLE  406-meridian-smoke-2-339-0-20260901
  UNVERIFIABLE  407-dryrun-gates
  UNVERIFIABLE  408-coldrun2-fabric-migration-lab
                no 'allocated_dir_name' recorded (allocated before this key existed); its
                run/unit_key imply '408-coldrun2-fabric-migration-lab', which matches - but that is INFERENCE
3 run(s): 0 intact, 0 moved, 3 unverifiable
  'unverifiable' is NOT 'intact' - those runs predate path recording and cannot be checked.
EXIT: 3
```

All three real runs report `unverifiable`, **not** `intact` — including `408-coldrun2-fabric-migration-lab`, whose derived name matches.

**Mutation testing, run before opening this PR** — 9 mutations, 9 killed. Because a perfect first result deserves suspicion, the harness was itself checked two ways: one mutation was re-run by hand and its failure inspected (`AssertionError: assert 'intact' == 'moved'` — a genuine assertion, not an import error scoring as a kill), and a **negative control** (rewording an unasserted `detail` string) correctly **survived**. Mutations killed: dropping the recorded key; upgrading a matching legacy run to `intact`; comparing absolute paths instead of names; dropping `moved` from the exit code; dropping the non-dict manifest guard; reverting `root` to `setdefault`; making the key clobberable via `extra_manifest`; deleting the prohibition from the docstring; reporting a moved run as `intact`.

## Two adjacent fixes, both tightly coupled

1. **`list_runs` could raise, contradicting its own docstring.** A `run.json` holding valid JSON that is not an object (`[]`, `"str"`, `42`, `null`) reached `data.setdefault(...)` → `AttributeError`, which sails past the `json.JSONDecodeError`/`OSError` guard. Pre-existing, and directly in the way: I add a second `.get()` on the same object. Now skipped; parametrized test over all four shapes plus malformed and empty payloads.
2. **`root` was `setdefault`, not assigned.** A `root` planted via `extra_manifest` shadowed the real location — a stale recorded value silently presented as current, which is the same failure mode the issue is about. `root` is derived, so it is now assigned unconditionally and documented as such.

## Documentation

- `scripts/work_dirs.py` — prohibition + reason + granularity in the **module docstring** and **`allocate_run`**, pointing at `docs/INDEX.md`, `docs/migration-phases.md` and `docs/operator-runbook.md`.
- `docs/migration-phases.md` Phase 1 — the granularity sentence was thin, and I can name it: line 57 read *"One numbered run directory per pipeline run"* with no counter-example, immediately above a table describing run 408 as a 48-workbook sweep. It now says what it excludes: run 408 is **one** run covering 48 workbooks, not 48 sibling runs, and per-workbook units live in its `bundle/pbip/`. The rest of the section was already correct and is untouched.
- `docs/operator-runbook.md` §6.4 — same rule plus the `--verify` command and the exit-3 caveat.
- `docs/operator-runbook.md` **§4.12 (new)** — the four Windows/PowerShell misreads, in the failure playbook where an operator already looks. No new file.
- `docs/windows-path-limits.md` — content was right, discoverability was not: the deletion recipe sat in a blockquote under *"Reproducing it"*. Promoted to its own heading naming the symptom (`Remove-Item` refuses it), stating why **both** halves are load-bearing.
- `docs/INDEX.md` — searchable keywords in the three affected descriptions.

**On the `requests`/certifi item, which the brief asked me to verify rather than repeat:** ✅ **`scripts/` contains no `requests` import at all** (`Select-String -Path scripts\*.py -Pattern "import requests|from requests"` → no matches). Every Tableau call goes through `urllib.request` (`tableau_http.py`, `capture_tableau_oracle.py`, `assess_estate.py`, `tableau_render_capability.py`, `deploy_estate.py`, `harvest_tableau_public.py`), and `tableau_http.py`'s `_DeadlineHTTPSHandler` forwards the context so *"certificate verification is unchanged"*. CPython's default `SSLContext` on Windows is populated from the Windows certificate store; measured on this machine: `len(ssl.create_default_context().get_ca_certs())` = **98**, `len(ssl.enum_certificates('ROOT'))` = **69**. `requests` verifies against certifi's static PEM instead, which is why an ad-hoc probe fails where the toolkit succeeds. Stated as the repo's mechanism, with the numbers.

## What I could NOT verify

- **The customer's 14-directory reorg.** Not reproduced; taken from the issue.
- **That bundle output actually embeds absolute self-paths.** The premise of the whole prohibition, asserted by `docs/migration-phases.md`, `docs/operator-runbook.md` and #234, and repeated here. I did not open a bundle and grep for absolute paths. If it is false, the prohibition is over-strict — the *detection* still behaves correctly, it would just be guarding less than claimed.
- **The cross-repo copy blind spot** (named above) is reasoned from the design, not exercised by a test — there is deliberately no code path that could distinguish it.
- **Whether `--verify` should be wired into `check_unit.py` or preflight.** It is opt-in; nothing calls it automatically. Deliberate: `_runs/` is machine-local scratch and a repo-wide exit 3 in preflight on every existing machine would be noise on day one. Worth revisiting once most runs carry the key.
- **The four Windows items** are prose, ungated by any test. ① and ④ were re-measured here; ② is this repo's own existing documented recipe; ③ is `Start-Process` semantics plus the measured consequence recorded in the issue — I did not re-run the mangled oracle capture.
- **Whether the module docstring must be ASCII on non-Windows** — the F003 guard simulates cp1252 via `PYTHONIOENCODING`; I did not test on a real Windows console with a different code page.

## Follow-ups (not done here, on purpose)

- **`AGENTS.md`** carries the same `_runs/` convention text and arguably deserves the `--verify` pointer. Not touched — it holds a synced block that regenerates four persona files, and editing it from this branch breaks a CI check I cannot see the far side of.
- **`scripts/README.md`** would normally get the new CLI mode; another branch owns that file right now.
- A `--prune`/migration helper that backfills `allocated_dir_name` is **deliberately not** offered: writing the key now would record where a run is *today*, not where it was allocated, i.e. manufacture an `intact` verdict with no evidence behind it.

## Files

`scripts/work_dirs.py`, `tests/test_work_dirs.py`, `docs/INDEX.md`, `docs/migration-phases.md`, `docs/operator-runbook.md`, `docs/windows-path-limits.md`. Nothing outside the assigned scope; no file owned by an in-flight branch.

### DIFF
--- a/docs/INDEX.md
+++ b/docs/INDEX.md
@@ -72,10 +72,10 @@ Eligible files: tracked Markdown knowledge/navigation files, `.github/pbi.kb/**/
 | Find migration guidance | [`docs/deterministic-tier-integration.md`](../docs/deterministic-tier-integration.md) | Engine and handoff integration. |
 | Find migration guidance | [`docs/dry-run-findings-2026-08-11.md`](../docs/dry-run-findings-2026-08-11.md) | Durable F00x dry-run findings. |
 | Find migration guidance | [`docs/engine-gap-harvest.md`](../docs/engine-gap-harvest.md) | Attribute and classify the `reports/` vs `pbip/` delta into engine-gap evidence — provenance delegated to the `--tamper` gate's own adjudicator. |
-| Find migration guidance | [`docs/migration-phases.md`](../docs/migration-phases.md) | The three-phase pipeline shape: what each phase produces, where it lands, which gate guards each hop, and the known gaps. |
+| Find migration guidance | [`docs/migration-phases.md`](../docs/migration-phases.md) | The three-phase pipeline shape: what each phase produces, where it lands, which gate guards each hop, the `_runs/<NNN>-<slug>` identity rule (never renamed/renumbered; one run per pipeline run, **not** per workbook), and the known gaps. |
 | Find migration guidance | [`docs/migration-programme.md`](../docs/migration-programme.md) | Programme phases and operating model. |
 | Find migration guidance | [`docs/migration-spec.md`](../docs/migration-spec.md) | `migration-spec.json` contract. |
-| Find migration guidance | [`docs/operator-runbook.md`](../docs/operator-runbook.md) | Manual pipeline operation. |
+| Find migration guidance | [`docs/operator-runbook.md`](../docs/operator-runbook.md) | Manual pipeline operation, the failure playbook, and §4.12 **Windows/PowerShell misreads** — `NativeCommandError` on a command that exited 0, long-path `Remove-Item`, `Start-Process -ArgumentList` mangling a multi-word `--workbook`, and `requests`+certifi failing TLS where the toolkit's `urllib` + Windows certificate store succeeds. |
 | Test the offline deployment rehearsal | [`docs/offline-mock-harness.md`](offline-mock-harness.md) | Offline Tableau-to-Fabric mock harness and fidelity boundary. |
 | Run the test suite | [`docs/parallel-test-loop.md`](parallel-test-loop.md) | Fast parallel loop vs the serial pre-PR gate, and the `serial` marker. |
 | Find migration guidance | [`docs/reference-capture.md`](../docs/reference-capture.md) | Tableau reference capture and evidence grading. |
@@ -84,7 +84,7 @@ Eligible files: tracked Markdown knowledge/navigation files, `.github/pbi.kb/**/
 | Improve review throughput | [`docs/review-throughput-postmortem.md`](../docs/review-throughput-postmortem.md) | Measured 2026-09-01/02 review-round costs, causes, counterexamples, and the replacement brief contract. |
 | Find migration guidance | [`docs/tableau-dax-translation-guide.md`](../docs/tableau-dax-translation-guide.md) | Translate Tableau calcs/LODs/table calcs. |
 | Find migration guidance | [`docs/tableau-map-to-azuremaps.md`](../docs/tableau-map-to-azuremaps.md) | Translate Tableau maps to Azure Maps. |
-| Find migration guidance | [`docs/windows-path-limits.md`](../docs/windows-path-limits.md) | Windows MAX_PATH ceilings a shipped bundle must respect. |
+| Find migration guidance | [`docs/windows-path-limits.md`](../docs/windows-path-limits.md) | Windows MAX_PATH ceilings a shipped bundle must respect — and how to DELETE a tree that is already over one (`\\?\` prefix plus a chmod handler, because git objects are read-only). |
 | Find migration guidance | [`docs/upstream-issue-gate.md`](../docs/upstream-issue-gate.md) | Route engine issues upstream vs local. |
 | Invoke subagent | [`.github/agents/dry-run-operator.agent.md`](../.github/agents/dry-run-operator.agent.md) | Full customer-shaped pipeline dry-run persona; read-only against the toolkit. |
 | Invoke subagent | [`.github/agents/pbi-migration-validator.agent.md`](../.github/agents/pbi-migration-validator.agent.md) | Read-only fidelity validator persona. |
--- a/docs/migration-phases.md
+++ b/docs/migration-phases.md
@@ -60,6 +60,21 @@ One numbered run directory per pipeline run, allocated by
 is decoration for human navigation; nothing may parse it back into a display name (two projects can
 legitimately hold same-named workbooks — issue #234, rule 2).
 
+⚠️ **"Per pipeline run" means per *invocation*, not per workbook.** Run 408 below is **one** run
+covering a 48-workbook site; its per-workbook units live inside that run's `bundle/pbip/`, never in
+48 sibling run directories. `unit_key` names what the run is *about* (a site, a project, or a single
+workbook when that is genuinely the whole job) — it is not a promise that each workbook gets a run.
+A customer's agent read `work_dirs.py`, inferred one-run-per-workbook, and renumbered **14** run
+directories in the resulting reorg, breaking the absolute self-paths inside each one (issue #470).
+
+⚠️ **Renaming an existing run is destructive, and it is now detectable.** `allocate_run` records the
+directory name it allocated, and `python scripts/work_dirs.py --verify` reports each run as
+`intact` (exit 0), `moved` (exit 1 — renamed or renumbered since allocation, naming both directories)
+or `unverifiable` (exit 3 — allocated before the check existed, so the question cannot be answered
+from the manifest). **`unverifiable` is not a pass**; every run predating the change lands there,
+including run 408 below. The comparison is on the directory *name*, so cloning or moving the
+repository — or working in a `git worktree` — does not report `moved`.
+
 `CANONICAL_SUBDIRS` (`scripts/work_dirs.py:72`) is the single source of truth for the layout:
 
 | subdir | produced by | contents | measured on run 408 |
--- a/docs/migration-phases.md
+++ b/docs/migration-phases.md
@@ -67,13 +67,26 @@ workbook when that is genuinely the whole job) — it is not a promise that each
 A customer's agent read `work_dirs.py`, inferred one-run-per-workbook, and renumbered **14** run
 directories in the resulting reorg, breaking the absolute self-paths inside each one (issue #470).
 
-⚠️ **Renaming an existing run is destructive, and it is now detectable.** `allocate_run` records the
-directory name it allocated, and `python scripts/work_dirs.py --verify` reports each run as
-`intact` (exit 0), `moved` (exit 1 — renamed or renumbered since allocation, naming both directories)
-or `unverifiable` (exit 3 — allocated before the check existed, so the question cannot be answered
-from the manifest). **`unverifiable` is not a pass**; every run predating the change lands there,
-including run 408 below. The comparison is on the directory *name*, so cloning or moving the
-repository — or working in a `git worktree` — does not report `moved`.
+⚠️ **Renaming an existing run is destructive, and it is now detectable.** `allocate_run` records both
+the directory name and the absolute path it allocated, and `python scripts/work_dirs.py --verify`
+reports each run as `intact` (exit 0 — recorded name *and* recorded path both still match), `moved`
+(exit 1 — renamed or renumbered since allocation, naming both directories) or `unverifiable`
+(exit 3 — the question cannot be answered from the evidence). **`unverifiable` is not a pass.**
+
+⚠️ **`--verify` is a GATE, so it never silently drops what it cannot assess.** A run directory whose
+`run.json` is missing, empty, truncated, malformed, locked, or holds valid JSON that is not an object
+counts as `unverifiable` and exits 3. It used to be skipped in silence and counted in *no* bucket at
+all, so a half-written run — exactly what an allocation interrupted between `mkdir` and the manifest
+write leaves behind — printed `0 run(s): 0 intact, 0 moved, 0 unverifiable`, exit 0.
+
+⚠️ **A run that is where it was allocated but at a different absolute path is `unverifiable`, not
+`intact`.** A basename is unchanged by a run-only move *and* by a copy, so the name alone cannot
+answer "still there": moving `source/_runs/001-acme` to `moved/_runs/001-acme` reported `INTACT`,
+exit 0. Whether the whole checkout moved (legitimate) or this one run was moved or copied
+(corruption) cannot be told apart from a single run's evidence — and **both** break the absolute
+self-paths embedded under `bundle/`, so neither is reported as healthy. Note that a clone and a
+fresh `git worktree` carry no runs at all (`_runs/` is git-ignored: `git check-ignore -v --
+_runs/001-acme/run.json` → `.gitignore:162:/_*`), so this is not a fresh-clone false alarm.
 
 `CANONICAL_SUBDIRS` (`scripts/work_dirs.py:72`) is the single source of truth for the layout:
 
--- a/docs/migration-phases.md
+++ b/docs/migration-phases.md
@@ -73,11 +73,19 @@ reports each run as `intact` (exit 0 — recorded name *and* recorded path both 
 (exit 1 — renamed or renumbered since allocation, naming both directories) or `unverifiable`
 (exit 3 — the question cannot be answered from the evidence). **`unverifiable` is not a pass.**
 
-⚠️ **`--verify` is a GATE, so it never silently drops what it cannot assess.** A run directory whose
-`run.json` is missing, empty, truncated, malformed, locked, or holds valid JSON that is not an object
-counts as `unverifiable` and exits 3. It used to be skipped in silence and counted in *no* bucket at
-all, so a half-written run — exactly what an allocation interrupted between `mkdir` and the manifest
-write leaves behind — printed `0 run(s): 0 intact, 0 moved, 0 unverifiable`, exit 0.
+⚠️ **`--verify` is a GATE, so it never silently drops what it cannot assess.** It rests on one
+invariant, implemented in one discovery function and one classification function: **every child
+directory of an existing `_runs/` root is a candidate run, any candidate whose identity cannot be
+positively established from its own evidence is `unverifiable`, and `intact` is only ever returned
+on positive proof.** Five shapes land there, and every one of them used to exit 0:
+
+| shape | what it printed before |
+|---|---|
+| `run.json` missing, empty, truncated, malformed, locked, or valid JSON that is not an object | `0 run(s): 0 intact, 0 moved, 0 unverifiable`, exit 0 — a half-written run, exactly what an allocation interrupted between `mkdir` and the manifest write leaves behind |
+| the directory renamed out of the `<NNN>-` pattern **and** its manifest deleted; or the `_runs/` root itself unreadable | `(no run directories found)`, exit 0 — discovery filtered on the *current* name and turned an `OSError` on the root into an empty list, while the tree was still on disk |
+| `run` contradicts the `<NNN>-<slug>` directory it sits in (`2`, or `10**100`, inside `001-acme`) | `1 intact`, exit 0 — only the *type* of `run` was checked, never its agreement with the directory. The number **is** the identity, so a manifest claiming a different one is contradictory evidence |
+| `allocated_abs_path` recorded as a **relative** path | `1 intact`, exit 0 — it was completed against the *verifier's* current directory, so the verdict depended on where the operator was standing |
+| recorded name matches but the recorded absolute path does not | see below |
 
 ⚠️ **A run that is where it was allocated but at a different absolute path is `unverifiable`, not
 `intact`.** A basename is unchanged by a run-only move *and* by a copy, so the name alone cannot
--- a/docs/operator-runbook.md
+++ b/docs/operator-runbook.md
@@ -1333,12 +1333,29 @@ workbook: a 48-workbook estate sweep is a single run whose per-workbook units li
 python scripts/work_dirs.py --verify          # exit 0 all intact / 1 something moved / 3 unverifiable
 ```
 
-⚠️ **Exit 3 is `unverifiable`, and that is NOT a pass.** A run allocated before the check existed
-recorded no self-path, so the question cannot be answered from its manifest — which is a different
-thing from answering it "fine". Measured on this machine: all three runs under `_runs/` report
-`unverifiable`, exit 3. Only `moved` (exit 1) is an established finding, and it names both the
-recorded and the actual directory. The comparison is on the directory **name**, so a repo that has
-been cloned, moved, or checked out as a `git worktree` stays `intact`.
+⚠️ **Exit 3 is `unverifiable`, and that is NOT a pass.** It means the question could not be answered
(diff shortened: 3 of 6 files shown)


Guust-Franssens:
## Blind review (issue + diff only, no author reasoning) — 3 findings, 2 High

Reviewed tip `22c1a1b`. All four requested gates independently exited 0, including **58 passed** in `tests/test_work_dirs.py`. Runtime mutation tests killed mutations to marker writing, renumber detection, legacy-state handling, malformed-marker handling, `list_runs` integration and exit-code selection.

The three findings are one problem wearing three hats: **`list_runs()` is inspection, `verify_runs()` is a gate, and the gate inherits inspection''s right to silently skip.** `list_runs`''s "never raises" contract is correct and should be preserved — a gate simply must not drop what it cannot assess.

### 1. High — fail-open verification (`scripts/work_dirs.py:469-480`)

`list_runs()` has three silent `continue`s: manifest missing (`:470-471`), unreadable or malformed (`:474-475`), and valid JSON that is not an object (`:479-480`). `verify_runs()` derives its counts from the result, so a run directory that exists but cannot be assessed contributes **zero to every bucket**.

Independently reproduced against the production CLI (a canonical `_runs/001-acme/`, first with no `run.json`, then with `not json at all`):

```
  (no runs with a readable run.json)
0 run(s): 0 intact, 0 moved, 0 unverifiable
EXIT=0
```

**The `unverifiable` bucket already exists in the output and stays at 0 exactly when something is unverifiable.** The reviewer also reproduced empty, truncated, and a manifest held open with Windows `FileShare.None` (`OSError`) — same result. This arises *naturally* when allocation is interrupted after `mkdir` but before `run.json` is written, so it is not an exotic input.

⚠️ **A new test enshrines it.** `tests/test_work_dirs.py:459-470` asserts `isinstance(runs, list) and not runs` — it *requires* the unassessable run to be omitted. Its underlying point is legitimate (`.setdefault` on a list raises `AttributeError`, escaping the `JSONDecodeError`/`OSError` guard), so it should not simply be deleted; inspection and gating need to be separated so both contracts hold.

### 2. High — a moved or copied run reports `intact` (`:332-335`, `:438`)

Allocation records only `run_dir.name`; verification compares only that basename. Allocating `source\_runs\001-acme`, moving that run alone to `moved\_runs\001-acme`, then running the CLI against both roots gave `INTACT 001-acme` exit 0 at the destination and zero runs exit 0 at the origin. Copying `001-beta` into a second `_runs` root also reported `INTACT`.

This matters because the run number is the identity and **bundle output embeds absolute self-paths** — a run at a new absolute location has stale self-paths regardless of its basename. The CLI claims to determine whether runs remain "where allocated".

The genuine difficulty: a whole-repository relocation is legitimate, a run-only move is corruption, and telling them apart needs stable allocation provenance. If they cannot be distinguished, the answer is `unverifiable` — not `intact`.

### 3. Medium — crash on a mixed tree (`:484`)

`manifests.sort(key=lambda m: m.get("run", 0))` sorts an unvalidated value. A tree holding both `{"run": 1}` and `{"run": "two"}` raises `TypeError: '<' not supported between instances of 'str' and 'int'`, violating the "never raises" contract and preventing an `unverifiable` verdict from being reached at all.

### Not verified

Customer-environment-specific documentation measurements (e.g. certificate-store counts) could not be checked — that environment was not available.

---

Routed to a fresh agent for the fix. Findings 1 and 3 have clear required behaviour; finding 2 is explicitly open on whether a run-only move can be distinguished from a whole-repo relocation, with `unverifiable` as the sanctioned answer if it cannot.

Guust-Franssens:
## Round 1 fix landed — independently verified

I re-ran the reproduction myself against the production CLI rather than taking the summary on trust.

| shape | before | after |
|---|---|---|
| `_runs/001-acme/` with **no** `run.json` | `0 run(s): 0 intact, 0 moved, 0 unverifiable` — **EXIT 0** | `1 run(s): 0 intact, 0 moved, 1 unverifiable` — **EXIT 3** |
| same with `run.json` = `not json at all` | **EXIT 0** | **EXIT 3** |

The output now also explains itself: *"''unverifiable'' is NOT ''intact'' - those runs could not be assessed."*

### The premise correction is the most valuable part, and it checks out

The name-only design was justified in-code by the claim that recording an absolute path "would fire on every clone / worktree". That is **false and checkable**, and I confirmed it:

```
git check-ignore -v -- _runs/001-acme/run.json
.gitignore:162:/_*    _runs/001-acme/run.json
```

`_runs/` is gitignored, so clones and fresh worktrees carry **zero** runs — nothing to fire on. The only remaining whole-tree relocation is moving the checkout, which breaks the absolute self-paths under `bundle/` exactly as a run-only move does, so exit 3 is honest there rather than noisy.

It also found that `test_a_relocated_repository_is_still_intact_not_moved` fixtured a `copytree` — i.e. **the corruption case** — and asserted the fail-open verdict. That test was pinning the defect.

### Finding 2 resolution

`intact` now requires the recorded name **and** a newly recorded `allocated_abs_path` to match. A relocated run is `unverifiable`, never `intact` and never `moved`, on the explicit grounds that a whole-checkout move and a run-only move are not distinguishable from one run''s evidence. That is the sanctioned answer from the review.

### Findings 1 and 3

Both closed: all six unassessable shapes now exit 3, and the mixed `{"run":1}` + `{"run":"two"}` tree returns `2 unverifiable` with no traceback instead of `TypeError`.

### Mutations: 7 applied, 7 killed

Notably **M1b moves the other side of the relationship** — inspection stops filtering — and is killed by the *inspection* list rather than the gate, which is the discipline that catches a mutation with too large a blast radius. M4 (path comparison always true) turned the relocation tests red while leaving the `intact` positive controls green.

### Residual, filed rather than blocking

- A run **deleted entirely** still reports 0 runs / exit 0 — there is no registry of expected runs (`_runs/INDEX.md` is #234 follow-up).
- Path comparison uses `abspath`, not `resolve`, so a run reached through a junction reads `unverifiable` — fail-safe direction, never a false `intact`.
- `FileShare.None` is exercised by monkeypatching `read_text` to raise `PermissionError`, not a real exclusive-share handle.

Gates re-verified by exit code: ruff 0, pylint 0, pytest 85 passed exit 0, `check_navigation_index` 0, `sync_agent_conventions --check` 0, full suite exactly six pre-existing engine-pin failures.

Queued for round-2 review by the same reviewer that found these, since it already holds the reproductions.

Guust-Franssens:
## Round 3 — **verdict: all three fail-open findings CLOSED.** One CI blocker remains, and it is a test defect.

Reviewed SHA `191d15341cd354e9f6e1bb4935f75be6193290e1`.

### The restructure worked
Fixed by **collapsing the surface**, not by adding three guards: 3 decision functions → 1, filtered discovery → unfiltered, and `RUN_LOCATION_INTACT` constructible at exactly one site.

| reproduction | before | after |
|---|---|---|
| relative `allocated_abs_path` from a matching CWD | `INTACT`, exit 0 | `unverifiable`, **exit 3** |
| renamed to `acme-without-number`, manifest deleted | `0 run(s)`, exit 0 | discovered, **exit 3** |
| `PermissionError` on `root.iterdir()` | `0 run(s)`, exit 0 | 1 unverifiable, **exit 3** |
| `run`=2 and `run`=10^100 in `001-acme` | `1 intact`, exit 0 | `unverifiable`, **exit 3** |
| 5,000-digit JSON integer | `ValueError` traceback, exit 1 | `unverifiable`, **exit 3**, no traceback |

Also confirmed: weirdly-named child directories are all discovered, regular files correctly excluded, and a matching derived name **cannot promote** a legacy run. The reviewer found **no remaining input returning `intact` without the full evidence ladder**.

### ⛔ CI is red — `tests/test_work_dirs.py:949-950`, a test defect

```python
(r"\_runs\001-acme", os.name != "nt")
("C:001-acme",       os.name != "nt")
```

Both are **relative POSIX paths**, so `_is_location_independent()` correctly returns `False` on Linux — the production code is right, the expectation is wrong. Linux CI: **4,696 passed, 2 failed**. Fix is to expect `False` on every platform.

⚠️ **This is a process failure on my side.** I set a batch policy of *"do not run the full suite, it runs once at the end"*, which hid Linux-only failures on this and two other PRs. New standing rule: local gates are necessary, **the CI conclusion is the gate**.

### Two corrections to the author''s claims

**The `isabs` justification overstates by one shape.** The claim was that `os.path.isabs()` misses two shapes. Measured: `C:001-acme` is **already rejected by `isabs()`**; only `\_runs\001-acme` is genuinely missed, and only on Windows Python 3.11. The replacement predicate remains justified and both facts are load-bearing — but the count is one, not two.

**The AST structural pins are weaker than described** — non-blocking, but worth recording since the pin is what is meant to stop a fourth construction site appearing later:

- `test_intact_is_constructed_at_exactly_one_place_in_work_dirs` **still passed** after adding a second construction via `state=alias`
- `test_discovery_applies_no_name_pattern_and_no_manifest_precondition` **still passed** when filtering was moved into an external helper

Production structure is correct today; the tests do not prevent indirect construction or an extracted filter.

### Also verified
The author''s self-reported vacuous test (first parametrized path test survived M1) is genuinely replaced with a load-bearing one. Control **C1** survives legitimately — reversing the anchor pair is a no-op because `all()` is order-independent.

Local gates all exit 0, **115 passed**, worktree clean.

PR #480, opened 2026-09-03
fix(oracle): name the zero-row views instead of only counting them

### LINKED ISSUE #471: Oracle capture: a zero-row view is recorded ok and never named (data_empty is counted but the list is discarded)
Reported by the SES team (Sandeep Munagala) from a live run against `tableauca.intelsat.com`
(Tableau Server 2025.3.3, REST 3.27), 38 workbooks / 94 views, project "Enterprise Dashboards".
Their trace is accurate on the mechanism. **Two of their three sub-claims are wrong, and the
correction narrows the fix** - so the report is reproduced here with each claim marked against
repo ground truth.

## The defect, stated correctly

A view whose `/data` export returns **zero data rows** is recorded `status: "ok"` and is
**not named anywhere**. The per-view `row_count` is in the manifest, and an aggregate count is
printed and stored - but the LIST of which views were empty exists only as a local variable that
is immediately reduced to its length.

Field impact: the reporter spot-checked 4 of 94 rendered PNGs and found all 4 blank (chrome only,
no chart body); the paired CSVs were 2 bytes. A later sweep put it at **12 genuine blank-data
views out of 91**. A fidelity reviewer signing off on this oracle has no way to know without
opening every PNG by hand.

## Claim-by-claim verification

| Their claim | Verdict | Evidence |
|---|---|---|
| `_capture_data()` returns `status: "ok"` unconditionally on HTTP success, merging `**summarise_csv(payload)` | ✅ **CONFIRMED** | `scripts/capture_tableau_oracle.py:849-857` |
| `summarise_csv()` already computes `row_count` at zero extra network cost | ✅ **CONFIRMED** | `scripts/tableau_payload_facts.py:59-71` |
| "Nothing reads that `row_count` back. I grepped the whole script." | ❌ **FALSE** | Three readers exist, all in *helper modules* rather than `capture_tableau_oracle.py` - which is why a single-file grep missed them: `tableau_oracle_manifest.py:81` (per-view console line prints `%5d rows`), `:137` (computes an `empty` bucket), `group_oracle_by_workbook.py:198` (per-workbook `data_empty`) |
| Suggested fix 2: "print a run-end summary line, e.g. N/94 views captured 0 rows" | ❌ **ALREADY EXISTS** | `tableau_oracle_manifest.py:305-316` prints `%d/%d captured (%d empty), ...`; the manifest carries `"data_empty"` at `:257` |
| Suggested fix 1: add an explicit per-view manifest flag, do not overload `status` | ✅ **VALID AND UNBUILT** | `empty` is computed as a **list** at `:137` and only `len()` escapes at `:257`/`:309`. No per-view flag, no named list |

## The strongest argument for the fix is already written in the same file

`tableau_oracle_manifest.py:264-268` justifies why `render_unestablished` is **counted AND named**:

> the count answers "is my reference set complete" and the list answers "for which pages is a
> visual-fidelity finding currently impossible to make" [...] the manifest must still SAY it, or a
> consumer has to re-derive it from three per-tier statuses and will not.

That reasoning applies verbatim to `empty` and was not applied to it. A zero-row capture makes a
**numeric**-fidelity finding impossible in exactly the way an unestablished render makes a visual
one impossible. This is an internal inconsistency, not a missing feature request.

## Second defect: "no rows" and "no query at all" are conflated

The reporter's follow-up correction: 2 of the 14 flagged views (`KPI Definitions`,
`Metrics Dictionary`) are **glossary/reference sheets with no underlying query at all**. `/data`
returning nothing there is correct behaviour, not a capture failure. Counting them as blank
overstates the defect (14 -> 12) and trains the reader to discount the whole list.

⚠️ **Measured, and it kills the obvious discriminator.** The tempting fix is to separate them by
the recorded `columns` (header present vs absent) or by the byte count. Neither works:

```
0-byte (glossary, no query)      bytes=  0 row_count=0 columns=[]
2-byte CRLF                      bytes=  2 row_count=0 columns=[]
header-only 1 col                bytes=  8 row_count=0 columns=['Region']
header + 2 rows                  bytes= 30 row_count=2 columns=['Region', 'Sales']
```

Both of the reporter's classes land in `row_count=0, columns=[]`, differing only in `bytes`
(0 vs 2) - and "glossary means 0 bytes, blank-data means 2 bytes" is an observation from one site
with n=14, not a documented Tableau contract. **Do not encode the byte count as the
discriminator.** The reporter's own suggestion is the principled one: a sheet with **zero
underlying data fields** is detectable from the workbook's own XML/metadata, which we already
parse.

## Scope

1. Name the empty views, not just count them: a per-view flag (`"flags": ["empty_default_state"]`
   or similar) plus a named list in the manifest. Do **not** overload `status` - the HTTP call did
   succeed, and `status` drives the exit code.
2. Raise per-view visibility to a WARN. `0 rows` in an INFO line among 94 is technically visible
   and practically invisible.
3. Classify zero-underlying-data-fields sheets separately from blank-data views, sourced from
   workbook metadata rather than the byte count.
4. Where cheap, carry the reporter's two observed root causes as sub-classes, since they need
   opposite remediation: **relative-date filter, data not landed yet** (re-runnable) vs
   **required filter defaulting to None** (never resolvable from a default-state capture; needs
   the `?vf_` state-pinning already tracked in #194).

## Explicitly out of scope

`?vf_` filter state-pinning (#194). This issue is about making the existing zero-row fact legible.

Reported-by: SES (Sandeep Munagala), 2026-09-03.

Guust-Franssens:
Fixes #471

A zero-row oracle capture was recorded as `status: "ok"` and the list of affected views was discarded, so a view that captured **no data at all** was indistinguishable in the manifest from one that returned 900,000 rows.

This is a customer-reported defect. Field report (SES, `tableauca.intelsat.com`, Tableau Server 2025.3.3, REST 3.27, 94 views, project "Enterprise Dashboards", 2026-09-03): four views were spot-checked by opening their rendered PNGs, all four showed chart bodies that were completely blank — only chrome, filter panel and logo — and all four paired `/data` CSVs were 2-byte header-only files. The manifest recorded every one of them as `ok`.

The row count was already being computed at zero extra network cost (`summarise_csv` in `scripts/tableau_payload_facts.py` returns `row_count`) and then never read back. Nothing warned, nothing rolled up, and finding these required knowing to grep `'"row_count": 0'` across the manifest after the fact.

## What changes

- The zero-row views are **named**, not merely counted.
- The signal is carried alongside `status: "ok"` rather than overloading `status` — the HTTP call did succeed, and conflating the two would lose that distinction.
- Downstream consumers (`group_oracle_by_workbook.py`, `manifest_scope.py`) carry the diagnostic through instead of dropping it.
- `docs/reference-capture.md` records the zero-row census where the oracle is described.

## Two distinct causes were observed in the field, and they need opposite remediation

1. **Relative "today" date filter with no data landed yet** — a near-real-time feed that had not ingested the day's data at capture time. Re-running later may fix it.
2. **A required filter defaulting to unselected (`None`, not `All`)** — a "pick a value to begin" pattern. This class will **never** produce evidence from a default-state capture no matter when it is re-run; it needs the `?vf_<field>=<value>` state pinning tracked separately in #194.

## Gates, measured by exit code

```
ruff check scripts tests                 exit=0
pylint (changed scripts)                 exit=0
pytest (targeted, 3 modules)             exit=0   535 passed
```

Includes a mutation campaign (`tests/mutation_oracle_empty.py`) and a scripted-HTTP-response test that proves the empty-capture chain end to end rather than asserting on a hand-built manifest.

## Note for whoever merges second

`tests/test_package_unit.py` (+39 lines here) is also being edited on `feat/package-canonical` (#463). Expect a merge conflict in that one file; the changes are additive on both sides.


Guust-Franssens:
## Blind review (issue + diff only, no author reasoning) — 2 findings, BOTH fail-open, both blocking

Reviewed SHA `868e47cd17736a0caeadd6b14ac76951324eeb51`. Gates verified green at review time: ruff **0**, pylint **0** (10.00/10), targeted pytest **0** (535 passed).

**The framing that matters: this PR fixes *"a zero-row capture reads as ok"* and reintroduces the same class one level up — *"an unassessable capture reads as ok"*.**

### 1. Medium, fail-open — missing `row_count` is treated as clean evidence (`tableau_oracle_manifest.py:114-116`)

```python
data = record.get("data") or {}
if data.get("status") != "ok" or data.get("row_count") != 0:
    return None
return EMPTY_QUERY_NO_ROWS if data.get("columns") else EMPTY_CANNOT_CLASSIFY
```

When `row_count` is **absent**, `data.get("row_count")` is `None`, `None != 0` is true, and the function returns `None` — which every consumer reads as *"not empty"* rather than *"cannot determine"*. Confirmed at source.

**This is a regression in the failure boundary, which is why it blocks:** the previous predicate raised `KeyError` (fail-closed); this PR converts that into a clean result.

Reproduction — `{"data": {"status": "ok", "columns": ["Region"]}}` with no `row_count`:

| consumer | observed |
|---|---|
| capture manifest | `exit=0 status=ok data_ok=1 captured_complete=1 data_empty=0 data_empty_views=[] flags absent` |
| `group_oracle_by_workbook.subset_manifest()` | `data_ok=1 data_empty=0 failed=0` |
| `package_unit._scope_oracle_manifest()` | `data_ok=1 status=ok row_count absent flags absent` |

Note the module already reasons carefully about honest states — `EMPTY_CANNOT_CLASSIFY` exists, with a docstring arguing the third state should be *named as missing rather than invented*. That reasoning simply was not applied to a missing `row_count`.

### 2. High, fail-open — any HTTP 200 body is accepted as numeric evidence (`capture_tableau_oracle.py:841-856`, `tableau_payload_facts.py:59-71`)

`_capture_data()` ignores **Content-Type** and records every 200 as `status: "ok"`; `summarise_csv()` decodes arbitrary bytes with replacement and uses a **non-strict** `csv.reader`, so any first line becomes a header and any further line a row.

| response | result |
|---|---|
| 200 `text/html`, `<html>\n<body>Error</body>\n</html>\n` | `status=ok`, `columns=["<html>"]`, **`row_count=2`**, no flags |
| 200 `text/csv`, `Region,Sales\r\nWest,"unterminated` | `status=ok`, **`row_count=1`**, no flags |
| 200 `application/octet-stream`, `not CSV at all` | `status=ok`, `row_count=0`, **classification `empty_query_no_rows`** |

The third is the worst: it is **confidently wrong**, reporting a specific diagnosis ("the query ran and returned no rows") for a payload never established to be CSV.

This also invalidates the claim at `docs/reference-capture.md:520` that a returned "header" proves a query ran.

### Verified as already good — must not regress

- A **missing** CSV correctly downgrades to `not_copied`, `data_ok=0`, `failed=1`.
- An injected `PermissionError` on an unreadable CSV **propagates** rather than reporting success.
- **The PR''s own mutation harness was independently audited and is honest**: all 21 snippets compile, every behavioural mutation executed its patched path 1–6 times, one cosmetic control survived, one invalid control had zero hits.

### Not verified

Behaviour against a live Tableau Server — no live credentials or site were used.

---

Routed to a fresh agent. Both findings are fail-open and block; the fix must make "cannot assess" an explicit state in both places without collapsing the genuine transport `status: "ok"` distinction.

Guust-Franssens:
## Round 2 — **BLOCK.** Both original findings closed; two NEW fail-open paths at surfaces round 1 did not cover.

Reviewed SHA `8c9cfd0c6b860ec0fd1ac93a2bf8abcd7236a4be`. Gates verified: ruff **0** · pylint **0** on all six changed scripts · targeted pytest **576 passed** · mutation harness **0** (39 caught / 2 survived / 1 invalid, all as declared).

### Confirmed closed
The three original malformed payloads fail closed with **no file written**. An absent Content-Type genuinely reaches the named unassessable state. A legacy manifest (no flags, no `row_count`) is now caught by `_scope_oracle_manifest`. Ragged rows → `format_mismatch`, no file, exit 3 — the unverified rectangularity assumption is **fail-closed**, non-blocking.

**The harness was independently audited and is honest**: all 42 mutations compile, every behavioural mutation hits its intended path, and the previously-reported import-order false positive is genuinely fixed — **both** `tableau_payload_facts.certify_csv` and `capture_tableau_oracle.certify_csv` are patched. The data-only certification mutation does change the production allowlist.

### 1. High, fail-open — a FOURTH consumer reads unassessable CSVs as numeric ground truth
`scripts/build_reconcile_items.py:162-181`

`build()` ignores `flags`, `certification`, `row_count` and `data_unassessable` entirely: any data leg with `status: "ok"` and a `path` is read and emitted as **Tableau numeric ground truth**.

```
manifest:  status=ok  path=data/view.csv
           flags=[data_unassessable, content_type_absent]
           captured_complete=0  data_unassessable=1

build_reconcile_items:  item_count=1  skipped_views=[]  tableau_value=10.0
```

Identical result for a legacy record with `row_count` absent. A fidelity verdict built on this is unfounded.

### 2. High, fail-open — `text/plain` certifies arbitrary error text as CSV
`scripts/tableau_payload_facts.py:61-64`, `:118-160`

```
HTTP 200  Content-Type: text/plain
body: Service unavailable\r\nRetry later\r\n

status=ok  certification=certified  columns=["Service unavailable"]  row_count=1
captured_complete=1  data_unassessable=0  flags ABSENT  file written  exit=0
```

The single-line variant `Service temporarily unavailable` was recorded `row_count=0` and **specifically diagnosed `empty_query_no_rows`** — confidently wrong, which is exactly the bug #471 exists to fix, recreated under an accepted MIME type.

## The pattern, and why the next round changes structure rather than adding a guard

Round 1 fixed **three named consumers**; round 2 found a **fourth**, plus a hole in the certification predicate itself. That is a fix that flags a bad state and relies on every consumer remembering to check the flag — and this is now the second PR in this batch showing the identical shape (#477 narrowed its class one site at a time across three rounds).

Applying this repo''s own stop rule — *"do not add another local guard: simplify, delete, split, or descope"* — the next round implements:

> **An uncertified capture must be impossible to consume as evidence, not merely flagged as uncertified.**

Concretely, the reviewer''s own suggestion: do not place uncertified retained bytes at a `data/*.csv` path that numeric-oracle consumers interpret as evidence. Plus `text/plain` demoted to unassessable, alongside absent-Content-Type — `text/csv` and `application/csv` remain explicit declarations and stay accepted.

The durable test is the one that proves a **new, unmodified** consumer cannot read an uncertified capture as evidence. Without that, the fifth consumer is a matter of time.

Guust-Franssens:
## Round-7 blind review at `8eadc301` — REQUEST CHANGES (2 blocking)

**Round 7's fix itself is correct and confirmed.** The platform half is properly closed, verified by driving both predicates under both flavours:

| input | Win non-rel | POSIX non-rel | Win unsafe | POSIX unsafe | current |
|---|---|---|---|---|---|
| `C:\Users\x` | T | **F** | T | **F** | **T** ✅ |
| `sub\..\x` | F | F | T | **F** | **T** ✅ |
| `\\server\share\x`, `/home/x`, `/Users/x` | T | T | T | T | T ✅ |

⚠️ My brief said "only the drive-letter form flipped" — that is true for the *non-relative* half only. **Backslash traversal flips too**, in the full unsafe predicate. Round 7 correctly fixed both; my summary was incomplete.

**#499's fixture was NOT weakened** — checked with a four-variant control:

| fixture | result |
|---|---|
| both certified, correctly typed | PASS; visual 2, numeric 2 |
| worksheet uncertified | NOT_CHECKED; numeric 1 |
| wrong worksheet type | NOT_CHECKED; visual 0, numeric 0, contested |
| both uncertified | NOT_CHECKED; numeric 0 |

Adding `certification` stops the test accidentally exercising legacy demotion instead of typed resolution; `test_a_legacy_uncertified_record_is_not_numeric_evidence` still holds the legacy negative. Assertions untouched.

Mutation harness **31/31 caught, negative control SURVIVED, exit 0**. The name-only spot-check behaves exactly as claimed: 2 failed, and the round-5 name test survived while both round-6 tests failed. All gates exit 0.

---

## The two blocking findings are NEW vectors, not the old one moved

Both defeat containment for a reason the intake approach does not cover.

### B1 — a host path EMBEDDED IN PROSE is not detected

`_declares_unsafe_path()` (`package_unit.py:499-515`) parses the **entire string** as a path. `classify_export_error()` prefixes response text with `HTTP 503: ` (`capture_tableau_oracle.py:278,597`) and persists it in `retry_reasons`:

```
raw:               C:\Users\blind-review-account\private\retry.log could not be opened
classified detail: HTTP 503: C:\Users\blind-review-account\private\retry.log could not be opened
package predicate: False
shipped artifact:  oracle/oracle-manifest.json
```

Quoted paths and `file:///C:/Users/...` fail the same way.

**This repo's own privacy regex already detects it** — `set_data_folder.py:31,108`. So the fix is to reuse that detector, not invent a third one: the predicate must ask *"does this string CONTAIN a host path?"*, not *"is this string a host path?"*

### B2 — an untrusted JSON KEY is re-emitted after containment

`_contain_unsafe_strings()` cleans dictionary **values** but preserves raw **keys** (`package_unit.py:551-553`). `project()` then turns an unknown key into a diagnostic path (`manifest_scope.py:141-143`). Compounding it, `_scope_oracle_manifest()` does its final containment at line 826 and then appends `dropped_fields` at line 838 — **after** the last containment pass.

Injected view field `"C:\\Users\\blind-review-account\\private\\secret": "safe scalar"` ships as:

```json
"scope": { "dropped_fields": ["views[].C:\\Users\\blind-review-account\\private\\secret"] }
```

`tableau_env.py:461-478` already documents that keys must be scrubbed and diagnostic paths must use the scrubbed key.

### Escapes that FAILED (the containment is holding elsewhere)

✅ `path` / `retained_path` / `packaged_from` values · nested list values in `retry_reasons`/`dimensions_px` · `view_name`, `flags`, `view_luid`, `view_type` contained before filenames, objects, omissions and handover derivation · provenance LUIDs · no consumer re-reads raw views after `package_oracle` · a dynamic `format_hints` key was dropped without leaking.

Clean containment is canonical-JSON byte-identical with `refused=[]`; relative `data/view-0.csv` survives; refusals name only the JSON path.

---

Two sub-classes, both fail-open leaks of customer-facing data, so both block. #516 (single rooted backslash) remains out of scope and does not block.

Guust-Franssens:
## Round-8 blind re-review at `1d30c920` — REQUEST CHANGES

**Both original reproductions are closed**, and two structural claims verified:

| B1 input | now |
|---|---|
| bare `C:\Users\<account>\...` | refused |
| `<path> could not be opened` (suffix) | refused — and the author's correction that this was *already* caught is **true**, the drive sits at index 0 |
| `HTTP 503: <path>` · quoted · `file:///C:/Users/...` | refused |

**B2 oracle key: closed** — the injected key now ships as `views[].<refused-by-packager>`, raw key absent everywhere.

**The ordering claim is verified, not just asserted.** Final containment is `package_unit.py:922`; the only subsequent mutation is assigning `refused_fields` at `:924`, then return, and the caller writes the returned object unchanged. No other post-sweep write exists.

**Re-anchored mutation: legitimate, not a weakening.** With `_declares_non_relative()` forced false, the original profile test **passes** (masked by the new disclosure layer) while the new `D:\builds\...` test **fails** — so the new anchor genuinely isolates the parse branch that the old one can no longer observe.

**False-positive control holds:** old tip vs HEAD, 14 files vs 14 files, identical lists, no canonical/byte diffs beyond `parsed_at`, `refused_fields` absent on both. Mutation harness **36/36, control SURVIVED, exit 0.** All gates 0.

---

## Still blocking: the guard's GRAMMAR is narrower than its invariant

The shared detector only recognises **profile-root** paths, so wrapping defeats it for every other absolute form:

```
HTTP 503: D:\builds\out\secret.log could not be opened
HTTP 503: \\customer-server\finance-share\secret.log could not be opened
HTTP 503: /var/lib/tableau/secret.log could not be opened
  -> discloses_host_path=False  _declares_non_relative=False  _declares_unsafe_path=False
  -> packaged retry_reasons value = KEPT, and shipped inside the handover slice
```

And the decisive one — a **real profile path with a real account name**, via the administrative-share UNC spelling:

```
HTTP 503: \\server\C$\Users\blind-review-account\private\secret.log could not be opened
  -> DISCLOSES=False  UNSAFE=False  SHIPPED=True
  -> account present in shipped bytes = True
```

Extended-UNC, redundant-separator and percent-encoded profile spellings escape the same way.

This contradicts the PR's own new anchor, which states `D:\builds\...` names the operator's machine and must be refused — allowing the identical location merely because `HTTP 503: ` precedes it cannot be right.

## Still blocking: the handover-key residual now HAS a reproduction

Round 8 declined to contain handover keys, citing "engine-authored, no reproduction". The reviewer built one, using the same hypothetical-future-field method the existing handover **value**-redaction test already uses:

```
HANDOVER_TOP_KEY_PRESENT=True
HANDOVER_NESTED_KEY_PRESENT=True
HANDOVER_RAW_TEXT_PRESENT=True
```

The customer-shipped `handover/Book.json` carries the private path. "Engine-authored" describes where a key comes from; it does not enforce the shipping invariant. With a reproduction in hand the stop rule no longer applies.

---

## The ask is ONE change, not another patch

Both residuals are the same defect: **the predicate matches a narrow spelling instead of the property.** Round 3-8 each widened that grammar slightly and each time a spelling escaped.

Make `discloses_host_path` answer *"does this text contain an absolute location on some host?"* for **every** absolute form — drive-letter, UNC (including `\\server\C$\...` and extended `\\?\`), POSIX-absolute, and normalised spellings (redundant separators, percent-encoding) — rather than profile-rooted paths only. Then apply the existing containment to handover keys as well as values.

The false-positive control above is the guard on that widening, and it must stay byte-identical.

⚠️ Watch one trap: `D:\builds\...` is currently a **deliberate** non-profile anchor proving the parse branch stays observable. Widening the detector will mask it exactly as it masked the profile test. Pick a new isolating case for that anchor rather than deleting it.

#516 remains out of scope.

Not verified by the reviewer: the full 5,261-test suite, a live Tableau/engine capture, whether today's engine naturally emits path-bearing handover keys, and Linux execution.

PR #482, opened 2026-09-03
fix(harvest): kill a STALLED download, not a slow one — and make the sweep tally add up

### LINKED ISSUE #472: harvest_estate_assets: a download failure is reported as SUCCESS in parse-sweep.md, the 600s ceiling is hard-coded, and a failed datasource silently orphans its workbooks
Reported by SES 2026-09-03, from a live 47-asset harvest against Tableau Server 2025.3.3 (REST 3.27):
**41/47 succeeded, 6 failed** — 2 `download failed`, 2 `timeout after 600s`, 2 `failed (HTML error page returned)`.

Their agent handled it correctly (it stopped and asked, per Gate B, and spotted a dependency risk on its
own). Everything below is about what the toolkit failed to tell it.

## 1. `parse-sweep.md` reports a download failure as SUCCESS — measured, not inferred

`summarise()` buckets rows by `ours.ok` / `theirs.ok`. A row that never downloaded has **neither key**,
so `r.get("ours", {}).get("ok") is False` is `False` and it lands in **no bucket** — while still counting
in `len(results)`.

Probe against the real `summarise()` with their exact 41-ok / 6-failed shape:

```
**47 asset(s)** — ours failed 0, his failed 0, both parsed 41.

## Our parser failed

_none_
```

- `47 != 0 + 0 + 41`. Six assets vanish into an unexplained arithmetic gap.
- The word **"download" never appears** in `parse-sweep.md` (measured `False`).
- **Neither failed asset name appears** (measured `False` for `DS_Sessions_by_Product` and
  `WCL Performance Dashboard`).

So the committed artifact reads **"nothing failed"**. The failure list exists only in the live console
and in `parse-sweep.json` rows — lose the terminal and you lose the finding. This is the **same defect
class as #471**: present in the denominator, never named in the summary. The rule that file already
states for itself — *"Group failures by SHAPE, because one root cause repeated 12 times is one feature
request"* (`harvest_estate_assets.py:462`) — is exactly what is not applied to download failures, even
though 6 of theirs collapse into 3 shapes.

**Fix:** a `## Downloads that never landed` section grouped by shape, plus make the header arithmetic
close (`ours failed N, his failed N, both parsed N, never downloaded N`).

## 2. The 600s download ceiling is hard-coded, and it is a WALL CLOCK over a per-read timeout

`harvest_estate_assets.py:387` — `subprocess.run(cmd, ..., timeout=600)`. No flag, not settable.

The engine's own downloader uses `urlopen(..., timeout=300)`
(`fetch_tds.py:407,423`), which is a **per-read socket timeout, not a transfer deadline**. A large
`.twbx` that streams slowly but steadily never trips 300s and is killed by our 600s wall clock — with
partial progress discarded. Same shape as #253 (a hard ceiling too low for real customer data, not
configurable); note #253 raised the *refresh* ceiling to 3600s and left this one untouched.

**Fix:** `--download-timeout` (default 600), and say the elapsed time in the failure detail so a
timeout is distinguishable from a hang.

## 3. Upstream: the download leg has NO retry, while the metadata leg has bounded retry

`fetch_tds.py` has a well-built `_http_json` with bounded exponential backoff, `Retry-After` honouring,
`401002` session-loss reauth, and fail-fast on the credential class (`:330-337`). But
`download_workbook` and `download_datasource` call **`_http` directly** (`:407`, `:423`), so they get
**one attempt and no retry**. A gateway blip is retried when listing workbooks and fatal when
downloading one.

That means all 4 non-timeout failures above got exactly one attempt — which is the main reason a plain
re-run is likely to clear them, and it is worth fixing upstream rather than papering over here.

→ File upstream as `Yarbrdab000/tableau-fabric-skills`. Keep this issue for legs 1 and 2, which are ours.

## 4. A failed DATASOURCE silently orphans the workbooks that bind to it

Their agent worked out by hand that `DS_Sessions_by_Product` failing means `IA IFC Sessions`
(which downloaded fine) would convert to an incomplete model. The harvester already knows that edge —
`dependency_datasources()` resolves exactly these bindings, and the run log even prints
*"plus N datasource(s) pulled in because a selected workbook binds to them"* (`:762`). Nothing
cross-references a **failed** datasource back to its dependents.

This is the documented "empty report" failure that migration ordering exists to prevent, and right now
catching it depends on the operator noticing.

**Fix:** after the sweep, for every datasource in the failed set, name the successfully-downloaded
workbooks that bind to it and mark them **do not convert yet**.

## Out of scope

`?vf_` state pinning (#194); the per-unit status roll-up (#469).

Guust-Franssens:
# What broke, and why raising 600 would not have fixed it

A customer's 47-asset harvest lost `IA Redemptions by Campaign Report` to `timeout after 600s`,
twice. There are **two nested timeouts and they catch opposite failures**:

| where | value | kind |
|---|---|---|
| engine `fetch_tds.py:407` / `:423` | `timeout=300` into `urlopen` | **per-socket-read** |
| ours, `harvest_estate_assets.py:387` | `subprocess.run(..., timeout=600)` | **total wall clock** |

A *stalled* connection trips the inner 300s. A **slow but healthy** download satisfies every
individual socket read, never trips 300s, and is then killed by our wall clock — so the customer's
asset was very likely progressing the whole time and we killed a working download. Raising 600 to
some larger number only moves that cliff, so the ceiling is not the fix: knowing whether bytes are
still arriving is.

**A file-growth watchdog cannot work here**, and that is a property of the engine rather than a
guess: `_http` ends `return resp.status, dict(resp.headers), resp.read()` — one unbounded read, the
whole asset buffered in RAM — and `save_outputs(raw, ...)` writes only afterwards. The destination
file is 0 bytes until the download has already finished. Verified in the installed plugin before
anything was built.

## The progress-signal measurement

Faithful simulation, 2026-09-03, Windows: a child running the same `urlopen(...).read()` shape
against a local slow-stream server, 15 MB at ~1 MB/s, versus a socket that goes quiet after 3 chunks.
The healthy child ran to completion (`stdout = 15728640`); the stalled one was killed at 20s.

| sampled | healthy download | stalled download |
|---|---|---|
| `Popen.pid` `OtherTransferCount` | Δ **0** (frozen) | Δ **0** |
| real interpreter `ReadTransferCount` | Δ 0 after startup | Δ 0 |
| real interpreter **`OtherTransferCount`** | **168688 → 170704, moves every second** | **frozen at 168752 for 19s** |
| `PagefileUsage` | frozen after startup | frozen |

Three things that settles, each of which would have produced a **wrong fix** on its own:

1. ✅ **The signal is sound, but it is LIVENESS, not volume.** Socket bytes land in Windows' *Other*
   I/O bucket and are counted as per-operation overhead — ~160 B/s observed while ~1 MB/s flowed. It
   is never reported as "bytes downloaded".
2. ❌ **The RAM-footprint half of the hypothesis is UNSOUND.** `PagefileUsage` did not move while
   15 MB was buffered; committed memory is taken in large chunks up front. Measured, not assumed —
   and reported as a negative result rather than quietly dropped.
3. ⚠️ **`Popen.pid` is the WRONG PROCESS under a uv venv.** `.venv/Scripts/python.exe` is a
   trampoline: measured `Popen.pid = 35152` while the child's own `os.getpid()` reported `20856`.
   Sampling the handle `Popen` hands you measures a stub whose counters never move, so the naive
   design would have declared **every** download stalled and killed all of them. The probe therefore
   sums the whole process **subtree**, and `terminate_tree()` kills the subtree too — `Popen.kill()`
   alone would have orphaned the real interpreter with the socket and session still open.

### The measurement was wrong twice before it was right

My first two runs showed *all* counters frozen, including for the healthy download — which looked
like a clean negative result and would have been reported as "the hypothesis is unsound, fall back to
a configurable ceiling". It was my harness that was broken: ctypes was called **without `argtypes`**,
so the 64-bit `HANDLE` was passed as `c_int`. A **positive control** is what caught it — a child that
allocated 60 MB and did 500 file reads showed the same frozen counters, which is impossible; a
self-process check then returned `ok_io=False` outright. With `argtypes` set, the control moved
+105 MB of memory and +3.3 MB of reads, and only then were the download numbers above trustworthy.

That is why the shipped `kernel32()` binding sets `argtypes` with a comment saying so: the failure
mode is not a crash, it is plausible-looking values that never change — which reads exactly like a
stalled download.

## What the fix does

* **`--download-stall-timeout` (default 120s)** — armed **only after movement has been observed at
  least once** for that child. A flatline is then a genuine stall and killing it is right.
* **`--download-timeout` (default 600s, `0` disables)** — the wall clock, armed **only while no
  movement has ever been observed**, i.e. exactly when we cannot tell slow from hung. That is the
  pre-fix behaviour, kept for the only case that justifies it. A download that keeps progressing past
  it is **not killed**; it is announced loudly instead.
* **Actionable failure text.** `timeout after 600s` told the operator only what *we* had configured.
  Now a stall says what was observed (`stalled: no progress for Ns (elapsed Ns)`) and names
  `--download-stall-timeout`; a ceiling kill says `no download-progress signal was available for this
  process, so a slow-but-healthy transfer cannot be told from a hang` and names `--download-timeout`.
* **A 60s heartbeat** reporting elapsed time and when progress was last seen, per AGENTS.md's "never
  block silently".

## Which legs of #472 this covers

| leg | status |
|---|---|
| 1. `parse-sweep.md` reports a download failure as SUCCESS | ✅ fixed — `## Downloads that never landed` grouped by shape (digits collapsed, so two timeouts differing only in elapsed seconds are one finding), a disjoint-bucket closure line that must sum to the denominator, and `report_failed_downloads()` naming them beside the closing tally. `download_error` was previously written into the record and surfaced nowhere an operator reads. |
| 2. the 600s ceiling is hard-coded, and it is a wall clock over a per-read timeout | ✅ fixed — above |
| 3. the download leg has no retry while the metadata leg does | ⛔ **upstream, not this PR** — already filed as `Yarbrdab000/tableau-fabric-skills#190`. The plugin tree is read-only to us. |
| 4. a failed datasource silently orphans the workbooks that bind to it | ✅ fixed — `dependency_edges()` keeps the workbook end of the same LUID-first/name-fallback join; `orphaned_dependents()` reports only workbooks that **did** land, in a `## Do not convert yet` section and a `DO NOT CONVERT YET` warning. |

Fixes #472

Legs 1, 2 and 4 are ours and are all covered; leg 3 is upstream by the issue's own routing ("Keep
this issue for legs 1 and 2, which are ours").

## Gates, by exit code

| gate | exit |
|---|---|
| `ruff format` + `ruff check --fix` (script + tests) | **0** |
| `pylint scripts/harvest_estate_assets.py` | **0** (10.00/10 — gated on `$LASTEXITCODE`, not the printed score) |
| `pytest tests/test_harvest_download_watchdog.py -q` | **0** — 43 passed |
| `pytest tests/test_harvest_{project_scope,output_guard,engine_gaps,download_watchdog}.py -q` | **0** — 246 passed, no regression |
| `python tests/mutate_harvest_watchdog.py` | **0** — 25/25 killed, both discriminating controls correct |

The full suite is deliberately **not** run here: repo policy for this batch is one full-suite run at
the end, after reviews settle.

## Round 2 — two ways this could still have killed a HEALTHY download

Blind review found four issues; two are filed as issues, and these two were fixed here because
although they are fail-closed, both defeat the purpose of the PR — #472 exists because we killed a
healthy download.

**1. The stall default was stricter than the fetcher's own read timeout.** 120s sat *below* the
engine's 300s per-socket-read timeout, so a bursty source pausing 120–300s between chunks satisfied
urllib and was killed by us and reported as hung: the wall-clock cliff had quietly become a shorter
inactivity cliff. The 120s came from analogy with `refresh_pbip_model.py`'s liveness timer — but that
one **warns** and this one **kills**, which is a different thing. Default is now
`ENGINE_READ_TIMEOUT_SECONDS + 120s` grace, so the child's own timeout always fires first and reports
the real error, and the CLI warns when an operator sets a value under 300s.

*Reproduced before the fix:* `BURSTY_HEALTHY stalled 0.49` → now `'' 2.31` (survives).

**2. Losing the probe after movement killed a healthy download.** `None` means "cannot read", but
once `progress_observed` was set, later `None` readings did not disarm the stall deadline. Access
denial, a descendant exiting between enumeration and counter read, or a **partial subtree read** all
produced it — and the partial case is the nasty one, because when the descendant carrying the network
I/O becomes unreadable, the readable trampoline flatlines, which is exactly a stall's signature.
Current *availability* is now tracked separately from historical *movement*: `transferred_bytes()`
returns `None` for an incomplete reading rather than a sum it cannot vouch for, losing the signal
falls back to the wall clock **measured from the moment of the loss** (not from process start, so a
transfer that progressed for nine minutes is not killed one minute later), and the loss is announced
once.

*Reproduced before the fix:* `PROBE_LOST_AFTER_PROGRESS stalled 0.33` → now `ceiling 5.15`, with the
honest label and a warning.

Both fixes are pinned by behavioural tests, not just by an inequality between two constants: the
bursty test scales every value from the real constants by one factor, so lowering
`DEFAULT_STALL_TIMEOUT` back under the read timeout turns it red while the pause and the child's read
timeout stay put.

## Mutation testing: 25/25 killed, and the campaign is now committed

`tests/mutate_harvest_watchdog.py`, ported onto the repo's existing campaign idiom
(`mutate_revision_key.py`, `mutation_capture_legs.py`). Previously the campaign lived in a gitignored
scratch folder and only its tally was quoted — a tally nobody can re-run is a claim, not evidence.
Every mutant must clear three bars before it is called KILLED: it must **compile** (a mutant that
does not import never executed the mutated path), its **anchor** is a pytest node ID rather than a
file (under `-x` a whole-file target credits a mutation to whichever test fails first), and its
**control** must stay green (so a mutation that reddens the module is not mistaken for one the anchor
observed). Plus the two discriminating controls: `control-cosmetic` **SURVIVED** as required and
`control-absent-anchor` was reported **INVALID** as required — which is what makes the other 25
verdicts worth reading.

⚠️ Two mutants had to be rewritten during the port, and the reason is itself a finding: **the round-2
fixes made the original mutations no longer reproduce their defect.** Disabling the stall arm no
longer restores the pre-#472 wall clock (because `blind_since` is cleared on movement), and flipping
`signal_live` alone no longer re-arms the stall deadline (because `blind_since` still disarms it).
Both now patch the pair of lines that genuinely restores the old behaviour.

Two findings from the first campaign, both worth stating because they are the failure modes that make
a mutation harness worthless:

* `timeouts-not-passed-through` (replace `args.download_timeout` with the module default at the call
  site) **SURVIVED** the first round — a genuine gap: every test called `download()` directly, so
  nothing asserted that `main()` forwards the flags. Parsing a flag and then ignoring it looks
  exactly like having no flag at all, which is the customer's original complaint.
  `test_main_forwards_the_operator_s_timeouts_to_every_download` now kills it.
* `trust-an-unmoved-probe` was initially scored **SURVIVED** because my mutation did not compile —
  and an earlier version of my scratch harness would have scored a non-compiling mutation as
  **caught**. That is the same class as this repo's measured 22/22-false-kills case. The committed
  campaign now refuses any verdict for a mutant that does not compile or whose anchor selected no
  test. (The repo's shared `mutation_harness.py` labels this `HARNESS-ERROR` via `recorded=False`;
  my earlier summary called it `INVALID`, which was the wrong name for the right behaviour.)

Full tally, all **KILLED**: `stall-never-fires`, `ceiling-kills-a-progressing-download`,
`trust-an-unmoved-probe`, `no-subtree-walk`, `no-descendant-kill`,
`stall-default-back-under-the-fetchers-read-timeout`, `a-lost-probe-counts-as-a-flatline`,
`partial-reading-passed-off-as-a-total`, `blind-ceiling-measured-from-process-start`,
`losing-the-signal-is-not-announced`, `no-cli-warning-for-a-stall-timeout-under-300s`,
`silent-heartbeat`, `no-over-ceiling-warning`, `stall-message-loses-the-flag`,
`ceiling-message-loses-elapsed`, `never-downloaded-is-invisible`, `shape-keeps-the-digits`,
`parser-buckets-count-a-failed-download`, `no-operator-report`, `timeouts-not-passed-through`,
`orphans-never-computed`, `orphans-flag-every-landed-workbook`,
`orphans-include-a-workbook-that-itself-failed`, `orphan-edges-lose-the-name-fallback`,
`orphans-not-written-to-the-report`.

## What I could NOT verify

* **No live Tableau server.** There are no credentials here, so the watchdog has never run against a
  real `fetch_tds.py` download. The measurement is a faithful simulation of the same `urlopen(...)
  .read()` shape, not the real thing.
* **Windows only.** The Linux `/proc/<pid>/io` (`rchar`/`wchar`) path is written to the documented
  interface and is exercised by no test on this machine — nothing in CI runs it either. It fails
  safe: an unreadable `/proc` yields `None`, which is "cannot tell" and falls back to the ceiling.
* **The 420s default stall timeout is a judgement, not a measurement.** It is anchored to the
  fetcher's own 300s per-read timeout plus a 120s grace, which is a defensible floor rather than an
  observed one; a source that legitimately goes quiet for longer would still be killed, and the flag
  and the message exist for that.
* **A spurious mover.** A child doing unrelated I/O forever would keep the probe moving and never be
  killed. That fails in the safe direction (we do not kill a working download) and is made visible by
  the heartbeat, but it is a real residual.

### DIFF
--- a/scripts/harvest_estate_assets.py
+++ b/scripts/harvest_estate_assets.py
@@ -50,17 +50,21 @@ fresh-per-asset completed 8/8. Slower, and the only thing that finishes.
 from __future__ import annotations
 
 import argparse
+import ctypes
 import json
 import logging
 import os
+import re
 import sqlite3
 import subprocess
 import sys
+import threading
 import time
 import traceback
 from collections import deque
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 from concurrent.futures import Future, ThreadPoolExecutor
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
@@ -81,6 +85,12 @@ from engine_source import EngineNotFoundError, engine_scripts_dir  # noqa: E402 
 from tableau_env import engine_child_env, pat_secret, redact, require, resolve_env  # noqa: E402  # pylint: disable=wrong-import-position
 
 LOG = logging.getLogger("harvest_estate_assets")
+
+# The download watchdog (below) pushes this module past pylint's 1200-line limit. The precedent in
+# this repo is a module-level waiver (`assess_estate.py`, `run_estate.py`, `deploy_estate.py` and
+# three others carry the same one) rather than a cosmetic split; the watchdog is nonetheless a
+# genuine seam and would be the thing to lift out if this file grows again.
+# pylint: disable=too-many-lines
 
 # The FILES this script writes under `--out`, probed as files on purpose. Two reasons:
 #   * a directory-only rule (`/_sweep*/`, or a hand-written `_myout/assets/`) is applied by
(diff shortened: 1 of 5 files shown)


Guust-Franssens:
## Blind review — 4 findings, NONE blocking

Reviewed SHA `c91ed598ef632ad0b69f300b22acbfe85c0488b6`. Reviewer was given the issue and diff only, and was told to treat the PR body''s rationale with suspicion rather than deference.

**Gates independently verified by exit code:** ruff **0** · pylint **0** (10.00/10) · watchdog tests **34 passed, 0** · four `test_harvest_*` files **237 passed, 0** · mutation-harness scoring tests **22 passed, 0** · `git status` clean.

**Confirmed working** — the join behaviour in leg 4 was attacked directly and held: an exact LUID beats a conflicting name, the case-insensitive fallback works, and a workbook that itself failed is correctly excluded. Feeding a deliberately non-compiling mutation to the harness produced `recorded=False` and no test outcome, i.e. it was **not** scored as caught.

### Being fixed in this PR (fail-closed, but they defeat its own purpose)

**1. High — the stall default is stricter than the fetcher''s own read timeout** (`:416`, `:696`). The 120s stall timeout sits **below** the engine''s 300s per-socket-read timeout, so a healthy transfer pausing 120–300s between chunks still satisfies urllib but is killed by the watchdog. Reproduction (one chunk, pause, continue; child read timeout 2s, scaled stall 0.25s):

```
BURSTY_HEALTHY  stalled None  0.49  True
```

The same probe correctly let a continuously-progressing 2.12s download survive a 0.35s wall clock and correctly killed a permanently stalled one — the defect is specifically the healthy gap between reads.

**2. High — losing the probe after movement kills a healthy download** (`:676-696`). Once movement sets `progress_observed=True`, a later `None` (probe unavailable) does not disarm the stall deadline, so access denial, a process exiting between enumeration and read, or an unreadable descendant reads as "no progress". Reproduction — probe `0`, `1`, then `None` forever against a healthy 30s sleeper:

```
PROBE_LOST_AFTER_PROGRESS  stalled 0.32  True
```

Both are being fixed here rather than filed, because #472 exists precisely because we killed a healthy download, and both of these can still kill a healthy download.

### Filed as #483 (diagnostic, ships anyway)

**3.** An indeterminate parser `ok` is counted as downloaded but enters neither bucket, so the new closure line can claim `1` against `0`.
**4.** Orphaned workbooks are a `set[str]` keyed on display name; Tableau permits duplicate names, so two distinct LUIDs collapse into one reported entry.

### Two corrections to the PR body

- The harness label is **`HARNESS-ERROR`**, not `INVALID`.
- **The 19-mutation campaign and its snippets are not in the committed diff**, so the reviewer could not verify that each compiled, hit its intended path, or moved only one arm of a relationship. Either commit the harness as durable evidence or state in the body that it was scaffolding.

### Not verified

No live Tableau server was available; the buffered urllib path was exercised with local slow-stream and stalled HTTP transfers instead.

Issue #478, opened 2026-09-03
package_unit.py: one failing unit aborts the whole batch - later units are never attempted and never reported


Guust-Franssens:
## Summary

One unit raising an exception aborts the **entire** `package_unit.py` batch. Every alphabetically-later unit is silently never attempted — not packaged, not skipped-with-a-note, not listed as failed.

Field evidence (SES, 47-asset estate, 2026-09-03): 29 units packaged, then `IA_Operation_Health_Summary_Dashboard` raised `shutil.Error: [WinError 3]`, and `IA_Policy_Change_Report` onward were never attempted. The operator got no record that they had been skipped.

## Ground truth

`scripts/package_unit.py:1107`:

```python
package_unit(bundle, unit, out_root, oracle_dir=oracle_dir, assets_dir=assets_dir) for unit in sorted(units)
```

A comprehension over `sorted(units)` with **no per-unit exception handling**. The first raise propagates and kills the run. `sorted()` also explains the observed alphabetical cut-off exactly.

## Why this is worth fixing separately from the path-length bug

It is a **resilience** defect, not a path defect. Whatever causes one unit to fail — a long path (#476), a malformed visual, a missing model, a locked file — the blast radius should be that unit. A 47-unit estate where unit 30 kills units 31-47 turns one bad workbook into a whole failed run, and the operator cannot tell "not packaged" from "packaged and fine" without diffing directories by hand.

It also compounds #476 badly: the operator''s only recovery is to enumerate every remaining unit by hand into repeatable `--unit` flags, because there is no `--exclude`.

## Proposed fix

1. Catch per unit, record `{unit, status, error}`, and continue.
2. Exit non-zero if any unit failed, but only **after** attempting all of them.
3. Report failures in the summary and in `--json`, so a partial run is machine-readable.
4. Consider `--exclude <unit>` (repeatable) as the natural complement to the existing `--unit`.

## Note on sequencing

`scripts/package_unit.py` currently has an in-flight fix round for #463. This is a new defect class, so per the scope-freeze convention in `AGENTS.md` it should not be bolted onto that PR. Queue behind it, alongside #476.

PR #484, opened 2026-09-03
fix(package): measure the path budget before assembling, and stop staging at the deepest point

### LINKED ISSUE #476: package_unit.py: WinError 206 - staging inside --out adds a path segment at the deepest point, with no long-path handling and no actionable error
## Summary

`package_unit.py` assembles each unit''s PBIR tree inside an **extra hidden staging segment under `--out`**, has **zero** long-path awareness, and does not catch `WinError 206`. On a live customer estate it crashed with a raw stack trace while packaging `Aircraft_Attributes_Dashboard`.

Reported from the field (SES, `tableauca.intelsat.com`, 47-asset estate), 2026-09-03:

> `package_unit.py` failed with `[WinError 206] The filename or extension is too long` while staging `Aircraft_Attributes_Dashboard`''s PBIR tree.

## Verified against `master`

Every claim below was checked, not taken on trust.

**1. The staging segment is pure path-length overhead, added at the deepest point** (`scripts/package_unit.py:900-906`):

```python
staging = out_root / f".{sanitize_staging_name(unit)}.staging"
shutil.rmtree(staging, ignore_errors=True)
result = _assemble_unit(bundle, unit, staging, oracle_dir=..., assets_dir=...)
replace_dir(staging, out_root / unit)
```

The whole tree is built under `<out>/.{unit}.staging/…` and only then moved to `<out>/<unit>/`. So every file is assembled at a path **longer than its final resting place**, by the length of `.{unit}.staging` minus `{unit}`. The atomic-replace design is right; the *location* of the staging dir is what costs.

**2. There is no long-path handling anywhere in the file.** Grep for `MAX_PATH`, `260`, `\\?\`, `long.?path`, `WinError`: **zero** matches. The only `OSError` handlers (`:194`, `:210`, `:497`, `:942`) cover JSON decoding and `rmtree`. Note `:950` already documents a `WinError 5` retry on a staging tree — so Windows filesystem quirks in staging were already known here, and 206 simply was not among them.

**3. Our own canonical `--out` is the deep one.** `work_dirs.py` and `docs/migration-phases.md` recommend `_runs/<NNN>-<slug>/packages/<batch>/`. Run number + slug + `packages` + batch name, then `.{unit}.staging`, then the PBIR shape `fabric/<Model>.Report/definition/pages/page-<generated-id>/visuals/<generated-id>/visual.json`. The reporter measured **256 characters for this unit before the filename**, against Windows'' 260 `MAX_PATH`.

**4. ⚠️ The part that makes this predictable rather than unlucky — we already measured it.** `docs/windows-path-limits.md` records, for a 45-workbook estate, **80 paths over 260 characters**. So this toolkit documented the hazard, measured that real estates blow it, and then recommended the deepest available location as canonical. `package_unit.py` is simply not covered by that doc.

**5. The technique is already in the repo.** `\\?\` is used in `check_path_ceiling.py:63` and `harvest_estate_assets.py:167`. `package_unit.py` just never adopted it.

## Fix, in preference order

1. **Stage outside `--out`** — assemble in a short temp root (`tempfile.mkdtemp()`), then move the finished tree to `<out>/<unit>/`. This removes the staging segment from the critical path entirely and makes the deep canonical `--out` safe again. Preferred: it fixes the cause rather than raising the ceiling.
2. **Catch `WinError 206` / `OSError` and raise an actionable error** naming the offending path, its length, and `docs/windows-path-limits.md`. A raw Win32 stack trace is not a usable failure.
3. `\\?\` prefixing as a belt-and-braces measure, matching the existing call sites.

Whatever lands must include a test that actually reaches the limit — a fixture whose staged path exceeds 260 characters, not a unit test asserting a helper returns a prefix.

## Operator workaround today (no patch needed)

Pass a **short `--out`** — `--out C:\pkg\<batch>` instead of the canonical `_runs/<NNN>-<slug>/packages/<batch>`. That reclaims 50+ characters immediately, and the package is self-contained, so it can be moved afterwards. Note `--out` must not sit directly beside a run''s `oracle/`/`assessment/` siblings (measured: `--out <run>/packages` exits 2, `--out <run>/packages/<batch>` exits 0), so a fresh short directory is the safe shape.

## Sequencing

`package_unit.py` is currently owned by the in-flight fix round for #463. This is a **new defect class**, not one of that PR''s seven findings, so per the scope-freeze convention in `AGENTS.md` it must not be bolted on. Queue behind #463 merging.

Guust-Franssens:
Fixes #476

> ⚠️ **Based on `feat/package-canonical` (tip `9144263`), not `master`.** PR #463 has ~520 lines of
> in-flight `package_unit.py` changes; branching from it keeps this diff mergeable. Retarget to
> `master` only after #463 lands.

## What was wrong

`package_unit.py` assembled each unit in `<out>/.{unit}.staging/` and only then renamed it to
`<out>/<unit>/`. The staging segment is inserted at the **deepest point**, so every one of the
hundreds of files in a PBIR tree was written `len(unit) + 9 - len(unit)` = **9 characters deeper than
its final home** — and the file had **zero** long-path awareness (no `MAX_PATH`, `260`, `\\?\` or
`WinError` anywhere in it). On a live 47-asset estate it threw `[WinError 206] The filename or
extension is too long` **having already packaged 29 units**, naming no path, no limit and no remedy.

## What this does — it SHORTENS paths; it does not raise the ceiling

Raising the ceiling was rejected, not overlooked. `docs/windows-path-limits.md` records Desktop
long-path support as **"no — nothing helps"**, measured with `LongPathsEnabled = 1` throughout, so
`\\?\` prefixing would convert a loud crash into a package **Desktop cannot open**. `\\?\` is also
specifically unsafe *here*: `conflicting_evidence_dirs()` **compares paths** (probing `out_root` and
`out_root.parent`), and a prefixed path may compare differently.

**1. The staging segment costs a constant instead of the unit name.** `staging_dir()` names it
`.` + 8 hex of `sha256(unit)` — 9 characters, whatever the unit is called. It stays a **sibling
inside `--out`**, deliberately against the issue's own `tempfile.mkdtemp()` suggestion: the swap is
`Path.rename`, atomic only *within one volume* and failing outright across two, and a temp root would
move assembly outside the directory `conflicting_evidence_dirs` has already cleared. **That guard is
untouched and sees exactly the directories it saw before.** `replace_dir`'s retired tree gets the
same treatment — `.{name}.replaced` made a package *being replaced* 10 characters longer than the one
anything had measured, and `shutil.rmtree` then walks it.

**2. The path budget is measured before anything is written.** `path_budget()` projects every
predictable path of every requested unit — at **both** the staged and the final root — against
`check_path_ceiling.py`'s ceilings, and `main()` refuses the whole run before `_package_each` is
called. The ceilings and the UTF-16 measurement are **imported, never re-declared**; `_utf16_len`
gained a public spelling (`utf16_len`) so a second module can reuse it without pinning a private name.

## Measured, on the reported field case

`IA_Operation_Health_Summary_Dashboard` at the canonical `--out` (99 characters):

| path | before | after |
|---|---:|---:|
| staged `…/visuals/<id>/visual.json` | **285** | **248** |
| final package (unchanged) | 276 | 276 |

**37 characters reclaimed** at assembly time. The final package is still 276 — so this unit genuinely
cannot ship at that `--out`, and it now says so up front instead of crashing 29 units later:

```
IA_Operation_Health_Summary_Dashboard: packaging it under C:\...\020-enterprise-dashboards\packages\batch1
would produce 2 path(s) Power BI Desktop refuses to open, so it was not assembled at all.
  deepest: 264 UTF-16 units, 17 over the 247-character directory ceiling
    C:\...\batch1\IA_Operation_Health_Summary_Dashboard\fabric\IA_Operation_Health_Summary_Dashboard.Report\definition\pages\page-ws-CA-Summadc2995b2\visuals\slicer-page-ws-Cb3f14546
--out is 99 character(s) long; it must be at most 82 (17 shorter) for this unit to fit.
Ceilings are measured in scripts/check_path_ceiling.py; background in docs/windows-path-limits.md.
```

⚠️ Note the binding rule is the **directory** ceiling (247), 12 tighter than the file ceiling, exactly
as `check_path_ceiling.py` warns. And the reported operator workaround is confirmed by the same
arithmetic: `--out C:\pkg\ent1\batch1` (18 characters) leaves budget **82** and **zero** overruns.

## Corrections to the brief, made by measurement

* **"Staging is never deeper than the final destination" is false below ~9 characters.** A unit named
  `B` stages under a 9-character hidden name, so it is 8 characters *deeper* than `<out>/B` — while
  still strictly shallower than the pre-fix `<out>/.B.staging` (10). What holds for **every** name is
  that the overhead is a constant instead of `len(unit) + 9`. This is why the budget measures **both**
  roots rather than assuming the final one is deeper, and why
  `test_the_budget_measures_the_STAGED_tree_too_not_only_the_final_one` exists.
* **`conflicting_evidence_dirs()` tolerates the move with no change at all**, because staging stays
  inside `out_root`. Nothing about that guard is touched.
* This PR fixes the **destination** side only. The issue's addendum documents an independent
  **source**-side overrun inside `_runs/<NNN>-<slug>/bundle/pbip/<Unit>/<Unit>.Report/…` (265
  characters before `--out` is consulted), whose lever is the engine repeating the unit name. That is
  upstream-owned and out of scope here.

## Gates (exit codes)

| gate | exit |
|---|---|
| `ruff format` + `ruff check --fix` on changed files | **0** |
| `ruff format --check conftest.py scripts tests .github/skills` (CI shape) | **0** |
| `pylint scripts\package_unit.py` | **0** |
| `pylint scripts\check_path_ceiling.py` | **0** |
| `pytest tests\test_package_unit.py tests\test_package_unit_gates.py -q` | **0** — 104 passed, 1 skipped (was 91/1) |
| `pytest tests\test_check_path_ceiling.py -q` | **0** — 62 passed, 2 skipped |
| `pytest tests\test_repo_layout.py tests\test_no_conflict_markers.py -q` | **0** |

Full suite deliberately **not** run — batch policy is one run at the end, after reviews settle.

## Mutation campaign — 9 new, all CAUGHT

Run through the committed harness (`tests/mutate_package_unit.py`), each against the single test it
must kill; a collection/import error scores **BROKEN**, never caught.

| mutation | verdict | anchors failed |
|---|---|---|
| staging: pre-fix `.{unit}.staging` | CAUGHT | **3** |
| staging: one shared `.staging` for every unit | CAUGHT | 1 |
| `replace_dir`: retired tree named `.{name}.replaced` | CAUGHT | 1 |
| budget: package anyway (reproduces WinError 206) | CAUGHT | **2** |
| budget: measure only the final root | CAUGHT | **2** |
| budget: drop the batch preflight in `main` | CAUGHT | **2** |
| message: drop the length / ceiling / overage | CAUGHT | 1 |
| message: reclaim figure that does not add up | CAUGHT | 1 |
| message: offer a negative `--out` length | CAUGHT | 1 |

`test_nothing_is_assembled_deeper_than_the_pre_fix_staging_tree` asserts `<`, not `<=`: under the
pre-fix mutation the two names are **equal**, so `<=` would have been satisfied by the very defect it
exists to catch.

⚠️ **Four campaign entries fail on this branch and none of them is mine — measured, not assumed.**
Restoring `scripts/package_unit.py` to base commit `9144263` and re-running reproduces both classes
identically: three `ANCHOR-SNIPPET x0` entries name code `feat/package-canonical` has since
refactored (`candidate = f"{digest[:width]}/{name}"`, `if not value.startswith("/"):`), and
`README: drop the data/ row` **SURVIVES** there too. They belong to PR #463 and are reported here
rather than fixed. The README's mutation count was also stale (25 stated, 45 present); it now states
the measured 54.

## What I could NOT verify

* **No end-to-end run against a real 260+ character tree.** Every test measures the destination side
  as arithmetic on strings that are never created, because neither CI runner can be assumed to write
  such a path — which is precisely how this defect reached a customer. The *arithmetic* is pinned at
  the real 259/247 boundary; the *filesystem behaviour* at that boundary is not re-measured here.
* **Oracle renders and localized `data/` copies are not projected.** Both depend on identity
  resolution that has not run at preflight time. They are shallower than the PBIR tree in every
  measured estate, and omitting them costs *protection* (a path can still be hit at assembly time),
  never a false refusal — stated in `_package_tails`' docstring rather than left implicit.
* **The batch refusal is all-or-nothing by choice.** A shorter `--out` moves every unit, so packaging
  the subset that fits would be redone wholesale anyway. If reviewers prefer per-unit skipping, that
  is a one-line change — but it interacts with #478 (`C:\tfmig\wt478`), which owns the per-unit
  exception handling in `_package_each`. **This PR deliberately does not touch that loop**; the new
  `PackagePathTooLong` is raised from `package_unit()` and the batch refusal sits in `main()` *before*
  the loop, so the two changes should merge mechanically.


Guust-Franssens:
## Blind review at `02e901e9` — REQUEST CHANGES (3 blocking)

### Headline: what this actually buys

For a unit name of length `N`, the staging segment goes `.{unit}.staging` (`N+9`) → `.<8 hex>` (`9`), so the headroom gained is **`min(N, 9)` — nine characters**, plus up to 10 more on repackaging (`.{unit}.replaced` → `.<8 hex>~`).

Measured on `Aircraft_Attributes_Dashboard` (N=29) with a field-shaped PBIR path:

| path | length |
|---|---|
| old staged deepest file | 268 ❌ |
| **new staged deepest file** | **239** ✅ |
| final deepest file | 259 (ceiling 259, refuses at 260) |
| final deepest directory | 247 (ceiling 247, refuses at 248) |

So the reported crash case now packages. **But a unit still over budget exits 2 with "use a shorter `--out`"** — an actionable refusal, not automatic long-path support.

The mutation coverage is genuinely good: **10 of 11 mutations caught** with tight blast radius, including restoring the long staging name (3 tests), removing the pre-write refusal (2), and measuring only the final tree (2). No test deleted or weakened.

---

### B1 — the projection is incomplete, so exit 0 can still ship an over-ceiling package

`_package_tails()` says outright it is not exhaustive. It omits **assets, localized data, oracle outputs, newly generated `expressions.tmdl`, and the retired tree** touched during replacement (`package_unit.py:2199-2220`, `:890-892`, `:1922-1924`, `:2663-2664`).

```
valid 204-char .twb filename
  projected maximum = 102, zero overruns
  packaged asset path = 279
  main_exit = 0, asset_written = True
```

This is the original defect surviving: it can still raise WinError 206, or ship a package Desktop refuses to open.

| output | covered by the budget? |
|---|---|
| `pbip/<unit>/**` under `fabric/`, root scaffold, `handover/<unit>.json` | ✅ |
| `assets/<original-name>` · `data/**` · `oracle/**` · generated `expressions.tmdl` · retired `.<digest>~/**` | ❌ |

### B2 — the staging name can silently DELETE an already-packaged unit

Unit names beginning with `.` are allowed, so a staging digest can collide with a real package name. `rmtree(staging)` then removes whatever sits there (`staging_dir()` ~`:2186`, deletion `:2365-2367`).

```
units: Victim  and  .d72cee2e     (.d72cee2e is Victim's staging name)
CLI: both reported packaged, exit 0
actual output: only Victim  -> the first package was deleted, silently
```

Data loss reported as success. The "use a shared `.staging`" mutation does **not** catch this.

Related, non-blocking: a genuine 32-bit stem collision exists — `Unit_17592` and `Unit_58987` both map to `.aff484f3` after 58,988 names. Serialized packaging tolerates that pair, but the stem is not collision-proof.

### B3 — Windows ceilings applied to absolute POSIX paths

On Linux: POSIX `--out` of length 297 → portable root budget 207 → **exit 2**. Bypassing only the budget produced a valid package, exit 0, deepest path 332. Packages are relocatable, so this refuses a valid POSIX build that would be fine under a short Windows root later.

### Non-blocking

- **The directory ceiling is unpinned.** Mutating `DIR_CEILING` 247 → 260 left the suite green (148 passed). Production uses the correct distinct ceilings; a regression would escape.
- One over-budget unit stops the whole batch (exit 2, empty output). Deliberate all-or-nothing preflight, and arguably right since all units share one `--out`; continue-on-failure is #487's job.

### Gates

pytest **0** (148 passed, 1 skipped) · ruff format/check **0/0** · pylint **0** · `test_check_path_ceiling.py` **0** (62 passed) · mutation harness **1** (51/54; the 3 `ANCHOR-SNIPPET x0` are the known pre-existing dead anchors).

Unmeasurable projection and digest failures did **not** fail open — they raised `OSError` / `UnicodeEncodeError` with nothing written. On a clean refusal, `package_unit()` creates nothing; the CLI leaves an empty `--out` root because it `mkdir`s before preflight, but no staging or half-package.

### Not verified

The real SES/Intelsat 47-asset bundle and its 183 paths were unavailable; Desktop was not launched to reopen a boundary package; the original WinError 206 was not reproduced — this host wrote the unmeasured 279-char path successfully, which is how the fail-open surfaced instead.

Guust-Franssens:
@copilot The re-review closed projection completeness and staging data-loss. One same-class portability bug remains; make only this bounded correction.

Current defect:
- `_budget()` computes both host `overruns` and Windows relocation `shipping`/`shipping_budget` from all three `package_roots()` (final, staging, retired).
- Staging and retired roots are real host paths and must remain subject to host limits, but they do not ship.
- With a short unit name and a legal long final tail, the final package fits a Windows root exactly while `.digest~` makes `shipping_budget=-6`, so a valid relocatable package is refused solely by transient scratch length.

Required contract:
1. Keep host `overruns` and `out_root_budget` calculated from final + staging + retired roots. Actual assembly/deletion must still refuse if any host scratch path is too long.
2. Calculate Windows portability `shipping` and `shipping_budget` from the final package root only, in both preflight `path_budget()` and post-assembly `assembled_budget()` paths.
3. The package JSON `shipping_root_budget` must therefore describe the tree that ships, not scratch directories.

Direct regression:
- Use a short unit such as `B` and a tail where the final package-relative file cost is within 259 but staging/retired cost is over it under `platform_limits("posix")`.
- Assert there is no Windows shipping refusal and `shipping_budget` reflects the final root.
- Add a negative control proving an actual host-limit overrun in staging/retired still refuses.

Hard scope:
- Change only `scripts/package_unit.py` and `tests/test_package_unit.py`.
- Do not change scratch names, path ceilings, replacement behavior, projection coverage, batch continuation, docs, or mutation infrastructure.
- Run the Python edit ritual, pylint by exit code, focused path-budget/package tests, and existing mutation tests that cover B3 if available.

Commit and push. The same reviewer will confirm only this portability reproduction and host-limit preservation.

PR #487, opened 2026-09-03
fix(package): one failing unit no longer aborts the batch - attempt every unit, report every one (#478)

### LINKED ISSUE #478: package_unit.py: one failing unit aborts the whole batch - later units are never attempted and never reported

Guust-Franssens:
Fixes #478

Batch packaging now attempts and accounts for every requested unit:

- all path budgets are measured before assembly, but unit-specific over-budget/unassessable outcomes no longer abort valid siblings;
- each requested unit lands exactly once in packaged, failed, or kept/refused accounting; gaps are non-clean;
- unknown exceptions and path-budget failures preserve sanitized exception/frame context;
- any diagnostic containing a host location is redacted as a whole message/line, avoiding account/path leakage without a brittle path grammar;
- operator interrupts remain terminal and failed units leave no partial package.


Guust-Franssens:
---

# Round 2 — blind-review High finding fixed (`63debe3`, `9233309`)

**Finding accepted in full, including its framing:** the per-unit continuation is what made a latent
shared-staging defect *reachable*. Fail-fast was accidentally masking it — the contaminated later
package was never produced, because the batch stopped first. Silent cross-unit contamination reported
as success is worse than the crash it replaces, so this blocks correctly.

**The naming half was left alone**, as instructed: PR #484 (`wt476`, same base) hashes the whole unit
name into a constant-width stem, and re-implementing that here would conflict in the file we
partitioned on purpose. `sanitize_staging_name` is untouched.

## What changed (cleanup only)

- **`_clear_staging(staging) -> str | None`** replaces `shutil.rmtree(..., ignore_errors=True)`.
  Bounded retries (3 × 0.1/0.2/0.3 s — the same Windows-handle race `_SWAP_ATTEMPTS` already exists
  for), the verdict taken from `exists()`, and the `OSError` text carried **out** as the reason
  instead of discarded.
- **`package_unit` refuses to assemble into a staging directory that survived cleanup.** That is the
  invariant, and it holds regardless of how the name is derived — so it still bites after #484, where
  it narrows to the same unit re-run into a stale directory. The unit lands in `failed[]`; the rest
  of the estate still runs.
- **A residue that outlives a *successful* build fails that unit too**, with an error that says the
  package itself is complete on disk.

## One deliberate deviation, with evidence

> *"Retry, then surface the failure"*

Surfaced — but **not by raising when an exception is already in flight.** An exception raised out of
a `finally` **replaces** the propagating one, so raising there would report *"staging survived"* for a
unit that actually failed on `shutil.Error`, and the root cause — the only actionable half — would be
gone. `sys.exception()` (3.11+, verified empirically before use) separates the two paths:

| path | what happens |
|---|---|
| build raised, residue left | residue recorded on that failure via `add_note()` (so it reaches `failed[].traceback`) + a stderr `WARN`; the **root cause is what `failed[].error` says** |
| build succeeded, residue left | `PackagingError` → the unit is reported failed, as asked |

The consequence of not raising on the first path is zero, because the pre-assembly refusal is what
closes the contamination — this is about *which* message the operator gets, not whether the residue
is acted on. `test_a_cleanup_failure_never_replaces_the_error_that_caused_it` pins it, and the
mutation that raises over the in-flight error is caught.

## Tests

- `test_a_stale_staging_directory_is_refused_rather_than_inherited` — the reviewer's shape: a unit
  fails leaving a populated staging tree (removal blocked), the next build on that path must
  **refuse**, and the package must not contain `FROM_THE_FAILED_UNIT.txt`. ⚠️ It takes the staging
  path **from what the packager hands the assembler**, never recomputing it, so it keeps testing the
  invariant after #484 changes the naming. It re-runs the *same* unit rather than a 60-char-prefix
  sibling for the same reason: a test built on the collision would go green-but-vacuous — or fail
  outright — the moment #484 removes the collision.
- `test_a_cleanup_failure_never_replaces_the_error_that_caused_it`
- `test_a_staging_directory_that_outlives_a_SUCCESSFUL_build_fails_that_unit` — needs a copying-swap
  stub, named as load-bearing in the test: `replace_dir` **renames**, so today this guard is
  unreachable through the normal path and exists for a swap that copies or a rename that
  half-succeeds.

## Mutations added (hit counts on unmutated source)

| mutation | verdict | site hits |
|---|---|---|
| assemble into a stale staging directory instead of refusing | CAUGHT (1 failed) | 4 |
| back to `rmtree(ignore_errors=True)` — a failed removal called a success | CAUGHT (2 failed) | 7 |
| raise the cleanup failure over the in-flight error | CAUGHT (1 failed) | 3 |
| let a staging dir outlive a SUCCESSFUL build unreported | CAUGHT (1 failed) | 1 |

## Gates (exit code)

| gate | exit |
|---|---|
| `ruff format --check` + `ruff check` (3 files) | **0 / 0** |
| `pylint scripts\package_unit.py` | **0** (10.00/10) |
| `pytest tests\test_package_unit.py tests\test_package_unit_gates.py -q` | **0** — **101 passed, 1 skipped** (was 98; baseline 91) |
| `python tests\mutate_package_unit.py --only batch` | **0** — 9/9, control survived |
| full campaign (informational) | **54/58**, both controls survived; the same 4 non-batch entries, reproduced as pre-existing in base `9144263` |

Full suite not run, per batch policy.

Guust-Franssens:
# CI fix — the failure came from the shared BASE, not from this branch

Both failing tests were already fixed on `feat/package-canonical`; this branch was simply **stale
against its own base**. Merged `c41994b` in (`7d8f0b0`).

`fix/package-batch-478` and `fix/package-longpath-476` (#484) both forked this base at `9144263`.
The base has since advanced five commits, two of which are the fixes:

| base commit | fixes |
|---|---|
| `2721a90` compose the `DataFolder` parameter with the base path's own separator | `test_the_data_folder_parameter_names_the_FINAL_package_not_its_staging_dir` — a literal `\` produced `/tmp/.../out/Book\data\` on Linux |
| `42621f1` stop reading `check_unit`'s `NOT_CHECKED` verdict as a usage error | `test_every_command_the_readme_prints_produces_a_verdict_not_a_usage_error` — exit 2 is argparse usage in one gate and a genuine VERDICT in the other |

That is why #484 fails with the identical pair: same base, same staleness. Taking the fix from the
base is also what keeps the two PRs from colliding — re-implementing either here would have written
the same repair twice into `scripts/package_unit.py`.

⚠️ Note only the **first** was a separator defect. The second is independent (a verdict/usage
conflation) and would not have been found by looking for hard-coded `\`.

## One behaviour change a reviewer must see: exit 4 → 5

The base claimed **exit 4** for a different meaning — `EXIT_NOT_SELF_CONTAINED`, *"the package exists
but ships without a source its model names"* — while this branch used 4 for *"a unit RAISED, or was
unaccounted for"*. Folding them together would leave a caller unable to tell **"packaged, some rows
missing"** from **"never packaged"**, which is the exact distinction #478 exists to restore. So:

```python
EXIT_UNIT_FAILED = 5   # returned BEFORE 3 and 4
```

Ranked highest deliberately: 3 and 4 describe a package that **exists**; 5 describes a unit about
which **nothing is known**, and an incomplete package must never mask a unit that was never assessed.
The fail-open shape stays impossible — `partition_gaps` still measures the buckets against the
**request**, and a gap still exits non-zero. Mutation `batch: exit 0 despite a failed unit` is still
**CAUGHT (2 failed)** after the renumber.

Renumbered with it: the module-docstring exit table, `_package_each`'s docstring, the
`scripts/README.md` row, the six #478 assertions (now `pkg.EXIT_UNIT_FAILED`, not a bare literal) and
the batch mutation in `tests/mutate_package_unit.py`.

## Other conflicts — resolved as a union of two independent guarantees

- `package_unit` keeps **both** the staging-survival refusal (this branch) and the under-the-lock
  digest re-check (the base);
- `_clear_staging` sits beside `replace_dir(..., verify=)`;
- `render` emits the `UNACCOUNTED` line **and** the `NOT SELF-CONTAINED` / `UNBOUND` lines;
- imports are the union minus `PureWindowsPath`, which the base moved into `scripts/path_flavour.py`.

Nothing from the base was dropped: every line the merge removes from `origin/feat/package-canonical`
is one this branch deliberately replaces (old `render`, old `main`, `rmtree(ignore_errors=True)`).

## One test needed a real change, not a renumber

`copying_swap`, the `replace_dir` stub in
`test_a_staging_directory_that_outlives_a_SUCCESSFUL_build_fails_that_unit`, was written against the
old two-argument signature; the base added `verify`. It now mirrors the full signature **and calls
the callback** — a stub that silently dropped it would disable the #460 edit re-check while still
looking like a swap.

## Gates (exit code)

| gate | exit |
|---|---|
| `pytest tests/test_package_unit.py tests/test_package_unit_gates.py -q` | **0** — 113 passed, 1 skipped (101 + 1 before the merge) |
| `uv run ruff format --check` (packager + both changed test modules) | **0** |
| `uv run ruff check` (same) | **0** |
| `uv run pylint scripts/package_unit.py` | **0** (10.00/10) |
| `python tests/mutate_package_unit.py --only batch` | **0** — 9/9 caught, negative control survived |
| **ubuntu CI, `7d8f0b0`** | **SUCCESS** — 4653 passed, 89 skipped, 10 deselected (was 2 failed / 4638 passed) |

Controlled experiment on Windows, driving the composers with POSIX-shaped input — the CI failure
string was `Book\data\` inside a POSIX path:

```
_package_data_folder(PurePosixPath('/tmp/.../out/Book'))    -> '<PACKAGE_ROOT>/data/'
_package_data_folder(PureWindowsPath(r'C:\runs\out\Book'))  -> '<PACKAGE_ROOT>\data\'
_path_separator('C:/runs/out')                              -> '\'   # flavour, not "contains a backslash"
```

## What I could not verify

- **Linux behaviour locally** — Windows-only box. Both underlying fixes were reproduced and verified
  under WSL by their own base commits; my evidence for this merge is the green ubuntu CI run above
  plus the POSIX-shaped probe.
- **The renamed test.** `test_the_data_folder_parameter_names_the_FINAL_package_not_its_staging_dir`
  no longer exists — the base replaced it (`a18c5a6`) with
  `test_the_data_folder_parameter_names_a_PLACEHOLDER_not_the_machine_that_built_it` plus a
  parametrised `..._never_mixes_separators_on_EITHER_platform`, because the value is now
  `<PACKAGE_ROOT>` and binding is a step of consuming the package. That is the base's reviewed design
  decision, inherited here, not a test weakened to go green.
- **A full local suite run is not clean, and that is pre-existing**: 6 failures in
  `test_issue_424_chart_type_pin.py` / `test_upstream_repro_pins.py` are engine-version pins
  (installed engine **2.365.0** vs the branch's pinned **2.260.0**; master retired these in
  `b0f8726`), and 2 in `test_gui_marker_gate.py` appeared only under my `-n 8` xdist invocation and
  pass sequentially. None reference `package_unit`, and CI runs all of them green.

Guust-Franssens:
Deferred, not abandoned. This branch's own diff no longer applies onto master: #463 was **squash-merged**, so its individual commits here do not match master's single squashed commit, and #463's round-2 fixes touch the same regions of scripts/package_unit.py and scripts/README.md that this PR changes. That is a semantic conflict to resolve once, not a mechanical rebase.

Sequencing it **after #484**, which is being fixed right now in the same file. Rebasing both concurrently would create a three-way collision in package_unit.py.

Recipe that worked for #484, for whoever picks this up: branch from master, apply this branch's own diff (from its merge-base with \eat/package-canonical\, not a two-dot diff), resolve against #463's round-2 changes, verify, then force-push.

Note the exit-code contract already agreed: \5\ = EXIT_UNIT_FAILED here matches #463's, deliberately, so a unit that failed is distinguishable from one that packaged incompletely (4) or could not be assessed (6).

Guust-Franssens:
@copilot The final review approved all #478 batch/cleanup/exit behavior and found one blocking privacy leak. Fix only crash diagnostic sanitization.

Current leak:
- `UnitCrashed.traceback = traceback.format_exception(...)` stores absolute source paths and raw exception text.
- `reason` also embeds raw `_one_line(error)`.
- A neutral `C:\Users\NeutralCanary\...\secret.csv` appears in stderr, `failed[].reason`, and `failed[].traceback`.

Required behavior:
- Preserve the module-qualified exception type (`_error_label`) and actionable frame/function/line information.
- Never serialize raw absolute/UNC/POSIX host locations or arbitrary private path text from exception messages, traceback filenames, source snippets, reason, JSON, stderr or summary.
- Reuse the existing `host_paths.discloses_host_location` / package containment semantics; do not add a second path detector.
- For traceback frames, emit safe relative/basename frame identity plus line/function, not absolute filenames. Sanitize source-line and exception-message strings through existing containment.
- Ordinary non-path exception text remains readable.
- Keep KeyboardInterrupt/SystemExit propagation, batch continuation, staging cleanup, outcome accounting and exit codes unchanged.

Add direct neutral-canary tests covering Windows drive, UNC and POSIX absolute paths in both exception messages and frame filenames, plus a non-path control. Change only `scripts/package_unit.py` and focused tests/mutation entry if required. No new redaction module/framework.

Run ruff, pylint by exit code, focused package tests and #478 mutation selection. Commit and force-push normally to the existing PR branch (no history rewrite beyond the current one-commit rebuild). The same reviewer will confirm only this privacy finding.

Guust-Franssens:
@copilot Scope-frozen correction required at current tip `9c9520e`; keep this PR on #478 only.

Two high-severity blockers were independently reproduced:

1. **Path budgeting still sits outside the per-unit boundary.** `_refuse_out_too_deep()` comprehends every unit before `_package_each`; a unit-specific `path_budget()` exception escapes after visiting only the prefix, and an over-budget unit calls `parser.error` for the whole run. Valid sibling units are never packaged and no per-unit JSON accounting is written.

Preserve #476's valuable rule that all requested units are measured before any assembly, but measure them **individually into outcomes**: precompute each budget under a per-unit exception boundary, record over-budget/unassessable units in the existing failed/cannot-assess accounting, and pass only safe units to packaging. Every original request must still land exactly once in `units[]`, `failed[]`, or `refused[]`; no global argparse exit for one unit-specific budget. A root that makes every unit fail may write no packages, but must still emit complete accounting and the correct non-clean exit.

2. **Diagnostic redaction leaks the tail of a spaced account name.** Current output is:

```text
C:\Users\Neutral Canary        -> [host location redacted] Canary
\\server\Users\Neutral Canary -> [host location redacted] Canary
/home/Neutral Canary           -> [host location redacted] Canary
'C:\Users\Neutral Canary'      -> [host location redacted] Canary'
```

Redact complete detected host-location spans, including a path that ends at a quoted or punctuated account name containing spaces. Keep ordinary non-path text and actionable traceback basename/line/function. Cover Windows, UNC, POSIX home, trailing punctuation/quotes, embedded path plus following ordinary prose, and false-positive controls.

Add direct tests and narrow mutations for these two production paths only. Do not add `--exclude`, redesign staging, touch #516, or create a new sanitizer framework. Run Ruff format/check, targeted package tests, mutation rows, and the privacy gate; push the corrected tip to this PR.

Guust-Franssens:
@copilot Stop-rule simplification for #487 at `6524a62`; do not extend the path-span heuristic again.

The latest sanitizer still leaks lowercase account suffixes and multiword filenames, and can consume capitalized ordinary prose:

```text
C:\Users\neutral canary              -> [host location redacted] canary
C:\Users\Neutral Canary\My Secret File.csv -> [host location redacted] Secret File.csv
C:\Users\Alice Failed again         -> [host location redacted] again
```

**Delete/simplify the span-continuation heuristic.** If any diagnostic message or source-line text contains a host location according to the existing host-location detector, replace that entire message/line with one fixed `[host location redacted]` token. If no host location is detected, preserve the ordinary text unchanged. Continue retaining exception class, traceback frame basename, line number, and function. This fail-closed whole-message redaction is deliberately less informative than parsing arbitrary paths with spaces and ends the repeated leak/prose-boundary class.

Also preserve actionable context for unexpected `path_budget` exceptions while keeping exit 6: include the sanitized module-qualified exception label and a safe traceback on the `UnassessableInput` outcome instead of dropping class/frame context to `traceback: null`.

Add direct controls for lowercase spaced accounts, multiword filenames, quoted/punctuated paths, capitalized/lowercase following prose (whole message redacted), ordinary non-path errors (preserved), and path-budget ValueError class/frame with exit 6/accounting intact. Do not add another path grammar, touch #516, or change per-unit accounting. Push for one final same-reviewer confirmation.

PR #523, opened 2026-09-04
Make the manual single-workbook path work end to end (bring your own .twb + screenshots)

### LINKED ISSUE #519: Support the manual single-workbook path: bring your own .twb + screenshots, no Tableau server (customer ask)
## Who and what

From an internal call on the migration toolkit, 4 Sep 2026 (organized by Amol). A reporting team — Arpita and Anthony — supports a small number of Tableau dashboards and wants to try the toolkit on **one dashboard**, providing the inputs by hand. Kwame is covering the same ground to share load across time zones.

Arpita, verbatim:

> "I think I'll start and Anthony will start with like **1 dashboard**. We gonna **provide everything**. We gonna give the screenshot. Like **we don't want to go that route of like, oh, connecting with the whole server**."

Amol restated the flow they expect:

> "we clone the repo ... inside of that folder ... **let's say I put my Tableau workbook in there**. And then in GitHub Copilot, I can say, hey, you have this tableau dashboard ... convert it over to Power BI ... **the screenshots will probably have to point them to them as well**."

## Why this is a gap

Acknowledged on the call:

> "more recently now I've been making it for **bulk migration** ... because now it's kind of made for that bulk kind of migration, but **I will definitely see together with you how we can then have it for one at a time** kind of thing."

The estate path (survey -> assess -> harvest -> `run_estate.py` -> `package_unit.py`) is where the recent work has gone. The single-workbook path exists but is agent-facing and undocumented for a first-time user.

## What already works (verified)

- **`AGENTS.md` documents the shape**: one `.twb`/`.twbx` -> `parse_tableau.py <file> -o <spec>` -> dispatch `@tableau-migrator`.
- **The manual screenshot provider is implemented**: `capture_tableau_reference.py::collect_manual()` adopts user-dropped `tableau-*.png` from `reference/`, validating PNG dimensions, and `--manual-validation-grade` makes a validation-grade claim explicit and attributable.

So the pieces exist. What is missing is the **path from "I have a .twb and some PNGs" to a running migration**, written for someone who has never run this.

## Proposed scope

1. **A single "start with one workbook" doc** — clone/fork, where to put the `.twb`/`.twbx`, where to put screenshots and **what to name them** (`tableau-*.png` in `reference/`), then the one command to start.
2. **Verify the one-at-a-time path end to end** on a workbook plus hand-dropped screenshots, with no Tableau server configured. This was an explicit commitment on the call.
3. ⚠️ **Check a silent-drop risk**: `collect_manual()` filters to `st_size > 20000`. A simple single-page dashboard screenshot can plausibly be under 20 KB, and would then be **silently ignored** — the user would see "no reference" without being told their file was rejected for size. Either lower/justify the threshold or report rejected files by name and reason. (Unassessable input reaching a quiet outcome is this repo's dominant defect class.)

## Other commitments and constraints stated on the call

- **A working session next week** — a "hello world" on one simple report, Amol leading, with Kwame and me available if it hits a roadblock.
- **Pre-existing data-source connections are a hard prerequisite.** An agent cannot see or set Power BI Desktop's saved connections, so the source must already be connected in Desktop or the agent burns time. Worth stating in the doc.
- **Expectations set**: ~1.5 h minimum for a simple report; roughly $50-100 for a medium-complexity report, more for complex ones; not a one-shot bulk conversion.
- Arpita and Anthony already have the Fabric skills, the Desktop bridge and the report-author CLI installed. The extra dependency is **.NET** for the tabular object model (used for refresh with live row counts, falling back to MCP).
- Feedback route agreed: direct message, **or file an issue here**.

## Known limitations that were called out, for the doc

- Desktop error dialogs still need a human to copy/paste the error to the agent (UI automation is the only alternative and is not recommended).
- Power Query has no validation path equivalent to TMDL, so incorrect generated M is not self-detectable yet.

Guust-Franssens:
## What this does

Makes the **manual single-workbook path** work end to end and documents it: bring your own
`.twb`/`.twbx` plus screenshots, no Tableau Server connection.

Fixes #519

## Verified end to end, with no server configured

No `TABLEAU_SERVER_URL`, no PAT. Real workbook (`fixtures/upstream-repros/issue-424-.../a`), a
hand-placed screenshot. All four steps exit **0**:

```
python scripts/parse_tableau.py migrations/workbooks/<slug>/source/<workbook>.twb -o migrations/workbooks/<slug>/migration-spec.json
python scripts/capture_tableau_reference.py migrations/workbooks/<slug>
python scripts/run_estate.py --input migrations/workbooks/<slug>/source --output _runs/001-<slug>/bundle
python scripts/check_reference_readiness.py _runs/001-<slug>/bundle --source migrations/workbooks/<slug>/source/<workbook>.twb --reference migrations/workbooks/<slug>/reference
```

Final entry-gate verdict: `READY - 1/1 expected page(s) ready`.

## What was broken

Three defects, all of the *"rejected input reaches a quiet outcome"* class, all of which fire on the
**first** thing a first-time user tries.

**1. Screenshots were silently discarded — then the tool asked for them.**
`collect_manual` filtered on `st_size > 20000`. PNG is lossless, so a legible 1440×900 flat-fill
dashboard measures ~6–7 KB (measured 7,154 and 6,053 bytes on two mocks). It was dropped with no
artifact, no log line and no exit-code difference, and the fail-closed error then said:

```
ERROR No reference image could be produced ...
  * Manual: drop tableau-<name>.png screenshots into <reference dir>
```

…naming the directory that already held the file it had just thrown away.

**2. An embedded thumbnail silently outranked the operator's own screenshot.**
The manual leg ran last and only `if not records`. A workbook carrying embedded thumbnails — 17/17 in
one measured estate, i.e. the normal case for anything saved out of Tableau Desktop — produced 192×192
layout-only records while the full-page dashboard PNG sat unread beside them, and
`--manual-validation-grade` silently did nothing.

**3. Found during verification: a manual record could never satisfy a page.**
Even a correctly named, sha256-attributed, `--manual-validation-grade` screenshot returned
`UNVERIFIABLE - name only; scope unknown cannot satisfy a dashboard page`. `reference_evidence` gives
the `manual` provider `KIND_UNKNOWN` and honours an explicit `view_type` on the manifest entry
instead — and nothing ever wrote one, so the gate's own hint asked for a field **no flag could
produce**. `docs/reference-readiness.md` still claimed validation grade widened scope; round 3 had
removed that, so the documented route was doubly unwalkable.

## The `collect_manual` decision

The 20 KB floor is **not** defensible for a hand-supplied file, and it is **kept** for a generated
one:

- `_is_valid_png` (Playwright output) keeps `> 20000`, because there size really is a proxy for "did
  we capture a viz or an error/consent page". Its docstring now says so, and says not to reuse it.
- The manual floor moves to **pixels**, reusing `reference_evidence.MIN_RENDER_EDGE` (64 px) — the
  same constant the readiness gate judges the same image by, so a file adopted here can never be
  rejected as illegible there. Justified in the existing code against the 192×192 embedded thumbnail.
- Every candidate is now **accounted for**: adopted, or named with the reason — wrong prefix, wrong
  extension, not a PNG, empty, below the px floor — in the warning stream *and* again in the
  fail-closed error (`N file(s) already in <dir> were REJECTED, not missing:`). Our own
  `powerbi-*.png` and `manifest.json` are excluded, so they are never reported as failed candidates.
- Scanning is casefolded rather than `glob("tableau-*.png")`. ✅ Measured on Windows: the glob already
  matched `Tableau-Page.PNG`. ⚠️ On POSIX `pathlib` matching is documented as case-sensitive, so the
  same file would have been invisible there — that half is documented, not measured here.

`view_type` is **derived, never guessed**: from `migration-spec.json`, joined on the same casefolded
exact name the gate uses (**not** slugified — so `tableau-sales-overview.png` will *not* match a
dashboard called `Sales Overview`). A name claimed by two kinds is dropped rather than resolved, and
`--manual-object-type {dashboard,worksheet}` is the explicit fallback. It asserts KIND only; grade
still never widens scope.

## Docs

`docs/start-with-one-workbook.md` — short and command-first: prerequisites (including the **Power BI
Desktop saved-connection** one that costs an hour when missed), where the workbook and screenshots
go, the exact `tableau-<exact object name>.png` naming contract, the verified sequence, how to hand
off to the migrator agent, honest time/cost expectations, and the two known limitations (a Desktop
error dialog still needs a human; Power Query M has no TMDL-equivalent validation path). Linked from
`docs/INDEX.md` and from `AGENTS.md`'s single-workbook input-shape row.

## Gates — by exit code

| gate | exit |
|---|---|
| `pytest tests/test_capture_tableau_reference.py -q` | 0 |
| targeted suites (capture + manual-grade + readiness + synthetic) | 0 (92 passed) |
| **full suite** `uv run pytest -q` | 0 — 5248 passed, 10 skipped, 10 deselected |
| `ruff format --check` | 0 |
| `ruff check` | 0 |
| `pylint scripts` | 0 (10.00/10) |

**Mutation-tested, 5/5 killed** — restore the 20 KB byte floor; swallow every rejection; run manual
last and only as a fallback; never write `view_type`; guess a winner on an ambiguous name. Each
mutation was confirmed to make the matching test fail, and the source was restored and re-verified.

## What I could not verify

- **No live database, no Power BI Desktop refresh, and no `@tableau-migrator` dispatch** — the
  verified sequence ends at the entry gate. The doc's ~1.5 h / cost figures are carried from the
  requirement, not measured here.
- **POSIX glob behaviour** is documented, not measured (Windows-only machine).
- The demo workbook has an `excel-direct` connection, so `run_estate.py`'s empty-model check reported
  "not judged (no local rows expected)". A workbook over a real warehouse may behave differently at
  that step.
- `preflight.ps1` exits 1 on this machine for a **pre-existing** skill-bundle staleness
  (`powerbi-report-gotchas`, `powerbi-semantic-model-gotchas` stale in the installed plugin),
  unrelated to this branch and deliberately not touched mid-session.


Guust-Franssens:
@copilot Exact-tip blind review at `5ccbd405` returned one reproduced blocker. Make only this documentation correction.

Fresh-clone reproduction:
- `uv sync --all-extras` created `C:\tfmig\wt519\.venv`.
- The documented bare `python` still resolved to the global interpreter.
- `python -m pytest --version` failed with `No module named pytest`.
- `uv run --frozen --directory C:\tfmig\wt519 pytest --version` succeeded.

Required fix:
- In `docs/start-with-one-workbook.md`, make every Python command after `uv sync --all-extras` execute in the environment that uv created. Prefer consistent `uv run --frozen python scripts/...` commands rather than relying on activation state.
- Ensure multiline PowerShell examples remain copy/paste-correct.
- Update only directly related tests if a command-pinning test exists.

Hard scope limits:
- Do not add manual-reference support to `package_unit.py`. The reviewer re-adjudicated that point against issue #519, `AGENTS.md`, and the `tableau-migrator` contract and withdrew it: the simple path intentionally passes the engine bundle and manual `reference/` directory separately.
- Do not change the migration flow, package flow, source layout, time/cost guidance, or bulk-migration behavior.
- Prefer changing only `docs/start-with-one-workbook.md` and an existing direct test if needed.

Run any focused documentation/route tests, commit, and push to this PR branch. It will receive a targeted re-review of this one finding only.

Issue #518, opened 2026-09-04
parse_data_sources returns a schema-valid EMPTY spec for an unrecognised workbook root shape, with no limitation recorded


Guust-Franssens:
## What

`parse_data_sources()` scans only `datasources/datasource` (`scripts/parse_tableau.py:650`). A
workbook whose root shape differs produces a **schema-valid, entirely empty** spec — no data sources,
and no limitation saying anything was missed.

Measured on a well-formed `.twb` with `<datasource>` directly beneath `<workbook>`, carrying custom
SQL that SQLite rejects (`SELECT 1\n+ 1`):

```
data_sources            = []
limitations_encountered = []
migration-spec.schema.json validation = 0 errors
```

Everything downstream then reads this as a workbook that genuinely has no data sources, rather than
as a file the parser could not understand.

## Why it matters

This is the repository's dominant defect class: **unassessable input collapsing into the clean
bucket**. A parser that returns zero sources *and* zero limitations for a file it did not recognise
is indistinguishable from one that correctly parsed an empty workbook. Downstream gates cannot
recover the difference, because nothing was recorded.

The narrow fix is not "support this shape" — it is that a workbook yielding **zero** data sources
should say so as a limitation rather than returning a clean empty spec.

## Scope and honesty about the trigger

- **Pre-existing, not introduced by #514.** Verified: `root.findall("datasources/datasource")` is
  byte-identical on `origin/master`. Descoped from that PR's review for this reason.
- ⚠️ **The trigger shape is synthetic.** It was constructed by a reviewer; nobody has established
  that Tableau itself emits a `<datasource>` as a direct child of `<workbook>`. So treat the *shape*
  as unconfirmed.
- What is **not** speculative is the general behaviour: for any root shape the scan does not match,
  the result is a silent clean empty spec. That is worth closing regardless of which shapes reach it
  in practice.

## Suggested direction

Emit a `limitations_encountered` entry when a parsed workbook yields zero data sources, naming what
was searched. Cheap, and it converts a silent fail-open into a visible one. If a stronger check is
wanted, additionally reject a root that contains `<datasource>` elements which the
`datasources/datasource` path did not reach — that distinguishes "genuinely empty" from "shaped
differently than expected".

## Found by

Blind review of #514 (issue #80 custom-SQL lint), while probing the fail-open class.

PR #524, opened 2026-09-04
Lint inherited custom SQL for doubled operators and literal escapes

### LINKED ISSUE #80: Inherited custom SQL is copied verbatim, so corrupt operators and literal escapes reach generated M
## Summary

✅ Verified: inherited Tableau custom SQL is copied into `migration-spec.json` verbatim, so customer/source corruption such as doubled comparison operators (`<<`, `>>`) or literal two-character `\n` / `\t` escapes can flow downstream without any parser-side lint.

Found from a real customer retrospective dated 2026-07-31, from the earlier agentic-only era. The deterministic parser narrows many classes of LLM drift, but this one is not solved by determinism: faithfully carrying corrupted source SQL still produces corrupted Power Query/native SQL later.

## Evidence

`scripts/parse_tableau.py:397-400` stores custom SQL text without normalization or lint:

```python
"source_relation": "custom-sql" if rel_type == "text" else rel_type,
"custom_sql": rel.text if rel_type == "text" else None,
```

## What breaks

- `<<` / `>>` can appear where the intended SQL comparison was `<` / `>`.
- Literal `\n` after a `--` SQL comment does **not** terminate the comment; following SQL can be swallowed while still looking line-broken in logs/reviews.

## Why existing gates miss it

Structural JSON parsing succeeds. The parser is doing exactly what it was written to do: preserve source custom SQL. `powerbi-report-author validate` is report/PBIR-focused and does not execute or lint inherited SQL.

## Reproduction without credentials

Create a minimal `.twb` relation with `type="text"` whose text contains either:

```sql
select * from t where amount << 10
```

or:

```sql
select * from t -- keep only active\nwhere active = 1
```

Run `scripts/parse_tableau.py` and inspect `migration-spec.json`: the `custom_sql` field will contain the same text.

## Suggested fix

Add a parser-side/custom-SQL lint pass that flags suspicious inherited SQL before any agent or repair script turns it into M:

- doubled comparison operators `<<` / `>>` outside strings/comments;
- literal two-character escape sequences `\n` / `\t` in custom SQL, especially after `--` comments;
- emit a `limitations_encountered` / parse warning rather than silently accepting.

A code check is feasible and preferable to docs-only guidance: this can be unit-tested directly against tiny `.twb` fixtures and does not require live credentials.

Guust-Franssens:
Fixes #80

Supersedes #514 without changing its completed code. GitHub refused to retarget the stacked PR from the closed #502 base, so this draft points the same current branch tip directly at `master`.

Bounded scope:
- report doubled `<<`/`>>` operators and literal `\n`/`\t` escapes in inherited custom SQL;
- never rewrite the source SQL;
- comment-only SQL is high-severity unassessable input;
- EOF comments do not falsely claim following SQL was swallowed;
- bracket-quoted aliases remain valid;
- an end-to-end parser test pins byte-for-byte preservation.

#518 remains separate and is not part of this PR.

Review contract: one targeted blind review against the prior four reproduced blockers. New non-safety classes become residual issues rather than expanding this PR.


Guust-Franssens:
@copilot The review closed three of four blockers. Fix only the remaining spaced postfix-subscript disambiguation.

Current miss:
- `SELECT (ARRAY[10,20,30,40]) [1 << 1]`
- `SELECT array_value(10,20,30,40) [1 << 1]`

Both are executable spaced postfix subscripts; DuckDB returns `20`. `_custom_sql_limitations()` currently masks the bracket as an alias solely because whitespace follows `)`, so the real `<<` operator is missed.

Required behavior:
- Both spaced postfix subscript examples must report the executable doubled operator.
- `SELECT 1 [a<<b]` and ordinary bracket-quoted aliases remain clean.
- Existing unspaced array/subscript cases and string/comment masking remain unchanged.

Refine the existing alias/subscript predicate; do not build a SQL parser or add a new module. Change only `scripts/parse_tableau.py` and focused tests. Add executable positive/negative controls, run the Python edit ritual, pylint by exit code, and the focused parser suite. Commit and push; the same reviewer will confirm only this case.

Issue #558, opened 2026-09-06
Packaged workbook entry gate cannot resolve its own source asset without --source


Guust-Franssens:
## Reproduction

Frozen build: repository `e54cb95`, canonical engine 2.368.0. A workbook was packaged with `package_unit.py` into the canonical flat `packages/<Unit>/` layout.

The package is self-describing:

- `package-manifest.json.artifacts.asset` names `assets/<luid>_<workbook>.twbx`;
- `package-manifest.json.workbook_identity` records the workbook LUID and `match: sha256`;
- `source-provenance.json` records the same input filename, SHA-256 and workbook LUID;
- that exact source asset physically exists inside the package.

### Negative: normal no-flags package invocation

```text
python scripts/check_reference_readiness.py <package> --json entry.json
exit 3
status: CANNOT_ESTABLISH
"no Tableau source workbook could be resolved ... pass --source"
```

The gate reports zero expected pages even though the source bytes and two machine identities are package-local.

### Positive control: identical package plus explicit source

```text
python scripts/check_reference_readiness.py <package> --source <package-source.twbx> --json entry-explicit-source.json
exit 1
status: FINDINGS
9/10 expected pages ready; 1 genuinely blind page
```

The positive control derives all **10** expected pages and attributes all nine admitted oracle records by workbook LUID. Exit 1 is the correct evidence finding; the defect is the no-flags invocation collapsing earlier into `CANNOT_ESTABLISH`.

## Consequence

The documented canonical per-unit package cannot pass through the entry gate using only its own manifest/provenance contract. Operators must supply redundant external knowledge, and automated package workflows stop before evaluating real evidence.

## Acceptance

1. A canonical workbook package resolves its source from the package-local manifest/provenance/asset contract and yields the same expected-page set as explicit `--source`.
2. Preserve strict identity: missing, foreign, hash-mismatched, or ambiguous assets remain non-clean; never pick by caption/fuzzy name.
3. Redaction of an absolute `handover.workbook.source_id` must not erase the package's independently recorded source identity.
4. Datasource-only packages remain earned `NOT_APPLICABLE`.
5. Add a direct no-flags package control and a negative package with a conflicting asset identity.

Guust-Franssens:
## September 7 implementation brief - combined #558 + #562

Base: `master` at `7600a4f7e94dfe98c4e83bb910e1415f4e6ce495`.

[...]

PR #570, opened 2026-09-07
oracle: recover view typing from a session-loss 401 on the Metadata call

### LINKED ISSUE #560: Oracle view typing drops all types on a recoverable Metadata 401 instead of reauthenticating
## Live reproduction

Measured September 6, 2026 at repository `afaee60d`.

`capture_tableau_oracle.py` signed in with REST API 3.29, selected two views, then its one site-wide `tableau_view_types.resolve_and_stamp()` call received Metadata HTTP 401:

```text
view type is UNKNOWN for every view in this run (metadata api returned HTTP 401)
```

The capture then recovered the same session on its first REST export (`re-auth x1`) and completed **2/2** data and PNG legs, exit 0. The manifest therefore contains two complete renders but `view_types.unknown = 2`, so neither can certify the one dashboard page.

### Positive control

Immediately afterward, the standalone census—which calls the same `tableau_view_types.fetch_payload()` transport—succeeded in 1.4s:

```text
workbooks: 48
dashboards_uuid: 60
sheets_uuid: 300
views typed by the shipped parser: 360
shipped parser refused the response: False
VERDICT: CONFIRMED
```

This rules out a persistent permission/capability failure. The recoverable session loss happened on the capture's one-shot Metadata call; the next request reauthenticated and worked.

## Code seam

`tableau_view_types.fetch_payload()` calls `session._request(..., api="metadata")` exactly once and turns any non-200 into a permanent unavailability reason. The same `TableauSession` already has bounded session-loss reauthentication in `export()`, but the view-type call does not participate in it.

## Consequence

A transient session loss permanently downgrades every oracle record in the run to unknown object type. The images and data exist, but the entry/exit gates correctly refuse them as dashboard evidence.

## Acceptance

1. A recognized session-loss response on the Metadata view-type call earns at most one bounded reauthentication and retry.
2. The recovery is recorded in the manifest; it must not be silent.
3. A persistent 401 remains unknown/non-clean, with no fallback to name-based dashboard typing.
4. Permission/GraphQL/schema failures are not retried as session loss.
5. Tests cover `401/session lost -> sign in -> 200`, persistent 401, and a non-auth failure, with reflected credential text redacted before reporting.

Guust-Franssens:
Closes #560 by giving the capture's ONE site-wide Metadata view-type call the bounded

re-authentication the export legs already had.

## What was wrong

`tableau_view_types.fetch_payload()` turned every non-200 into a permanent unavailability

reason, so a **recoverable** session loss (HTTP 401) marked every view `unknown` for the

whole run. Measured 2026-09-06 at `afaee60d`: the same `TableauSession` re-authenticated on

its very next REST export (`re-auth x1`) and the capture completed 2/2 data and PNG legs at

exit 0 — complete images and data that could certify no dashboard page — while the standalone

census, on the same transport moments later, succeeded in 1.4s (48 workbooks, 360 views typed).

## What changed

| file | change |

|---|---|

| `scripts/tableau_view_types.py` | `_fetch_once` (one round trip, now returning the HTTP status) + `_fetch_recovering` (**at most one** re-auth and retry). `resolve_and_stamp` returns `{"unavailable_reason", "reauths"}` and warns on recovery |

| `scripts/capture_tableau_oracle.py` | keeps that record and hands it to `write_manifest` |

| `scripts/tableau_oracle_manifest.py` | new manifest key `view_type_resolution`, as a **parameter** not a `CaptureRun` field — the same reason `server_info` is one |

| `tests/test_capture_tableau_oracle.py` | 11 direct tests, 2 of them through the real `main()` |

| `tests/test_diagnostic_redaction.py` | the taint gate's own declarations: a new seed + three certifications (it found all of this itself) |

| `tests/mutation_view_types.py` | 3 rotted anchors re-anchored; 5 new mutations with `INTENDED` nodes |

| `docs/reference-capture.md` | one paragraph beside the "Metadata API is the only route that tells you a view is a dashboard" claim |

**Direction.** Recognition is **HTTP 401 only**. A 403 (permission), a GraphQL `errors`

block on an HTTP 200, an oversized/undecodable body and a transport exception all keep the

one-shot behaviour they have always had — signing in again cannot change any of them. The

bound is **structural**: `_fetch_recovering` contains no loop, so "at most one" cannot be

raised by editing a counter. A persistent 401 stays `unknown`, with **no** fall-back to

name-based dashboard typing.

## The three questions

**1. Can the tests reach the production path?** Yes, and two of them do nothing else.

`test_the_REAL_capture_run_recovers_its_view_typing_and_says_so_in_the_manifest` and its

persistent-401 twin drive `capture_tableau_oracle.main()` with a scripted session and assert

the **manifest a consumer reads** (`view_types` census, `view_type_resolution`,

`views[0].view_type`), not a helper's return value. The proof that this is load-bearing:

`metadata-recovery-not-recorded` leaves behaviour entirely correct and deletes only the

evidence — it is caught **there and nowhere else**.

**2. Does the guard cover the full surface its name claims?** The surface is one call site:

`fetch_payload` is the module's single transport hop, and both of its callers

(`view_types`/`resolve_and_stamp` for the capture, `tableau_luid_census`) go through

`_fetch_recovering`, so neither can decide separately what a 401 means. The enumerated

failure classes are: 401 (retried once), 403/5xx/any other status, GraphQL `errors`, a

body over the ceiling, undecodable/non-UTF-8 JSON, a transport exception, a failing

`sign_in`, and a session object with **no** `sign_in` — each has a test, and the last three

fail **closed**. ⚠️ Deliberate residual: a server that reports session loss as an HTTP

**200** carrying a GraphQL error is *not* retried, because a 200 `errors` block is also how

a schema gap arrives and typing an estate from an answer we do not trust is the fail-open

direction this module exists to refuse.

**3. Does the evidence prove the verdict?** Five mutations, run through the existing #402

campaign against their `INTENDED` node alone (so the named test is the sole reason to refuse):

```

CAUGHT  metadata-401-never-retried            -> test_the_REAL_capture_run_recovers_its_view_typing_and_says_so_in_the_manifest

CAUGHT  metadata-retry-is-unbounded           -> test_a_PERSISTENT_401_stays_unknown_and_is_retried_at_most_once

CAUGHT  any-failure-retried-as-session-loss   -> test_a_failure_that_is_not_a_session_loss_is_never_retried_as_one[first0-HTTP 403]

CAUGHT  metadata-recovery-not-recorded        -> test_the_REAL_capture_run_recovers_its_view_typing_and_says_so_in_the_manifest

CAUGHT  signin-failure-message-reported       -> test_a_reflected_credential_in_a_SIGN_IN_FAILURE_never_reaches_the_reason

SURVIVED cosmetic-comment-reworded            -> 252 passed

INVALID  absent-anchor-stale-source-text      -> pytest never started

```

The last two are the controls: the campaign is not scoring everything CAUGHT, and a stale

anchor is INVALID rather than credited.

Suite: `pytest -m "not slow"` → **6083 passed, 11 skipped, 2 failed**. Both failures are

`tests/test_harvest_engine_gaps.py` (`unreadable_directory_does_not_fabricate_additions_beneath_it`,

`a_traversal_failure_that_cannot_be_located_suppresses_the_whole_pair`) and are **pre-existing** —

reproduced on this branch with the change stashed. `ruff format`/`ruff check` clean;

`pylint` 10.00/10, exit 0, on the three changed production modules.

## Stated as unverified

* **No live Tableau site was involved.** Every fixture here is scripted; the reproduction in

  the issue is the only live evidence, and the fix is asserted against it by construction.

* **The 401 body is not asserted**, only the status. The live log recorded `HTTP 401` without

  a `401002` code, and the Metadata endpoint is not obliged to spell it the way REST does —

  so making the body part of recognition risked shipping a fix that is inert on the very case

  measured. If a site is found that answers session loss with a non-401 status, this misses it.

* **`tableau_luid_census` inherits the retry** (it shares `fetch_payload`, deliberately —

  divergence there is what the `census-does-its-own-transport-again` mutation guards). Its

  "one query" claim therefore becomes "one query, or two after a 401". Its own suite passes;

  a live census against a 401 was not run.

* **`-n auto --dist loadfile` aborts with an xdist "different tests were collected" error** on

  this machine, on master as well as here; the suite figure above is from a serial run.

---

## Correction rounds (post-review)

**Round 1 — `view_type_resolution` had to survive the hops.** The record was written by the
capture and then **dropped** by both consumers: `group_oracle_by_workbook.subset_manifest` and
`manifest_scope.ORACLE_MANIFEST_ALLOW`. A per-workbook or packaged manifest therefore asserted a
`view_types` census with nothing saying it had been arrived at through a 401 → reauth → 200, or not
arrived at at all — and `merge_batches`, which takes non-view fields from the **newest** batch, let
a clean re-run overwrite an older batch's recovery. Fixed by one shared shipping rule in
`manifest_scope` (value-checked, closed reason vocabulary, per-batch rows beside a conservative
aggregate), used by grouping and packaging alike.

**Round 2 (this commit, `3b64896b`) — two same-class findings: a count that is PRESENT but
impossible was being read as an ABSENCE, so the record came out clean.**

| # | finding | fix |
|---|---|---|
| 1 | a per-batch record whose `reauths` was unreadable (measured: the string `"1"`) normalised to `reauths: null` and was classified **`resolved`**; the merge counted it among the batches with nothing to report and emitted `{"reauths": 0, "unavailable_reason": null}` | a malformed count is **evidence**: the record carries `COUNT_REFUSED`, its row is classified `unreadable`, and the aggregate reason is non-null in consequence |
| 2 | `_shippable_reauths` accepted any non-negative int, so `2+` shipped as recovery evidence no capture could have produced | an individual capture/batch/row is bounded to **exactly `0` or `1`** (`MAX_BATCH_REAUTHS`, mirroring `tableau_view_types.MAX_REAUTH`); bool, string, float and negative are refused |

**Why a sibling-reading sanitiser.** Only a MERGE may total more than one, and only as the sum of
per-batch rows validated at that same bound. `ORACLE_MANIFEST_ALLOW` therefore reads
`view_type_resolution` **beside** `view_type_resolution_by_batch`
(`Sanitized(scope_grouped_resolution, siblings=True)`) and **recomputes** the total from the rows
rather than believing the document. Disagreement in either direction — an inflated total, or one
erased back to zero — is flagged `COUNT_RECOMPUTED` and named in `scope.dropped_fields`. Outside a
mapping spec the marker **raises**, so it cannot silently degrade to reading the aggregate alone.

`_row_total` deliberately does **not** re-validate: every row reaching it was built moments earlier
by `_scope_resolution`, so a second copy of the bound there is unreachable, untestable, and would
become a second place the rule is read from. The bound that matters is the one on the way **in**,
and `test_a_forged_per_batch_row_count_is_never_a_summand` drives a forged row through it. (This
replaced an earlier defensive filter that a mutation proved **equivalent** — no input could reach
it unvalidated.)

### Controls — `tests/test_view_type_resolution_scope.py`, 58 passing

* string / float / bool / negative / `2` / `7` individual counts refused → `{reauths: null, unavailable_reason: COUNT_REFUSED}`
* **negative control:** `0` and `1` still ship verbatim with **no** refused paths
* a **missing count** inside a present record ≠ an **absent whole record** (`{reauths: null, reason: null}` vs `None`)
* one clean + one malformed batch cannot aggregate to a clean record; the malformed batch's row reads `unreadable`
* two recovered batches aggregate to **2** with **two supporting rows**, and survive packaging with no refusal
* a forged aggregate `2` with **no rows at all** is refused
* inflated (`2`←1 row), erased (`0`←two 1-rows) and unsupported (rows themselves unreadable) aggregates are **recomputed from the rows**
* `merge_batches` → `subset_manifest` → package scope stays **non-clean** for a malformed batch, judged against the existing positive control that the same path *does* emit a clean record when every batch is readable
* the paired rule is **idempotent** across the grouping → packaging hop

### Mutation evidence — 10 bounds, 10 killed, each on its intended assertion

```
CAUGHT ceiling-removed-individual-count-unbounded   assert {'reauths': 2...} == {'reauths': None...count refused"}
CAUGHT bool-accepted-as-a-count                     assert {'reauths': True...} == {...count refused"}
CAUGHT float-accepted-as-a-count                    assert {'reauths': 1.0...} == {...count refused"}
CAUGHT negative-accepted-as-a-count                 assert {'reauths': -1...} == {...count refused"}
CAUGHT malformed-count-leaves-no-evidence           assert {'reauths': None, reason: None} == {...count refused"}
CAUGHT malformed-batch-classified-resolved          assert ['resolved','resolved'] == ['resolved','unreadable']
CAUGHT aggregate-believed-rather-than-recomputed    assert None == 1 / assert 0 == 2
CAUGHT aggregate-ceiling-lifted-without-rows        assert {'reauths': 2...} == {...count refused"}
CAUGHT sibling-rows-not-consulted-at-the-allowlist  "a corroborated merge lost its evidence": assert None == 2
CAUGHT rows-summed-before-they-are-validated        assert 3 == 1
```

Harness is scratch and **not committed** (it patches `scripts/manifest_scope.py` in place and
restores it); a stale anchor scores INVALID rather than CAUGHT, and the run is judged by exit code.

### Gates

`origin/master` `da1ec87c` merged in (`996e3c70`, no conflicts, no rebase/amend).
`ruff format` clean; `ruff check conftest.py scripts tests` → exit **0**;
`pylint scripts` → **10.00/10**, exit **0**.
Targeted `pytest` (view-type scope, group-oracle multi-batch + by-workbook, capture-oracle,
diagnostic redaction, package-unit, package-unit gates, check-unit) → **1254 passed, 1 skipped**,
exit 0 (the skip is the pre-existing symlink-privilege skip).

### Stated as unverified (correction rounds)

* **No live Tableau site, and no live merge of real capture batches.** Every control drives the
  real seams (`grp.run`, `merge_batches`, `package_unit._scope_oracle_manifest`) over synthetic
  captures on disk.
* **The `0/1` bound is structural, inferred from `tableau_view_types.MAX_REAUTH == 1`** — it is
  correct for records this repository produces. A manifest written by a *future* producer that
  legitimately re-authenticates twice per capture would be refused rather than carried; that is the
  intended fail-closed direction, but it has not been observed in the field.
* **`unavailable_reason` remains a closed-vocabulary allowlist**, so a producer wording change
  fails a test rather than shipping — but the two new refusal strings are this module's own, and no
  live document carrying them has been read back yet.
* Full-suite `pytest -m "not slow"` was **not** re-run for this correction; the targeted set above
  is what was executed.


Guust-Franssens:
Closing under the customer-ready review stop rule. After the original review and two bounded same-class corrections, package scoping still permits a malformed sibling `reauths` row to validate a clean top-level aggregate by omitting the refused count from `_row_total`. Another local guard would be a fourth pass over the same evidence class. The branch is preserved. #560 remains open as a fail-closed polish gap: a transient Metadata 401 may leave view types unknown, but no wrong page is certified. Package-level recovery evidence should be reconsidered with the strict #558/#562 package verifier rather than patched again here.

Guust-Franssens:
Policy clarification after the September 10 closure audit: closing this branch did not complete or abandon #560. A smaller continuation is explicitly queued after the strict package/S2 authority, where malformed per-batch recovery evidence can be assessed without another local aggregate guard. The issue remains priority-next and fail-closed.

Issue #192, opened 2026-08-17
Project-scoped lineage must derive stable survey identities without widening scope


Guust-Franssens:
## Current executable scope — September 7, 2026

The remaining product defect is the lineage denominator: a filtered survey can be widened by Metadata-only rows, and the failed #553 attempt showed that caption/LUID predicates still omit valid partial-Metadata and duplicate-caption occurrences. The next implementation must derive scoped survey occurrences by stable identity first and use Metadata only as enrichment. The harvest all-failed exit and download liveness claims in the original title are fixed; bytes/ETA is not part of this issue. Project-scoped oracle capture remains the documented repeated-`--workbook` composition unless customer evidence proves that insufficient.

---

[...]

PR #574, opened 2026-09-08
Accept a pasted Tableau project URL, and refuse its numeric project id legibly

### LINKED ISSUE #191: Tableau's numeric project id (the one in the site's own URL) has no public API mapping - --project scoping needs a resolution path
While scoping a migration to a single project via a URL like `https://<site>/#/projects/35`, there is **no way to resolve that numeric `35` back to a project** via the public Tableau REST API or the Metadata (GraphQL) API. Both expose only GUID-based `id`/`luid` fields.

Confirmed by direct query against a live Tableau Server site (2026-08-17):

- REST `GET /sites/{id}/projects` — returned the site's projects; **none** had an `id` or `name` matching the literal `"35"` from the URL.
- Metadata API `projects` field query — returned **`FieldUndefined`**.
- Root cause confirmed by asking a human to open the link and read the project name off the page.

So the numeric id in Tableau's own web UI routes is a **legacy internal identifier with no public API surface**.

## Why this matters for #185

#185 proposes `--project <name>` / `--project-id <luid>` so a customer can try a migration on one project instead of the whole site. But **pasting a project URL out of the browser is the single most natural way a customer identifies "the one project I want"** — and that URL contains only the unresolvable numeric id.

So the ergonomics gap is: the most likely input a customer supplies is the one input we cannot accept. Today the only workaround is asking a human to open the link and read the name off the page, which is exactly the kind of friction #185 exists to remove.

## Suggested handling

1. **Document the gap plainly** wherever `--project` is documented: a project URL is *not* resolvable via the public API, and the customer must supply the project **name** or **LUID**.
2. **Add a `--project-url <url>` convenience flag** that parses a pasted URL and:
   - if it contains a **GUID**, uses it directly (some Tableau URL shapes do carry the LUID);
   - if it contains a **numeric id**, prints a clear, actionable message — *"Tableau's numeric project id (`35`) has no public API mapping. Open the link and pass the project **name** instead: `--project \"<name>\"`."*

   The point is to fail **legibly at the moment of pasting**, rather than silently returning nothing or attempting a wrong-endpoint lookup that 404s or returns an empty set. An empty result set is the dangerous outcome here — it looks like "that project has no content" rather than "we could not resolve your input".
3. Consider accepting a **project name** with a helpful near-miss suggestion when it does not match (`did you mean "00 - Certified Sources"?`), since names are what humans will actually have.

## Note on identifier ambiguity

Project **names are not unique across a project tree** in Tableau — nested projects can repeat a name under different parents. So `--project <name>` needs a defined behaviour when a name matches more than once: either list the matches with their LUIDs and stop, or require `--project-id`. Worth deciding alongside this.

## Provenance

Reported from a live customer migration session, 2026-08-17. Customer-identifying details are deliberately omitted — this repository is public.

Guust-Franssens:
Fixes #191.

A customer scoping a migration to one project reaches for the URL in their browser bar. Tableau's own
web-UI route carries a **numeric** project id (`https://<site>/#/projects/35`), and nothing public can
resolve it — verified against a live site 2026-08-17 (REST `GET /sites/{id}/projects` exposes only GUID
`id`s; the Metadata API answers `FieldUndefined`), and field-confirmed independently (SES, 2026-08-21:
the workaround was a human opening the link and reading the name off the page). So the most natural
input was the one input the CLI could not accept, and pasting it produced an **empty selection**, which
reads as *"that project has no content"* rather than *"we could not resolve your input"*.

## What changed

`scripts/harvest_estate_assets.py` gains `--project-url URL` (repeatable), at the existing project-scope
entry surface:

| input | behaviour |
|---|---|
| URL carrying a **LUID** (fragment route, path route, REST shape, trailing slash, query, percent-encoded, braced, uppercase) | normalised into the existing exact `--project-id` path, **canonicalised to the lowercase unbraced form via `uuid.UUID`** — no new resolver, no new endpoint, no network call |
| URL carrying a **numeric** id | refused **before** the `--out` guard, engine resolution, the `.env` read and any sign-in or download; echoes the number and names the two inputs that work |
| malformed authority / no `/projects/<id>` / non-http / neither LUID nor numeric / a percent-decoded control or zero-width character | explicit **sanitized** usage error (argparse exit 2 — never a raw `urlsplit` traceback, and never exit 1, which in this script means *nothing could be assessed*) |
| two different `/projects/<id>` segments in ONE url | explicit ambiguity error — never a best guess |
| the same GUID twice, or a GUID already given as `--project-id` | one project, not two |
| a mixture containing one unresolvable URL | refuses the **whole** invocation — no partial scope |

`--project`, `--project-id`, descendant/dependency selection, exact-match ambiguity rules and every exit
code are unchanged. No fuzzy name matching was added, and nothing pretends a numeric id is resolvable.

Docs: the `scripts/README.md` row for the script, and one paragraph in `docs/operator-runbook.md` beside
step 4 of the happy path.

## Three review questions

**1. What is the invariant, and which way does it fail?** An input that cannot be turned into an *exact*
LUID must produce a loud usage error, never a scope. Fail-open here is the dangerous direction: a silent
empty selection is indistinguishable from an empty project and sends the operator back to a site-wide
scan. The refusal is therefore total (`project_ids_from_urls` validates every URL before applying any)
and early (first statement after `parse_args()`, ahead of the `--out` guard, `engine_scripts_dir()`,
`resolve_env()`, `require()` and `download`).

**2. What is the closed surface?** One CLI flag reaching one existing consumer. Every project identifier
still arrives at `scoped_todo(con, project_names, project_ids, workbooks_only)` — the URL path *is* the
`--project-id` path, so descendant selection, dependency pull-in and the ambiguity rules have no second
code path to drift from. Segment extraction reads only `urlsplit().path` and the fragment route, scanned
**separately** (a path ending in `projects` can never pair with the first fragment segment), and never
the query. Residual, stated deliberately: a LUID that is well-formed but absent from `estate.db` still
exits 1 with `no projects matched …` — correct, and not something a URL parser can know.

**3. What is the independent oracle, and was it mutation-tested?** ✅ The equivalence claim is judged by
running `main()` twice over the same fixture estate — once `--project-id <guid>`, once
`--project-url …/#/projects/<guid>` — and comparing the resulting `parse-sweep.json` selections, not by
inspecting the parser. ✅ The early-refusal claim is judged by stubbing the `--out` guard, engine
resolution, the `.env` read and `download` to raise if reached. ✅ Mutation-checked: deleting the
refusal from `main()` leaves all parser tests green and fails **only** the production CLI tests
(`2 failed, 41 passed`), which is the control the brief asked for. ⚠️ Redaction is proved by a negative
control (`admin`, `hunter2`, `SEKRIT-TOKEN`, `:token` absent from the message) plus a positive control
(host and the numeric `35` present) — a denylist over a constructed URL, not a proof over all URLs;
the mechanism it rests on is that only `urlsplit().hostname` is ever formatted, which drops userinfo.

## Verification

- `pytest tests/test_harvest_project_scope.py` — ✅ **43 passed** (was 20).
- `ruff format --check` / `ruff check` over `conftest.py scripts tests .github/skills` — ✅ exit 0.
- `pylint scripts/harvest_estate_assets.py` — ✅ 10.00/10, exit 0 (CI lints `scripts`, not `tests`).
- `python scripts/check_navigation_index.py` — ✅ exit 0.
- Manual CLI: numeric URL with userinfo + query token → exit 2, message carries `tableau.example.com`
  and `35` and none of the secrets; no `_sweep` directory created.
- ⚠️ `tests/test_harvest_engine_gaps.py` has **2 pre-existing failures** on this machine, reproduced on
  untouched `master` at `ed2dadcd` — unrelated to this branch.

## Not verified

- ❌ Not exercised against a live Tableau site: every test is offline. The numeric-vs-GUID fact itself
  is #191's live-site evidence, not re-measured here.
- ⚠️ The Tableau URL shapes covered are the ones in the issue plus the REST shape; a site or version
  emitting a project route this parser does not recognise fails **closed** (usage error naming the two
  usable inputs), which is the intended direction but means an unknown-but-valid shape is refused.
- ❌ `tableau_lineage.py` (#192), `assess_estate.py`, oracle capture and `run_estate.py` are untouched —
  they keep their current project-scope surfaces.

---

**Round 1 correction applied in `f7105fc9`**, after merging `origin/master` `da1ec87c` (merge, not rebase). Three reproduced URL classes: a raw `ValueError` from a malformed authority becoming exit 1, a verbatim uppercase GUID selecting no row, and `%0A` becoming an invisible newline inside an accepted project id. Detail, the mutation table (including one mutation that **survived** the first attempt) and the no-widening sweep are in the review comment below. Tests 43 → 68.


Guust-Franssens:
## Review round 1 — three URL classes corrected

`origin/master` `da1ec87c` merged in (no rebase, no amend); correction is a new commit `f7105fc9`.

All three were **reproduced against `e6b8f72d` before being changed**, and all three share one shape:
the failure was invisible.

| # | reproduced on `e6b8f72d` | now |
|---|---|---|
| 1 | `https://[::1/#/projects/35` → `ValueError: Invalid IPv6 URL` escaped as an uncaught traceback, **exit 1** | parsed **once, inside the guard** → sanitized `ProjectUrlError` → **exit 2** |
| 2 | `…/projects/A85BDE90-…` → `['A85BDE90-…']` forwarded verbatim | canonicalised through `uuid.UUID` → `['a85bde90-…']`; `{braced}` / `%7B…%7D` accepted too |
| 3 | `…/projects/<luid>%0A` → `['a85bde90-…\n']` accepted as a project id | refused: `non-printable character(s) (U+000A)` |

**Why each was a wrong outcome, not an ugly one.** Exit 1 in this script *means* "NOTHING COULD BE
ASSESSED" — a harvest verdict standing in for a typo. And Tableau's REST API and `estate.db` both hold
LUIDs **lowercase**, so both #2 and #3 produced a perfectly well-formed exact-match `--project-id` that
matches **no row** — the silent empty selection this flag exists to prevent, reached from a URL naming
a real project.

**Sanitization.** The urllib message is never echoed: one of its forms quotes the netloc back
(`netloc '...' contains invalid characters under NFKC normalization`), and the netloc is where
`user:password@` sits. Only the exception *class* name survives, `raise … from None` suppresses the
chained context, and the control-character refusal names only codepoints (`U+000A`), never the segment.
A test asserts `__cause__ is None and __context__ is None` alongside the leak denylist.

**`uuid.UUID` is the normaliser, not the gate.** The strict dashed regex still decides what a LUID *is*,
so `urn:uuid:…`, undashed 32-hex, `%20`-padded and suffixed forms stay refused — `uuid.UUID` would
accept the first two. Both matches are now `fullmatch`.

A valid URL beside any of these still refuses the **whole** invocation (two new CLI cases cover
valid+malformed and valid+control).

## No widening — the descope condition was checked, not assumed

The brief said: if another URL grammar class appears, drop to numeric-refusal-only. An adversarial
sweep — backslashes, `..` segments, `urn:uuid:`, undashed hex, IDN host, invalid port, `%FF`, fullwidth
digits, `;jsessionid=`, empty segments, uppercase `/PROJECTS/`, a 5 kB segment, `#` twice — produced
**only** exact LUIDs or sanitized refusals, and no raw exception. So no new class appeared and the GUID
convenience stands.

## Mutations (each killed by an end-to-end CLI test, not only a parser test)

| mutation | result |
|---|---|
| parse outside the guard | 5 failed — incl. `test_the_cli_refuses_a_malformed_or_control_bearing_url_as_a_usage_error[urls0, urls2]` |
| `return inner` (no canonicalisation) | 5 failed — incl. `test_the_cli_selects_the_real_project_from_an_uppercase_or_braced_url[×2]` |
| drop `_refuse_control_characters` | 6 failed — incl. the CLI test `[urls1, urls3]` |
| `PROJECT_LUID.match` instead of `fullmatch` | ⚠️ **survived the first round** — masked by the control check. Fixed by adding the printable-suffix cases (`<luid>x`, `<luid>%20extra`, `35x`); now 3 failed, incl. the CLI test `[urls4]` |

That fourth row is the honest part: the first correction was under-tested, and the mutation harness —
not review, not reading — is what said so.

## Gates

- `pytest tests/test_harvest_project_scope.py` — ✅ **68 passed** (43 → 68), exit 0.
- `ruff format --check` / `ruff check` over `conftest.py scripts tests .github/skills` — ✅ exit 0.
- `pylint scripts/harvest_estate_assets.py` — ✅ 10.00/10, exit 0.
- `python scripts/check_navigation_index.py` — ✅ exit 0.
- Manual CLI: malformed-authority URL with userinfo + token → **exit 2**, no traceback, no secret;
  `<luid>%0A` → **exit 2**; no `_sweep` created in either case.

## Residual, unchanged from the original PR

- ❌ Still not exercised against a live Tableau site — every test is offline.
- ⚠️ An unknown-but-valid project route fails **closed** (usage error naming `--project` /
  `--project-id`), which is the intended direction but refuses a shape we have not seen.
- ⚠️ Redaction rests on the mechanism (only `hostname` and codepoints are ever formatted) plus
  positive/negative controls over constructed URLs — not a proof over all inputs.
- Docs updated to match: canonicalisation and the exit-2-not-exit-1 rule are now stated in
  `scripts/README.md` and the `docs/operator-runbook.md` paragraph.

PR #571, opened 2026-09-08
Project semantic-model table paths before estate conversion

Guust-Franssens:
## Recovery checkpoint

Recovered intact after the September 7 Windows Terminal crash. The branch had seven uncommitted files; `git diff --check` was clean, ruff was clean, and the three targeted path suites passed **140 tests** before checkpointing. It has now been merged with current `master`.

This PR projects both pre-conversion path families: canonical PBIR visual paths and semantic-model `definition/tables/<table>.tmdl` paths derived conservatively from the source documents. It names the binding family and refuses when table names cannot be bounded.

**Review status:** independent blind review is still required. This body deliberately uses only a plain reference to #564; do not close the issue until the exact current tip is reviewed and CI-clean.

Refs #564.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

PR #584, opened 2026-09-09
Bound the semantic-model table path before conversion (#564)

### LINKED ISSUE #564: project_estate_path_ceiling ignores semantic-model TMDL table paths
## Finding

`run_estate.project_estate_path_ceiling()` projects only the deepest canonical PBIR visual path:

```text
pbip/<unit>/<unit>.Report/definition/pages/<page>/visuals/<visual>/visual.json
```

It has no corresponding semantic-model term such as:

```text
pbip/<unit>/<model>.SemanticModel/definition/tables/<table>.tmdl
```

Measured on canonical engine 2.368.0:

- current real estate PBIR tail: 139;
- current model tail: 129;
- after experimentally shortening page/visual IDs, the report tail fell below the unchanged model tail, so the model became the actual path-budget constraint.

The shipping-time `check_path_ceiling.py` walks the real tree and catches this. The gap is in the **pre-conversion** safety projection: once PBIR identifiers/folders become shorter, a long table/relation name can exceed Desktop's path limit while the projected PBIR-only path still fits.

## Current severity

Fail-closed/latent on today's engine because the conservative PBIR envelope is usually longer. It becomes fail-open if the upstream identifier/folder shortening proposal lands without adding the model term.

Engine-backed path-shape coverage is now committed through #563: ordinary datasource, field-swap datasource and real workbook, exact file/directory boundaries and production attribution.

## Acceptance

1. Project both PBIR and semantic-model path families before conversion.
2. Derive or conservatively bound the emitted model/table filename using source metadata available before the engine writes.
3. Return `CANNOT_ASSESS`, never clean, when the longest emitted table filename cannot be bounded.
4. Identify which path family binds the result and report its measured UTF-16 length/ceiling.
5. Add controls where PBIR is longest, where TMDL is longest, and where a one-unit understatement flips pass to refuse.
6. Keep the existing exact 259-file / 247-directory boundaries.

Related upstream proposal: `Yarbrdab000/tableau-fabric-skills#194`.

Guust-Franssens:
## What

`run_estate.project_estate_path_ceiling` projected only the canonical PBIR visual path, so a
`pbip/<unit>/<model>.SemanticModel/definition/tables/<table>` part longer than that path could pass
the **pre-conversion** gate. Canonical engine 2.368.0 emits exactly that shape: a required 273-unit
table part at the skill's ordinary 22-unit run root (#565), while the `.pbip` pointer itself is
legal. The projector now covers **both** families, refuses if **either** breaches, and reports which
family binds and why.

This is a clean replacement for the closed PR #571, not a resumption of it.

## The design decision that matters

The model term is a **generic source-component envelope**, not a per-class reimplementation of the
engine's naming. Three emitted filename classes each defeated the class-by-class approach in turn -
`combine_descriptors`' duplicate-relation `Relation (<datasource>)`, a long range/what-if parameter
caption, and the per-island `Date (<datasource>)` calendar - because the class inventory is owned by
the engine, so a class it has and we lack is *silently absent* rather than loud. (Extract-collapse
is a fourth, source-visible candidate.)

Instead the bound is: **the longest source-owned identity component in the input documents x the
largest number of components any censused write site combines (2), plus that shape's fixed
punctuation and a uniquification allowance.** The write-site census was read on 2.368.0 from every
expression that CREATES a `definition/tables/<name>` part (`assemble_model.py` data tables, date
dimensions, `_Measures`, DirectLake, `_inject_field_param_tables`; the remaining occurrences are
enrichment sites guarded by `if path in parts` and cannot mint a filename) plus the name-composing
expressions those sites consume.

Two properties are load-bearing and intentional:

* **Version-gated.** An engine outside the audited set, a write-site census fingerprint that no
  longer matches, or an unreadable source yields `cannot_establish` - and therefore a refusal -
  never `ok`. Re-auditing is the unblock; the refusal says so.
* **It over-refuses.** A workbook whose own captions are long can be refused at every output root
  even though the engine would have fitted. The refusal names the binding family and the component
  driving it, so the actionable move (shorter run root, or shorter identity in the source) is
  visible.

The existing PBIR projection and the exact 259-file / 247-directory UTF-16 boundaries are unchanged.

## Evidence

Controls run the **canonical engine** and compare `projected >= actual` on the bytes it really wrote
(never against an expected filename, which is blind to exactly the class that defeats it):

| control | emitted deepest model part | projected | actual |
|---|---|---|---|
| duplicate relation + per-island `Date` | `regional_sales.csv (Beta Consolidated ... Source).tmdl` | 283 | 164 |
| what-if parameter | `Rolling Inventory Turnover Threshold for FY2026 Q3 Review.tmdl` | 282 | 136 |
| astral identity | `ventas_🌍_consolidado.csv.tmdl` | 225 | 142 |
| two long components (kill control) | `Regional Sales ... Q3 Detail Snapshot (Delta Regional ... Source).tmdl` | 285 | 214 |

Plus: the committed issue-194 long/short pair, the field-swap datasource, the ordinary datasource
and a real workbook (`tests/test_datasource_path_envelope.py`), a one-unit boundary flip on the
model family, and `cannot_establish` for an unaudited version / a moved census / an unparseable or
identity-free source.

**Mutations, each required to fail independently:**

1. drop the second source component -> a 101-unit bound under a 134-unit emitted filename;
2. drop the fixed overhead (literal + uniquifier) -> 131 under 134;
3. widen the version gate -> the unaudited-engine refusal disappears, proved on the canonical tree
   with a simulated version so the census half is genuinely satisfied.

⚠️ Stated residual: the two sub-terms of (2) are killed together. The uniquifier allowance is
deliberate slack for the `" 2"` / `"_2"` climbs, and no committed control kills it alone.

## Notes for review

* `tests/test_run_estate.py` gets an autouse fixture that stubs the model evidence for the ~40
  `main()` tests built on stub engine trees (which cannot satisfy a version-gated census). The gate
  is not left unproven: `test_main_refuses_when_the_model_family_cannot_be_bounded` restores the
  real function and drives `main` through it.
* `tests/test_e2e_offline.py` needed the same one-call stub around its mock engine.
* `scripts` + `conftest.py` pylint 10.00/10, ruff clean, full suite 6122 passed (the two
  `test_harvest_engine_gaps.py` failures reproduce on unmodified `master` and are unrelated).

Fixes #564

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -124,6 +124,9 @@ EXPECTED_ENGINE_SKIP_REASONS_BY_NODEID: dict[str, str] = {
     "tests/test_datasource_path_envelope.py::test_the_engine_itself_resolves_both_declared_swap_controllers": (
         "deterministic tier not installed"
     ),
+    "tests/test_datasource_path_envelope.py::test_the_model_family_projection_covers_every_emitted_shape": (
+        "deterministic tier not installed"
+    ),
     "tests/test_datasource_path_envelope.py::test_the_production_identifier_envelope_stays_above_the_engines_"
     "measured_cap": "deterministic tier not installed",
     "tests/test_datasource_path_envelope.py::test_the_production_projection_refuses_each_emitted_shape_at_its_own_"
@@ -162,16 +165,33 @@ EXPECTED_ENGINE_SKIP_REASONS_BY_NODEID: dict[str, str] = {
     "tests/test_harvest_download_watchdog.py::test_the_ceiling_constants_match_the_INSTALLED_engine": (
         "canonical engine not installed, so its constants cannot be read"
     ),
+    "tests/test_issue_194_long_pbir_path.py::test_dropping_the_fixed_overhead_understates_real_engine_output": (
+        "deterministic tier not installed"
+    ),
+    "tests/test_issue_194_long_pbir_path.py::"
+    "test_dropping_the_second_source_component_understates_real_engine_output": "deterministic tier not installed",
+    "tests/test_issue_194_long_pbir_path.py::test_each_control_reproduces_the_naming_class_it_exists_for": (
+        "deterministic tier not installed"
+    ),
     "tests/test_issue_194_long_pbir_path.py::test_the_engine_warning_is_asserted_only_where_it_was_established": (
         "deterministic tier not installed"
     ),
     "tests/test_issue_194_long_pbir_path.py::test_the_long_case_crosses_the_file_ceiling_at_the_skill_default_root": (
         "deterministic tier not installed"
     ),
+    "tests/test_issue_194_long_pbir_path.py::test_the_model_projection_covers_every_emitted_table_part": (
+        "deterministic tier not installed"
+    ),
+    "tests/test_issue_194_long_pbir_path.py::test_the_model_projection_covers_the_committed_long_and_short_pair": (
+        "deterministic tier not installed"
+    ),
     "tests/test_issue_194_long_pbir_path.py::test_the_output_root_length_is_what_decides_this_boundary": (
         "deterministic tier not installed"
     ),
     "tests/test_issue_194_long_pbir_path.py::test_the_short_control_stays_inside_both_ceilings_at_the_same_root": (
+        "deterministic tier not installed"
+    ),
+    "tests/test_issue_194_long_pbir_path.py::test_the_version_gate_is_what_refuses_an_unaudited_engine": (
         "deterministic tier not installed"
     ),
     "tests/test_issue_424_chart_type_pin.py::test_permanent_invariant_survives_any_future_fix["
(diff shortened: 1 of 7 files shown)


Guust-Franssens:
CI note (second commit): the repository's engine-integration job pins **2.356.0** while the write-site census was audited on **2.368.0**, so `model_envelope_evidence` correctly returns `cannot_establish` there - that is the designed production outcome, not a defect. The engine-backed controls now assert the claim their engine establishes (the bound where it applies, the refusal where it does not), and the two arithmetic mutations read the version-independent component bound, so neither arm is a no-op and nothing is skipped. All four checks pass: `checks`, `engine-integration`, `windows-bundle` (`engine-drift` skipped).

Guust-Franssens:
Closing under the path stop rule. Blind review proved the generic preprojection still false-cleans: the census is substring counts rather than canonical source fingerprint, multi-definition `.tdsx` selection differs from the engine, and decimal collision suffixes are unbounded. The branch also did not contain current master and therefore lacked the merged prompt-injection fix. A third local naming model would continue the same failure class. Replacement: retain the existing conservative PBIR preflight, run the canonical engine, then invoke the existing actual-tree path gate immediately before provenance/agent work. Python can emit the tree; downstream Desktop work must not start when actual files/dirs exceed 259/247 or cannot be assessed. Permanent filename shortening remains upstream #194.

PR #585, opened 2026-09-09
feat: closed PBIR/model census and revision digest (#363 A1a)

Guust-Franssens:
Slice **A1a** of #363 and nothing else.

## Invariant

Given an already-selected local `fabric/` root and ONE report artifact, produce a strict typed
PBIR/model artifact census plus a deterministic revision digest, **or an explicit refusal**.

Out of scope by construction (later slices): package-manifest selection, Desktop/bridge/PID,
`reload`, screenshots, iterations, receipts, comparison, sign-off. This module has no live evidence,
so it can never claim Desktop LOADED these bytes - the independent A1 audit is explicit that a
disk-only check is not a loaded-revision oracle. Input is `(fabric_root, report_dir)` so A1b can
supply verified package roles later.

## API

```python
establish_revision(fabric_root, report_dir) -> PbirRevision | RevisionRefusal
```

`PbirRevision` carries `pbip/report/model/model_binding`, `page_order`, `pages` (each with folder,
document id, display name and visuals), a per-file SHA-256 map, the explicit `excluded` list with
reasons, and `digest`. `RevisionRefusal` carries a code, a detail and evidence. A refusal is
**falsy** and a revision **truthy**, so `if revision:` fails closed.

## Closed, not permissive

Every entry inside the report and the model is either **included**, **excluded by an explicit named
rule** (`.pbi/` local Desktop state, `TMDLScripts/` authoring scratch, volatile sidecars), or
**refused as unknown**. No `rglob("*")` anywhere.

The allow-lists are derived from the committed corpus, not assumed. Measured across
`examples/*/fabric` (16 reports, 36 pages, 869 visuals):

* page folder == `page.json.name` in **36/36** - the folder IS the page document id;
* visual folder != `visual.json.name` in **133/869** - the folder is NOT the visual id, so both are
  recorded and uniqueness is asserted **report-wide** on the document id;
* `fixtures/large-refresh` commits a page with **no** `visuals/`, so that is a census outcome
  (`visuals: ()`), not a refusal.

## Authority enforced

* `report_dir` contained under `fabric_root` **both lexically and resolved**, with no
  symlink/junction/reparse point on any component; a case alias refuses as `path_identity_mismatch`.
* Strict JSON everywhere it is load-bearing: duplicate keys, `NaN`/`Infinity` and non-object roots
  all refuse.
* Exactly one immediate `.pbip` declares this report (all are parsed, so a second claim cannot hide);
  zero -> `pbip_absent`, unrelated -> `pbip_report_unrelated`, two -> `pbip_report_ambiguous`.
* `definition.pbir` must hold exactly one **local, contained** `byPath` `.SemanticModel`;
  `byConnection`, zero/multiple bindings, dangling, report-target, escaping and backslash spellings
  all refuse.
* `pages.json.pageOrder` is the page-order authority; page directories must agree exactly
  (`page_missing` / `page_orphan`), `activePageName` must be a member.
* Report-level render bytes (`report.json`, `version.json`, pages metadata, `StaticResources`) and
  every deployable model definition byte (`.platform`, `definition.pbism`, TMDL/culture/
  relationships/expressions) are covered. A **dangling** registered resource refuses; an
  **unregistered** resource file is included instead, so its bytes still move the digest.
* Digest = SHA-256 over an explicit version tag plus length-delimited, sorted fabric-relative paths
  and raw bytes. Enumeration order cannot move it - proved at both layers.
* Every shareable field, refusal detail and evidence item is fabric-relative or an opaque token; no
  absolute path reaches `repr()`/`asdict()`.

## Controls (111 tests)

* **Corpus**: all 17 committed PBIP-bearing trees establish (**1231** files covered, 37 pages, 869
  visuals). The 8 committed gate fixtures with no `.pbip` refuse as `pbip_absent` - listed in an
  exact classification table, so a newly committed report shape must be classified rather than
  adopted by silence. The `shared-datasource` fixtures, given the `.pbip` they omit, are the real
  binding corpus: local -> establishes, absent model -> `model_unresolved`, resolvable-but-outside
  model -> `model_not_contained`, missing `datasetReference` -> `binding_malformed`.
* **Negative controls**, each asserting its OWN code: one-byte changes in `.pbip`/`definition.pbir`/
  report/page/visual/resource/`.platform`/`definition.pbism`/TMDL; missing `page.json`/`visual.json`;
  pageOrder mismatch/orphan/duplicate; duplicate page and visual document ids; wrong model target;
  `byConnection`; junction and case alias; unknown file/dir at every level; malformed types and
  duplicate JSON keys.
* **Mutation-tested**: 18 rule mutations applied to the module one at a time, **18/18 caught**, each
  by the test written for that rule. Two gaps this found and closed: the digest's own sort was
  unproven (the filesystem-level control was vacuous for it), and a `localSettings.json` name rule
  nothing exercised - removed rather than left as untested surface.

## Residuals

* A fabric root with **no** `.pbip` refuses. Desktop opens the `.pbip`, so such a root is not a local
  capture target - but it is why 8 committed gate fixtures refuse here.
* Unsupported-but-real shapes (`SharedResources/` on disk, `roles/`, `perspectives/`, bookmarks)
  refuse rather than being adopted. Widening is a deliberate edit with corpus evidence.
* `_is_reparse` answers `False` when `lstat` itself fails; the subsequent read/list then refuses.
* Disk-only: this proves what is on disk now, never that Desktop loaded it. That is A1b.

Refs #363. Draft - do not merge.

### DIFF
--- a/scripts/pbir_revision.py
+++ b/scripts/pbir_revision.py
@@ -1,0 +1,856 @@
+"""
+purpose: strict census + deterministic revision digest for ONE local PBIR report and its bound model (#363 slice A1a).
+usage:   import pbir_revision; pbir_revision.establish_revision(fabric_root, report_dir)
+
+What this is
+------------
+Slice A1a of #363 and nothing else: given a `fabric/` root that somebody else has already selected and
+ONE report artifact inside it, either return a strict typed census of every byte that a local capture
+would render from - plus a deterministic digest over exactly those bytes - or refuse and say why.
+
+Deliberately NOT here (later slices own them): package-manifest selection, Power BI Desktop, PIDs,
+`reload`, screenshots, iteration allocation, receipts, comparison, sign-off. There is no live
+evidence in this module, so it can never claim that Desktop LOADED these bytes - only what is on
+disk right now. The independent #363 A1 audit is explicit that a disk-only check is not a
+loaded-revision oracle; A1b adds the PID-scoped barrier on top of this census.
+
+Why a closed census rather than a glob
+--------------------------------------
+The digest is only worth having if the set it covers is decided by an explicit rule. A permissive
+`rglob("*")` silently adopts whatever a future Desktop build writes beside the definition - the exact
+shape that lets a report change while its "revision" does not, or lets a machine-local file make two
+identical reports disagree. So every entry inside the report and the model is either:
+
+* **included** - a byte that a render or a deploy depends on, or
+* **excluded by an explicit named rule** (`.pbi/` local Desktop state, `TMDLScripts/` authoring
+  scratch, volatile sidecars), or
+* **refused** - unknown, so this module does not get to guess.
+
+The allowed sets below were derived from the committed corpus, not assumed. Measured on this repo at
+`bf4a79f1` across `examples/*/fabric` (16 reports, 36 pages, 869 visuals):
+
+* report top level is exactly `.platform`, `definition.pbir`, `definition/`, `StaticResources/`;
+* `definition/` is exactly `version.json`, `report.json`, `pages/`;
+* a page folder holds `page.json` and (in `fixtures/large-refresh`, zero-visual) an OPTIONAL
+  `visuals/`; a visual folder holds exactly `visual.json`;
+* **page folder name == `page.json.name` in 36/36 cases**, so the folder IS the page document id;
+* **visual folder name != `visual.json.name` in 133/869 cases**, so the folder is NOT the visual
+  document id - both are recorded, and uniqueness is asserted on the document id, report-wide;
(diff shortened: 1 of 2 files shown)


Guust-Franssens:
Closing at the A1a split/delete trigger. Blind review found independent failures in TOCTOU/reparse containment, outside-root resources, context-free load-bearing exclusions, path aliases, malformed PBIP types, cross-drive exceptions and corpus vacuity. The 856-line module duplicates the package verifier's file/revision authority and still cannot produce a stable snapshot. Do not correct or resume this branch. Phase 3 will consume the strict package working-verification result and only add PID/status/reload identity with pre/post package-revision equality; there will be no second PBIR filesystem census.

Issue #692, opened 2026-09-17
Add explicit cost-disclosed exact validation for custom SQL


Guust-Franssens:
## Operator/customer impact

- **Why it matters:** Some custom SQL returns few rows but performs expensive scans, resumes compute, or runs a complex non-folding plan. It must never execute merely because the customer asked whether Power BI can connect.
- **Journey change:** Default probing proves connection scope without running customer SQL. Full exact-query validation becomes an explicit, cost-disclosed action.

## Proposed explicit mode

Add a positive opt-in such as `--custom-sql-mode exact`. Before execution, disclose that the original statement may resume compute, scan substantial data, incur cost, or have side effects.

Requirements:
- Preserve original SQL byte/character semantics exactly; only M string escaping is allowed.
- No comment removal, whitespace collapse, TOP, LIMIT, WHERE 1=0, or wrapper query.
- Fully consume the query result before claiming success; first-row success is insufficient.
- Native-query approval returns `OPERATOR_REQUIRED`.
- Completed rows return `DATA_OK`; completed zero rows return `DATA_EMPTY`.
- Credential, access, network, timeout and unknown-popup verdicts remain distinct.
- Never run by default and never infer approval from the migration brief unless the exact costly action was authorized.

## Controls

- Server-observed SQL hash equals the original statement.
- Expensive/side-effect counter remains zero without explicit mode.
- Exact mode executes once and fully consumes results.
- Query permission failure never inherits connection-only success.
- Zero rows are honest `DATA_EMPTY`, not a manufactured row.

## Plan required

Define the exact CLI, files, audit identity, cost wording, full-consumption evidence and interaction with operator-driven refresh. Keep popup observation in #687/#146 and default connection proof in #690.

Refs #690.

Guust-Franssens:
[...]

Issue #687, opened 2026-09-17
Qualify real Power BI credential and native-query prompts before automatic recognition


Guust-Franssens:
## Operator/customer impact

- **Why it matters:** Agents must distinguish a credential prompt from native-query approval, refresh progress, an unknown error, and no dialog. Testing only synthetic windows or one local-file error can produce confidently wrong recovery guidance.
- **Journey change:** Before: popup semantics are inferred from incomplete fixtures. After: each customer-facing verdict is qualified against the actual Power BI Desktop application in foreground and background conditions.

## Evidence-only scope

This issue creates actual-application qualification evidence and no production PR. Automatic recognition remains blocked until the mandatory positive and negative controls pass.

## Mandatory actual Desktop fixtures

1. Credential popup: SQL Server connection form and, if safely available, browser-hosted connector form. Only positive authentication evidence may support `CREDENTIAL_MISSING`.
2. Native-query approval: privacy-safe local SQL source with `Value.NativeQuery`; must produce `DIALOG_NEEDS_HUMAN`, never credential.
3. Refresh progress: negative/exoneration control. TOM/trace/elapsed heartbeat remains progress authority; prove no second refresh is stacked and progress is never a human-stop verdict.
4. Unknown local-file error: remains `DIALOG_UNRECOGNIZED`.
5. No-dialog persisted/local model: no capture or prompt verdict.

Foreground, background-visible and fully occluded modes are required. Minimized mode must be exercised and may safely return `UNKNOWN`/`TARGET_CHANGED`, never no-dialog. Include a real content-transition freshness control.

## Safety

Use only fictitious local sources and loopback services. No customer endpoint/data, entered credentials, retained authentication images, dependency installation, tenant resources, or permanent global Desktop security-setting change. UIA is first; OCR remains unapproved unless mandatory positive fixtures prove UIA insufficient.

## Acceptance

A machine-readable private matrix records actual Desktop version, sanitized fixture role, PID/HWND ownership, focus, DPI-aware completeness, UIA evidence, optional OCR, ground-truth class, and cleanup. Missing safe fixture is a documented blocker, not a synthetic substitute.

Refs #146.

Guust-Franssens:
## Actual Power BI Desktop popup evidence

These are DPI-aware captures from actual Power BI Desktop 2.158.758.0 using only fictitious/local disposable sources. No credentials were entered, no native query was approved, and no customer endpoint, identity, path, or data appears in the images.

[...]

Issue #146, opened 2026-08-14
Route positively qualified Desktop prompt evidence through probe_live_source


Guust-Franssens:
## What happened

Power BI Desktop was opened on a PBIP whose model points at a **live SQL Server that does not exist in this tenant** (`superstore.sample.db;sample_superstore`). Desktop showed its **"SQL Server database - Please specify how to connect"** credential dialog and stopped accepting operations.

**Nothing in our tooling said so.** What the tools reported instead:

```
powerbi-desktop status      -> "Host is not ready to accept operations" (retryable: true)
powerbi-desktop screenshot  -> "Print metadata is not available."      (retryable: false)
probe_desktop_query.py      -> hung >130s, no verdict
refresh_pbip_model.py       -> hung >130s, no verdict
preflight.ps1               -> "Ready to migrate", exit 0
```

Every one of those reads as a technical fault. **Two independent agents and the coordinator all concluded the Desktop Bridge had been broken by that morning's Desktop auto-update (2.157.627.0 -> 2.157.828.0)**, and it was escalated as a workshop blocker. It was not: on a model with a persisted cache, the same Desktop build and the same bridge 0.1.2 return `status: ready` and capture a screenshot with data, first try.

Cost: several hours, two agent dispatches, and a nearly-filed upstream defect report against a tool that works.

## Why this matters beyond the wasted time

`/AGENTS.md` already states the rule this violates:

> **A MISSING CREDENTIAL is not transient - try ONCE.** ... the credential sits behind a **modal sign-in dialog no automation can fill**. Stop and ask **even in an unattended run**.

We have `preflight_source_credentials.py` to classify live sources and `credential_gate.py` to stop and ask. Neither is consulted on the `pbip-model-refresh` path, so the one failure the rules single out as "stop immediately, ask a human" is the one that presents as a 130-second hang followed by a misleading technical error.

An unattended run hits this and burns its whole time budget, then reports a bridge defect.

## The dialog is trivially detectable

A visible owned window on the Desktop PID, before any query is attempted:

```
TITLE      CLASS                             SIZE
(empty)    WindowsForms10.Window.20008...    702x355     <- credential dialog
Superstore WindowsForms10.Window.8...        2011x1298   <- the real window
```

One `EnumWindows` pass, milliseconds, no new dependency. A dialog-class window owned by the target PID means **a human is required** - that is a verdict, not a timeout.

## Proposed fix

[...]

PR #693, opened 2026-09-17
fix: make custom-only probes stop before Desktop

### LINKED ISSUE #690: Do not execute full non-folding custom SQL during the default connection probe
## Operator/customer impact

- **Why it matters:** A Tableau custom-SQL relation can be small in returned rows but extremely expensive to execute. The current probe wraps it in `Table.FirstN(..., 1)`, but folding is off, so Power BI still runs the complete customer query. A credential check can therefore take minutes, consume warehouse capacity, time out, or look hung.
- **Journey change:** Before: custom SQL is fully executed during the default reachability probe. After: the default proves Power BI connector/credential reachability with a minimal source-native operation and reports the original custom query as separately validated, pending, or operator-required.

## Current measured/code evidence

`scripts/probe_live_source.py` explicitly documents that custom SQL is run through `Value.NativeQuery` / connector Query and that `Table.FirstN` still executes the full custom query because folding is off. Existing measurements include a one-row suspended-warehouse probe taking 167 seconds. Hemang's observed failures include low-volume but inefficient oracle/query workloads whose wall-clock cost is unrelated to returned row count.

## Goal

Split two different claims:

1. **Connection/credential proof:** Power BI Desktop can use its own credential store to reach the exact connector/server/database/warehouse scope.
2. **Custom-query proof:** The original custom SQL finishes successfully and returns an assessable result.

No-prompt evidence alone proves neither claim.

## Proposed default behavior

- Ordinary navigable tables retain the existing one-table real-row probe where `Table.FirstN` can fold.
- Custom SQL does **not** execute by default merely to test credentials.
- The default custom-SQL path performs the smallest connector-native operation through Power BI's credential store using the same connection scope. A shell/ODBC/CLI identity never substitutes.
- Success produces a distinct non-final verdict such as `CONNECTION_OK_QUERY_UNVALIDATED`; it must not be relabeled `DATA_OK` or clear query correctness.
- The dispatcher applies the migration brief's already-authorized build-only/degrade decision, or stops with one explicit operator/full-query validation action.
- Full custom-query validation is costly and explicit: operator-driven refresh or a clearly named opt-in after displaying the risk. It never occurs because a flag was forgotten in the opposite direction.

The independent plan review must decide exact verdict names, whether connection-only proof may clear only a credential sub-gate, and the safest connector-native operation for SQL Server, Databricks and Snowflake without rewriting customer SQL.

## Invariants

- Default custom-SQL reachability never executes the customer query.
- Connection success never claims table/query permissions, schema, semantics, rows, or final refreshability.
- A credential/native-query popup is still detected and routed immediately.
- No popup remains non-success evidence only.
- Original custom SQL is never automatically wrapped, rewritten, explained, sampled, or prefixed with TOP/LIMIT/WHERE 1=0 as if semantics were preserved.
- Query validation, when explicitly selected, uses the exact original query through the Power BI artifact/credential path and retains bounded/operator routing.

## Controls

Positive:
- A deliberately slow/non-folding custom query has an execution counter; default connection proof leaves the counter at zero.
- Minimal connection operation succeeds through Power BI and returns connection-only verdict.
- Explicit query-validation mode executes the exact custom query and can earn query success.
- Credential popup returns `NO_CREDENTIAL`; native-query approval returns `OPERATOR_REQUIRED`.

Negative:
- Connection succeeds but original query later fails permission/schema: never `DATA_OK`.
- Shell SQL/ODBC succeeds while Power BI has no credential: never accepted.
- A rewritten TOP/LIMIT wrapper is rejected as proof of the original query.
- Timeout/no-popup cannot become credential or connection success.
- Custom SQL with side-effect-shaped text is never run in the default path.

## Plan review questions

1. Can the connection-only operation use existing connector navigation/metadata without a native-query approval prompt, or is a minimal `SELECT 1` through the same connector the least-worst proof?
2. Which connection identity fields define equivalence for SQL Server, Databricks and Snowflake?
3. Should `CONNECTION_OK_QUERY_UNVALIDATED` permit model-only authoring automatically only when the migration brief already authorized degradation?
4. How do empty ordinary tables earn proof without requiring an artificial row?
5. What exact files and tests form a closed PR surface?

## Non-goals

- No replacement of `probe_bundle.py` shipped-artifact checks.
- No shell-client proof.
- No automatic custom-SQL optimization or rewriting.
- No claim that connection proof establishes data correctness.
- No full model refresh or cache persistence for reachability probing.

## Priority

Priority now: this directly addresses customer runs where a small-result custom query is computationally expensive and makes the default credential/reachability check appear unstable.

Guust-Franssens:
## Operator/customer impact

The approved #690 **SIMPLIFY TO C** behavior is unchanged: without eligible same-invocation, exact-scope ordinary `DATA_OK`, custom-only sources perform **zero customer-query, M/PBIP generation, DNS/network, or Desktop work** and earn no connection claim. Existing eligible ordinary evidence can establish connection-only reuse, never custom-query success. Both custom outcomes are nonzero, keyed `probe-error` observations with the credential/data-access gate blocked.

[Approved contract](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/issues/690#issuecomment-5717460150). Fixes #690.

## Exact head and approved N=5 delta

- **Current head:** `d21d87844afe65fff4d467ac8c6a88919f72f461`
- **Reviewed four-file parent:** `c404634ae966b9c904c00f49907e12ad0a04f8dd`
- **Exact integrated master/base:** `0fca2ae2fc5d2af760619ffc365f5d407a3a9c2e`
- This commit changes **only `tests/test_data_access_contract.py`: +38 / -17**. The four existing PR files retain their exact parent blobs and modes. No product behavior or repository documentation was changed for this delta; no merge or history rewrite was performed.

The independently approved closed surface is now **exact N=5** against that base:

1. `scripts/probe_live_source.py`
2. `tests/test_probe_live_custom_sql.py`
3. `tests/test_data_access_contract.py` — the previously omitted integration consumer
4. `docs/credential-gate.md`
5. `.github/skills/live-source-reachability/SKILL.md`

The six stale expectations are reframed around the opaque-payload refusal invariant. Empty/comment-only/None, boolean, and valid SQL inputs all require `(1, "OPERATOR_REQUIRED")`, a keyed `probe-error`, exact detail `OPERATOR_REQUIRED: custom SQL not executed; no connection claim earned`, and `blocked / probe-error` data-access state. Existing ordinary-earned setup remains a positive control. Audit-prefix assertions allow exactly the refusal to be appended, rejecting any intervening earned proof or clear. The valid-string privacy assertion is retained. There is no payload inspection or new test/helper file.

## Reproduced validation evidence

All commands were judged by exit code. The base replay used a new detached worktree at the exact base above; its candidate test blob was identical to the head test blob: `928914b0762333a4ab062311361a21e345d3dae3`. The detached worktree and run-owned test scratch were removed after verification; logs/XML/command evidence remain under ignored `_review690/n5-delta/` in the PR worktree.

| Control | Result | Exit |
|---|---|---:|
| Original six expectations, exact base | 6 passed | 0 |
| Original six expectations, reviewed parent | 6 failed: four ERROR expectations, expected TypeError, and expected DATA_OK | 1, expected |
| Revised exact six cases, current candidate | 6 passed | 0 |
| Identical revised test bytes, exact base | 6 failed: five outcome assertions plus the boolean payload's old TypeError; no collection/import failure | 1, expected |
| Exact focused #690 controls | 72 passed | 0 |
| Smallest broader offline probe/data-access matrix | 489 passed | 0 |
| Revised six after isolated mutations | 6 passed | 0 |
| `ruff format tests\test_data_access_contract.py` | Already formatted | 0 |
| `ruff check tests\test_data_access_contract.py --fix` | Clean | 0 |
| `pylint tests\test_data_access_contract.py` | 10.00/10, unchanged from pre-edit baseline | 0 |
| `git diff --check` and protected four-file comparison | Clean; all four parent blobs/modes unchanged | 0 |

### Exact test selections and commands

These PowerShell arrays expand to the exact node arguments used; the first array selects four parameterizations plus two single cases.

```powershell
$Six = @(
  'tests\test_data_access_contract.py::test_custom_sql_operator_required_invalidates_earned_key'
  'tests\test_data_access_contract.py::test_nonstring_custom_sql_is_opaque_and_invalidates_earned_key'
  'tests\test_data_access_contract.py::test_custom_sql_without_ordinary_evidence_never_earns_proof'
)
$OldSix = @(
  'tests\test_data_access_contract.py::test_custom_sql_preprocessing_failure_invalidates_earned_key'
  'tests\test_data_access_contract.py::test_malformed_custom_sql_type_also_records_its_terminal_error'
  'tests\test_data_access_contract.py::test_custom_sql_success_is_keyed_and_earns_only_after_clear'
)
$Focused = @(
  'tests\test_probe_live_custom_sql.py::test_custom_only_has_no_automatic_operation'
  'tests\test_probe_live_custom_sql.py::test_cli_loader_leaves_payload_opaque'
  'tests\test_probe_live_custom_sql.py::test_custom_payload_is_not_read_and_direct_table_entry_is_also_closed'
  'tests\test_probe_live_custom_sql.py::test_parser_self_closing_text_relation_cannot_clear_as_a_physical_table'
  'tests\test_probe_live_custom_sql.py::test_both_custom_outcomes_keep_the_real_audit_and_gate_armed'
  'tests\test_probe_live_custom_sql.py::test_verdict_mutations_fail_the_intended_assertion'
  'tests\test_probe_live_custom_sql.py::test_reintroduced_custom_operations_are_detected'
)
$Matrix = @(
  'tests\test_data_access_contract.py'
  'tests\test_probe_live_custom_sql.py'
  'tests\test_probe_earned_clear.py'
  'tests\test_probe_live_source_verdict.py'
  'tests\test_probe_live_source.py::test_ordinary_probe_requires_child_row_evidence'
)

# Revised candidate/head: each exit 0.
python -m pytest -q --tb=short --basetemp=_review690\n5-delta\head-six @Six
python -m pytest -q --tb=short --basetemp=_review690\n5-delta\focused72 @Focused
python -m pytest -q --tb=short --basetemp=_review690\n5-delta\matrix @Matrix
python -m pytest -q --tb=short --basetemp=_review690\n5-delta\post-mutation-six @Six

# Original expectations on reviewed parent: exit 1, six stale failures.
python -m pytest -q --tb=short --basetemp=_review690\n5-delta\old-head-six @OldSix

# Exact detached base, before/after applying only the candidate test delta:
python -m pytest -q --tb=short --basetemp=_n5-validation\old-six @OldSix       # exit 0
python -m pytest -q --tb=short --basetemp=_n5-validation\revised-six @Six   # exit 1
```

### Direct mutation controls — no new harness or tracked file

Each mutation was process-local/in-memory; the shipped product files were never edited. `python -` invoked the existing pytest nodes. JUnit results independently confirmed the stated intended assertion failures with **zero setup/import errors**; the result-verification command exited 0.

| Mutation | Existing/new oracle exercised | Observed kill |
|---|---|---|
| In `_probe_one_table`'s custom branch, read `target[0].get("custom_sql")`, then check `isinstance(sql, str)` / `sql.strip()` before returning the same refusal | Existing `test_custom_payload_is_not_read_and_direct_table_entry_is_also_closed` | 1 failed, exit 1: `the default path read the custom SQL payload` |
| Replace `_custom_sql_stop` with the same return and exact `probe-error` detail, but call `credential_gate._audit` without `sources` | All six revised nodes | 6 failed, exit 1: exact-key assertions observe `None` instead of `[KEY]` |
| Append unearned `probe-data_ok` and `probe-cleared` observations before the otherwise unchanged keyed refusal | All six revised nodes | 6 failed, exit 1: `custom SQL must append only its refusal, never earned proof or a clear` |

The unkeyed control demonstrates that a blanket, correct-looking OPERATOR_REQUIRED is insufficient. No additional test file or framework is needed.

## Limits and review status

- This is offline evidence: external Desktop/database effects are mocked. No live database, connector qualification, customer-query execution, or real Desktop validation is claimed.
- Full Linux/Windows CI at this new SHA was not rerun locally; the push starts fresh CI. Earlier full-suite/lint results at the parent are historical, not verification of this new SHA. The unrelated Windows ACL/modal cohort is outside this test delta.
- No #692/#694 strategy, shared-loader, scaffold, credential-gate implementation, popup, or other scope expansion.
- **Keep draft and unmerged.** The N=5 head still needs the promised blind review/mutation review. Reviewed-parent approval is not represented as approval of the new SHA.


Guust-Franssens:
## Design hold - September 17, 2026

The user requested reconsideration of the navigation-only policy against an immutable generated `SELECT 1 AS ProbeOK` through the same Power BI connector. Navigation-only is **not final/merge-approved**. This PR remains draft and unmerged at `5c736c540ff5a4fc5b7dd16ff94a277901b2c122`; no production change, approval click, customer-SQL execution, or #692 implementation is being made during comparison. Renewed adjudication must cover scope, connector cost, and qualified native-approval routing.

Guust-Franssens:
## Comparison result - renewed policy approval required

The production head remains **5c736c540ff5a4fc5b7dd16ff94a277901b2c122**, draft and unmerged. No production file, commit, push, native-query approval, Desktop security setting, or #692 behavior changed during this comparison.

An isolated, unshipped comparison lab exercised both M alternatives through the same one-table PBIP, child-dispatch fixture, real verdict/audit code and unchanged gate reader. **67 tests passed (exit 0); Ruff and Pylint 10.00/10 passed (exit 0).** These are offline dispatch/shape/authority controls, not live connector performance or billing measurements.

| Connector | Requested scope retained | Generated constant compared with current navigation |
|---|---|---|
| SQL Server | Normalized server with port/instance and exact database | `Sql.Database(server, database, [Query="SELECT 1 AS ProbeOK"])` replaces the explicit buffered object-navigation result. This is the clearest reduction in emitted metadata work. Driver-internal calls are not measured. |
| Databricks | Host, HTTP path, exact catalog/database; same `Databricks.Catalogs(..., null)` target | The candidate selects the same database node and submits the fixed statement through `Value.NativeQuery`. It removes explicit buffering of that node's schema names, **not** the catalog lookup. Exact provider/version execution remains unqualified. |
| Snowflake | Account, warehouse, role and exact database; same `Snowflake.Databases` target | Same database-node selection, then the immutable constant. It removes explicit schema-name buffering, not all metadata calls. The prototype normalizes the single result-column label to `ProbeOK` without modifying the SQL statement: unquoted Snowflake identifiers are uppercased. |

For both alternatives, all declared connection fields still constrain invocation-local reuse. An ordinary same-scope DATA_OK avoids either extra operation. Different database, port/instance, HTTP path, warehouse, role or credential hints could not borrow evidence. Neither candidate reads/rewrites the customer SQL payload; the side-effect sentinel remains absent from invoked M/argv/logs, and its dispatch counter stays zero. The only native SQL text supplied by the prototype is the immutable generated statement. Connector-internal setup/metadata statements are not counted or constrained by this lab; one physical server statement is not promised. This says nothing about login/audit/startup side effects of making a database connection.

Successful child evidence remains **CONNECTION_OK_QUERY_UNVALIDATED / exit 1**, the keyed `probe-error` audit envelope carries that explicit detail, and the gate remains armed. Neither creates authorization, proof names, probe-cleared, customer-query permission/schema/row proof, or model-only permission. A no-popup transcript or timeout remains non-success. Empty navigation still refuses under the current row contract; a fixed constant is not logically dependent on having an enumerable table/schema row (demonstrated with controlled child outcomes, not a live grants experiment).

### Cost and exact-scope caveats

- Current navigation deliberately uses eager `Table.Buffer`; its documented cost includes reading/buffering scalar cells. Removing that explicit operation is useful but is **not a measured wall-clock or billing win**.
- A Databricks SQL warehouse can auto-start on a driver connection. Therefore neither approach can promise zero resumed compute; a one-row constant is not a zero-cost guarantee.
- Equal requested M parameters do not attest every effective server session setting. Microsoft documents that an unavailable Snowflake role is not set by the connector. A constant alone cannot prove which role/warehouse was actually effective, and navigation alone does not solve that either. Renewed approval must keep this claim ceiling explicit.

### Native approval: evidence versus current implementation

The actual #687 matrix and [its public disposition](https://github.com/Guust-Franssens/tableau-to-powerbi-migration/issues/687#issuecomment-5713376016) confirm the real LocalDB constant-query approval form on Desktop 2.158.758.0. **Run was not pressed.** But the production detector in that acquisition returned **DIALOG_UNREADABLE**; the qualification report explicitly sets `automatic_recognition_approved=false`. OCR read the approval wording only in the evidence experiment.

The comparison confirms today's parent maps DIALOG_UNREADABLE, and even its existing mixed DIALOG_NEEDS_HUMAN native/authentication token, to **ERROR**, not OPERATOR_REQUIRED. Its generic OPERATOR_REQUIRED directive is not yet suitable either. The desired future routing is **qualified native approval -> OPERATOR_REQUIRED**, without automatically approving it or inferring its class merely because a constant was attempted. Positive credential, unknown, progress and no-dialog cases must remain separate. **#146/#687 remain the recognition/consumer dependency**, not something the M change can assume is already implemented.

### N=4 and adjudication

The generated constant, fixed result-label projection, unchanged connection-only verdict/audit, and corresponding focused tests/docs can mechanically fit the original **N=4**. Nevertheless, replacing the explicitly approved no-native-statement policy is a substantive plan change and **requires renewed independent approval**. Qualified prompt production/consumption is a separate dependency; it must not be smuggled into this slice. The protected legacy test-contract update also still needs explicit scope approval.

The constant is a viable candidate for the narrower credential/session question; the mere existence of an approval prompt is not sufficient reason to reject it. There is no final choice or merge approval here. **Hold #693 pending adjudication.**

Primary references:
- [Sql.Database options](https://learn.microsoft.com/en-us/powerquery-m/sql-database)
- [Table.Buffer evaluation/cost](https://learn.microsoft.com/en-us/powerquery-m/table-buffer)
- [Native-query security/approval](https://learn.microsoft.com/en-us/power-query/native-database-query)
- [Databricks native-query configuration](https://learn.microsoft.com/en-us/azure/databricks/partners/bi/power-bi-desktop)
- [Databricks ODBC compute/start behavior](https://learn.microsoft.com/en-us/azure/databricks/integrations/odbc/compute)
- [Snowflake role/driver behavior](https://learn.microsoft.com/en-us/power-query/connectors/snowflake)
- [Snowflake identifier case](https://docs.snowflake.com/en/sql-reference/identifiers-syntax)

Guust-Franssens:
## Binding closed-dialect constraint for renewed adjudication

This supersedes the universal-literal assumption in the earlier unshipped comparison. Production stays frozen at **5c736c540ff5a4fc5b7dd16ff94a277901b2c122**; this comment is not implementation or merge approval.

Required contract:

1. Select health SQL from a **closed connector/class + supported compatibility/version profile + template revision** mapping. No generic `SELECT 1`, SQL-like-class inference, implicit alias, or guessed version policy. Oracle and other connectors remain outside #690.
2. Unsupported or unestablished compatibility returns nonzero **`ERROR: no safe connection probe` before generation or reuse**. There is no fallback to customer SQL or catalog enumeration. Required database-node navigation within an explicitly selected Databricks/Snowflake connector is not a fallback.
3. Retain exact source credential/session scope. Cache identity additionally binds connector, selected profile, template revision and exact generated statement identity. No cross-connector/template/version reuse.
4. Audit the same immutable selection used for M: connector, compatibility profile, template ID/revision, exact generated SQL and its full UTF-8 SHA-256, plus the method. **Ordinary-row reuse must say `ordinary-row-reuse`, not that the health query was executed.** The existing keyed non-success audit envelope is retained; no new unrecognized action, probe-cleared, authorization or gate lift.
5. Supported completion stays **CONNECTION_OK_QUERY_UNVALIDATED / nonzero / gate armed**. Customer SQL is never read to select, construct, normalize or hash a health statement. Qualified native approval still needs OPERATOR_REQUIRED, never automatic approval or a credential guess. #692 remains separate.

### Narrow contract evidence, not production activation

An isolated, ignored lab now has **41 passing controls** (pytest exit 0; Ruff and Pylint 10.00/10 exit 0). These check exact class selection, unsupported/missing/version refusal before cache reuse, cross-connector template/render/audit rejection, exact SQL decoded from the one-table PBIP matching its audit identity, explicit reuse provenance, and mutations to a generic statement, customer SQL, revision or version.

The three separately identified **syntax candidates** exercised are:

| Connector | Candidate template | Identity family |
|---|---|---|
| SQL Server | `SELECT 1 AS [ProbeOK]` | `sqlserver-select-one/v1` |
| Databricks | `SELECT ProbeOK FROM VALUES (1) AS health(ProbeOK)` | `databricks-values-one/v1` |
| Snowflake | `SELECT 1 AS "ProbeOK"` | `snowflake-select-one/v1` |

The Databricks candidate uses its explicitly documented inline VALUES form rather than treating FROM-less SELECT as a universal grammar rule. These are fixed generated statements with no customer-object access. Tests use clearly labeled **`fixture-only-v1` approvals**; that is not a real server-version qualification or a new production allowlist. No native query was approved or executed against a source.

The current parser/spec does not establish database-server/connector version provenance (the captured Tableau version is not that evidence). Renewed adjudication must name the supported compatibility policy and its authoritative input; the lab does not invent either. This can be represented inside the originally owned script/tests/docs once approved, without an extra production module.

**No all-database matrix gate is being added.** Per the latest #690 disposition, #694 remains enhancement/qualification follow-up, not a merge prerequisite. The immediate safety fix may cover its explicitly justified connector profiles and refuse everything else. The current strategy/adjudication and protected legacy-test scope still need resolution; #693 remains draft/unmerged.

Primary syntax evidence: [SQL Server SELECT grammar](https://learn.microsoft.com/en-us/sql/t-sql/queries/select-transact-sql?view=sql-server-ver17), [Databricks VALUES](https://learn.microsoft.com/en-us/azure/databricks/sql/language-manual/sql-ref-syntax-qry-select-values), [Snowflake SELECT and quoted aliases](https://docs.snowflake.com/en/sql-reference/sql/select). These support syntax reasoning; they do not silently activate every historical or future release.

Guust-Franssens:
## Blocked qualification handoff — superseding the earlier readiness disposition

The September 17 user direction withdraws #690 implementation-ready approval. **#693 remains draft/unmerged at 5c736c540ff5a4fc5b7dd16ff94a277901b2c122. Neither navigation nor generated SQL is finalized as production.** Earlier comments describing live qualification as merely optional follow-up no longer express the current decision. No live source, credential, Desktop instance, new production strategy, commit or push was used in this handoff.

### Bounded offline evidence retained

A fresh before/after run used the actual probe code at base `9be30501ebc08d72673f71bd30ef1b3aa0e7fad1` and the frozen draft, through resolution, generated one-table PBIP and a **mock child dispatch**. For each of SQL Server, Databricks and Snowflake, a customer-SQL identifier sentinel reached the mock child **once at base and zero times at the draft**. The base returned `(0, DATA_OK)` against the synthetic child; the draft returned `(1, CONNECTION_OK_QUERY_UNVALIDATED)`. Spec bytes were unchanged. These are dispatch/emission results, **not server execution counts or live connection evidence**.

The focused suite was rerun: **81 passed, exit 0**, including seven intended-assertion mutation controls. Earlier 67 M-comparison and 41 dialect-contract controls remain offline-only. Their version approvals were synthetic fixtures. They do not prove supported backend versions, connector behavior, costs, effective sessions, grants, or automatic popup routing.

### Candidate M shapes only

| Connector | Navigation candidate shape | Generated-query candidate shape |
|---|---|---|
| SQL Server/Azure SQL family | `Sql.Database(exact server/port/instance, database)` -> eager scalar navigation projection | Same configured connector scope with `Query=<closed dialect template>` |
| Databricks SQL Warehouse | `Databricks.Catalogs(host, http_path, null)` -> exact database/catalog node -> eager scalar navigation projection | Same configured root/database node -> `Value.NativeQuery(..., <closed dialect template>)` |
| Snowflake | `Snowflake.Databases(account, warehouse, Role)` -> exact database node -> eager scalar navigation projection | Same configured root/database node -> `Value.NativeQuery(..., <closed dialect template>)` |

The strings establish only what was emitted. They do not establish whether metadata is cached, whether a round trip occurs, which operations/permissions are needed, which effective role/catalog is used, whether a warehouse resumes, or how a provider handles approval and column types/casing. No faster/cheaper/default-strategy conclusion follows from these shapes.

### Required actual Power BI matrix

**Six baseline cells: these three connector families x navigation/generated-query.** Test each on an explicitly authorized, disposable **non-customer** environment, with named supported backend/connector versions. Use paired warm and controlled cold/start conditions where supported and authorized. This is the current three-connector scope, not a claim to cover every database or future release.

Each connector's cells need:

1. **Independent scope evidence:** exact PBIP/M hash, Desktop version, connector implementation/driver and backend version; actual principal and database/catalog/HTTP-path/warehouse/role where applicable. Correlate the PID/model-bound Power BI refresh with source-side request/query IDs or session/trace evidence. A shell/ODBC success is never the connection oracle. Establish the version/profile input rather than substituting Tableau's version.
2. **Positive and object-access controls:** valid Desktop credential; a real accessible one-row fictitious table; a definitely nonexistent fully qualified table; an existing visible table with SELECT denied; empty/restricted navigation where constructible. Record actual results, driver error/SQLSTATE and navigation visibility. Connection evidence must not become object/customer-query access evidence.
3. **Credential and prompt controls:** missing/rejected/revoked Desktop credential, including warm/cached-metadata conditions; actual native-query approval versus credential/progress/unknown/no-dialog evidence. Do not auto-approve. A positive constant-query execution, if approval is required, needs an explicitly consenting operator; otherwise that cell stays blocked. #687's SQL Server popup does not substitute for Databricks/Snowflake evidence. #146/#687 own qualified prompt consumption.
4. **Cost/liveness evidence:** separate Desktop open/catalog/refresh times, observed metadata requests/rows, source query activity, source startup/resume state and measured compute/billing telemetry when available. Record internal connector setup/metadata SQL separately from the fixed health template. No cost claim from returned row count; no inference from no popup or a timeout.
5. **Customer-SQL non-execution:** a harmless, owned side-effect-shaped sentinel fixture with an independently checked counter/attempt log. Verify the probe never submits its SQL, rather than reading a zero counter that could also mean the query was rejected before the side effect. Do not execute real customer SQL. Retain exact generated-template identity and decoded M SQL; distinguish client-emitted text from server-observed statements.
6. **Refusal/reuse/authority:** unsupported connector/class/version -> `no safe connection probe`, no customer-SQL or catalog fallback; mismatch database, HTTP path, warehouse, role, credential, connector or template identity must not borrow evidence. Demonstrate same-scope ordinary-row reuse separately. Success remains CONNECTION_OK_QUERY_UNVALIDATED/nonzero, audit is keyed and names method/query identity, and no probe-cleared, authorization or gate lift occurs. Record marker/ACL/status plus verify result in context: an empty, correctly enforced gate can be compliant without being cleared.

Every cell is **PASS / REFUSE / BLOCKED / CANNOT_ESTABLISH** from actual evidence. A missing environment, principal, telemetry source, native approval or readable prompt is a blocked/unestablished cell, never a synthetic pass. Keep authentication imagery private under the existing retention policy; publish only sanitized evidence. Dispose of only test-owned resources and processes.

**Current result: BLOCKED / PENDING ACTUAL CONNECTOR QUALIFICATION.** The source-query forwarding removal is demonstrated offline; connector strategy selection is not demonstrated. #692 exact customer-query execution remains out of scope.

